PR #264, opened 2026-07-11
docs(specs): Workspace & Identity schema + security contract v1 (Gate 1, spec-only)

Talchain:
## Gate 1 deliverable — Workspace & Identity lane (ROADMAP 3.9)

**SPEC ONLY — nothing in this PR executes or authors runnable DDL.** Gate 2 authors migration files from this contract after A1's Gate 1 review (adversarial + security, per charter).

Authorized by `parallel-briefs/A1-GATE0-VERDICT-workspace-2026-07-12.md` (Gate 0 PASS AS AMENDED; Gate 1 on the unblocked surface).

### Contents
- Full DDL-as-spec: `workspaces` / `workspace_members` / `workspace_invites` (dark) / helpers / owner-atomic `create_workspace` / personal-workspace provisioning + 16-user backfill (§1); `scenarios.workspace_id` + trigger stamping + immutability + 18-row backfill, scenarios RLS byte-unchanged (§2); `person_profiles` (owner-only forever) + **immutable `profile_consent_events` ledger** (INSERT-only as hardware incl. service_role UPDATE/DELETE revoke) + `has_active_consent` fold + Lock-0 `get_coaching_profile` shapes + `delete_person_profile` (§3)
- Function-access matrix — caller class / DEFINER-vs-INVOKER / pinned search_path / grants / subject-derivation per function (§4)
- RLS policy matrix + F9 grants baseline + migration reversibility classes + ledger-reconciliation hard blocker (§5)
- Extended P12 disposition matrix for every `user_profiles` column, evidence-backed (§3.5)
- Cross-cutting integrity rules incl. strategy-layer composite FKs and the decision_records derive-from-locked-row ruling (§6)
- Answers to the collab-tenancy audit's XQ-1/XQ-2 per A1 order (§7)
- Org-spine + strategy-tree drafts explicitly **PENDING-CQ** (§8); Gate-4 test-pack obligations (§9)

### Review asks
1. A1 adversarial review + **security review** (everything here is RLS/grants/RPC-shaped)
2. Ruling on §3.4 Shape A vs Shape B for `get_coaching_profile` (recommendation: B only)
3. Consent-ledger + P12-extension + XQ-1/2 answers ride the queued Paul ratification batch

Evidence base: `parallel-briefs/workspace-lane-evidence/gate0/` (SHA256SUMS'd catalog + migration registry). Push note: pre-push hook bypassed for this docs-only branch (validate-prepush cannot run in a bare worktree; the hook's own header defers to CI).

[...]

PR #265, opened 2026-07-11
feat(migrations): W&I Gate 2 — M1 tenancy core + M2 scenario scoping (AUTHORED, not executed) + F3 hotfix record

Talchain:
## Gate 2 (scoped) — Workspace & Identity lane (ROADMAP 3.9)

**MIGRATION-ONLY, AUTHORED NOT EXECUTED.** Scope per the Gate-1 verdict (`parallel-briefs/A1-RULING-F3-AND-GATE1-2026-07-12.md`): **M1 + M2 authoring only; M5 held** (CQ-6/R-1/R-2). Contract of record: `docs/specs/workspace-identity-schema-contract-v1.md` **v1.2** (PR #264, Gate-1 PASS-WITH-CONDITIONS — all four conditions folded there before this DDL was authored). **Execution is blocked until the ledger-reconciliation register precedes it**, then A1/Paul-batched.

### Files
- `20260712063000_f3_drop_cross_org_profile_policy.sql` — the F3 containment **already executed by A1** under the emergency-security exception; this is the durable, replay-safe record the exception terms require (idempotent no-op on re-apply).
- `20260712100000_workspace_identity_m1_tenancy_core.sql` — wsid_definer NOLOGIN role · workspaces + workspace_members (FORCE RLS, deny-by-default, explicit `TO wsid_definer` policies — PG15 CREATEROLE cannot mint BYPASSRLS, contract §5.1a) · role_rank/is_workspace_member/is_workspace_role · guard triggers WS001/WS002/WM409/WM410 (exactly-one-owner + personal single-member) · owner-atomic create_workspace (advisory-lock serialised) · personal-workspace provisioning trigger + **dynamic** backfill · C1/C2 auth.users deletion trigger pair (WM412 transfer-required interim rule, CQ-5-pending). Deliberately absent: invites (PENDING-P9), org_id (M3-atomic), the orchestration RPC (M5 — trigger pair is the enforcement point).
- `20260712101000_workspace_identity_m2_scenario_scoping.sql` — authored **against live state** (611-guest-row reality, dynamic counts): scenarios.workspace_id (FK **ON DELETE SET NULL**, records-outlive-container) + `UNIQUE(id, workspace_id)` (composite-FK target) + WS020-abort stamping trigger + WS010/WS011 immutability (sentinel-guarded value→NULL — FK SET NULL fires row triggers, v1.2 erratum; claim-shaped NULL→value) + orphan-tolerant backfill **sequenced before the immutability trigger exists**. Scenarios RLS byte-unchanged (P3), asserted in-transaction.
- `supabase/MIGRATION-DRIFT-REGISTER.md` — the A1-adopted 'applied, pending ledger row' register; entry #1 = the F3 hotfix; rules for future entries; pointer to the Gate-0 three-class drift inventory.
- Rollbacks (`.do-not-apply`) for M1/M2 with adoption/cutover refusal guards (M1 refuses if non-personal workspaces exist; M2 refuses post-cutover).

### Review asks
1. A1 adversarial + security review (verify DDL ↔ contract v1.2 clause-by-clause — the exact-matrix spec exists for mechanical comparison).
2. Confirm the M1→M2 execution order note + the drift-register-first execution gate.
3. Gate-4 test pack (allow/deny per policy row, guard races, claim-shape transitions) follows in the Gate-4 PR per contract §9 — not in this migration-only PR (charter).

Push note: pre-push hook bypassed (validate-prepush cannot run in a bare worktree; SQL-only change; CI validates on the PR).

### DIFF
--- a/supabase/MIGRATION-DRIFT-REGISTER.md
+++ b/supabase/MIGRATION-DRIFT-REGISTER.md
@@ -10,9 +10,14 @@ PII/security exposure + single reversible statement + immediate register entry).
 
 ## Applied, pending ledger row
 
-| # | Migration file (repo path) | Content sha256 | Applied | Authority | Evidence | Ledger row |
-|---|---|---|---|---|---|---|
-| 1 | `supabase/migrations/20260712063000_f3_drop_cross_org_profile_policy.sql` (DGAI) | `fa946a710e708c189ff70f9f623766a76993f5b37db550f581de905989ad07f3` | 2026-07-12 ~06:00, A1 direct (transactional, strengthened pre/post-checks + simulated-JWT allow/deny) | Emergency-security exception, `A1-RULING-F3-AND-GATE1-2026-07-12.md` §F3 | `acceptance-evidence/security/F3-CONTAINMENT-2026-07-12.md` + `parallel-briefs/workspace-lane-evidence/gate0/` (exposure quantification) | **PENDING** reconciliation plan |
+Each entry distinguishes FOUR artefacts (they are not interchangeable): the **executed
+script + transcript** (byte-exact, what actually ran), the **canonical replay migration**
+(idempotent repo record — the sha256 column hashes THIS file, not the executed script),
+**behavioural evidence**, and **rollback/recreation SQL**.
+
+| # | Replay migration (repo path) | Replay-file sha256 | Executed script + transcript | Applied | Authority | Behavioural evidence | Rollback | Ledger row |
+|---|---|---|---|---|---|---|---|---|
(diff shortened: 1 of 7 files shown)


Talchain:
## ABANDONED — closing. Collaboration was solved a different way, and the one security item here was already executed.

Re-derived at UI staging tip `37f626b6` (17 Aug), five weeks after this PR was opened.

**The architecture this PR prepares for was never adopted.** At the tip there is no tenancy layer at all: `workspace_id` = **0 files**, `org_id` = **0**, `workspaces` = **0**, `tenancy` = **0** across `supabase/` and `src/`. Multi-user collaboration — the capability tenancy was meant to underwrite — shipped instead on its own token model: `src/collab/participantToken.ts`, `src/collab/ownerAccessToken.ts`, `src/collab/collabService.ts`, ES256-verified, journey-witnessed. Enterprise tenancy, SSO and procurement controls are out of PoC scope by standing ruling, and the PoC's collaboration proof no longer depends on them. So these migrations are not "pending" — they are prepared for a design the programme routed around.

**The F3 item is a paper record, not an unlanded fix — checked before closing, because closing away a live RLS hole would be unacceptable.** `20260712063000_f3_drop_cross_org_profile_policy.sql` documents a drop that was **already executed on staging Supabase (`etmmuzwxtc`) on 2026-07-11 18:24 UTC** under the emergency-security exception, as the file's own header states. And a rebuild cannot re-open it: the leaky policy `"Users can view accessible profiles"` is **created by no migration at the tip** — its only occurrence anywhere under `supabase/` is `supabase/backups/staging_pre_scenario_schema_v2.sql`, a pre-change snapshot, which is a record of the old state and not a thing that runs. Controls on that sweep: positive `"Users can view own profile"` = 4 files, contrast `"zzq fabricated policy"` = 0. Read at the bytes (`grep -a`, then `od -c` on the matching line) after an earlier rendering of the same file looked mangled — the bytes are `ON public.user_profiles`, and the absence claim is byte-level, not display-level.

**What is genuinely left behind is documentary**, and deliberately so: `supabase/MIGRATION-DRIFT-REGISTER.md` and the Gate-0 drift inventory (11 name-only-ledgered CEE files, 6 applied-but-unledgered, 1 authored-never-applied, the `20260226010000` cross-repo version collision). That is register/reconciliation work, which the programme's standing freeze keeps out of implementation seats unless it blocks a product lane. It does not.

Merging as-is would apply three unexecuted tenancy migrations against a schema that has moved 105 migrations since they were authored. The branch retains everything; nothing is deleted. Closed as abandoned, not as done.

PR #646, opened 2026-08-10
PR3: a menu the assistant raised says so, and says how to take it back

Talchain:
## Acceptance measure (one falsifiable sentence)

**A user whose menu opened without them touching it can now read, inside that menu, that Olumi opened it and that Escape takes it back — and we would observe it by raising `top_bar_menu` through `requestOverlaySurface` and finding the exact strings "Opened by Olumi" and "Esc to dismiss" rendered inside `role="menu"`, while the same menu opened by the user renders neither.**

Both halves are witnessed in a real browser below, not only in jsdom.

---

## ⚠ CORRECTED PREMISE — READ THIS FIRST

The brief for this lane said *"the assistant can now raise a menu"*. **That is true of the store and false of the wire, and the difference decides how this PR should be valued.**

Derived at the bytes at `60c4b082`:

- **`requestOverlaySurface` has ZERO production call sites.** Complete manifest of every reference under `src/` (`rg` over all `.ts`/`.tsx`): `src/stores/uiStore.ts` (declaration ×2, impl, 3 comment lines) and three **test** files. `applyV5State.ts` never calls it.
- **Schema `0.39.0`** (`vendor/talchain-schemas-0.39.0.tgz` → `dist/boundary/blocks.d.ts`): `UiDirectiveVerb = ["highlight","focus","open_inspector","open_panel","open_section"]`, and `UiDirectiveUiTargetSchema` is a discriminated union over `kind ∈ {"tab","model_section"}`. **No overlay verb and no overlay target kind exist.** The schema's own comment records the ratified wave-2 verdict: *"NO new verbs"*.
- `applyV5State`'s panel branch dispatches `open_panel → forceActivateOutputTab` and `open_section → requestModelTabSection + forceActivateOutputTab('diagnostics')`. Neither touches `activeOverlaySurface`.

**Therefore `overlaySurfaceOrigin` can only ever be `'user'` on a real turn today, and `data-overlay-origin="assistant"` is unreachable from the wire.** This is trap 16-inverse: the code path is live, the producer cannot feed it.

**What that means for this PR.** It completes the display half of a capability whose state half (#645) is dark for the same reason — it does not itself make anything reachable. Rung on the status ladder: **TESTED + browser-witnessed against the store seam; NOT wire-witnessed, NOT journey-witnessed.** I did **not** widen scope to fix reachability: the smallest enabling change is a `{kind:'overlay_surface'}` target in `olumi-schemas` plus a CEE emit and an `applyV5State` dispatch — three repos, and a Paul-ratifiable contract decision that reverses a recorded verdict.

**More valuable adjacent finding, deliberately not actioned.** `open_panel` / `open_section` **are** reachable today and **do** move the user's UI on a real turn — switching their tab, opening a model section — with **no attribution of any kind**. That is the *reachable* instance of exactly the harm this lane was scoped to fix. It lives in `OutputsDock`/`uiStore`, owned by other lanes.

---

## What the badge says, and what it refuses to say

An outlined pill at the top of the dropdown, above the existing divider:

> ✨ **Opened by Olumi**    ·    Esc to dismiss

- **Where it sits:** inside the surface it explains. The answer is where the question ("why did that open?") is asked, not somewhere the user has to hunt.
- **Calm:** non-interactive, no animation, no focus steal, no colour alarm. A user who did not notice the menu open is not startled by the explanation of it.
- **It does not say WHY, because no reason is reachable.** `ui_directive.note` is the only free-text carrier on the 0.39.0 wire and **nothing in `src/` reads it** (`rg '\.note\b|note:' src/v5/applyV5State.ts src/v5/blocks/mapV5Blocks.ts` → 0 hits). A plausible-sounding rationale on the provenance channel is worse than no badge, so the copy is pinned as a literal in both the component and the spec, and a test asserts the badge contains **nothing else**.
- **The dismissal hint is a verified fact, not reassurance:** TopBar's Escape handler runs `closeMenu()` unconditionally on origin.
- **Accessibility:** inside `role="menu"` only menuitems are reliably announced, so a bare `<div>` would give a screen-reader user — the person most disoriented by a menu that appears unprompted — nothing. The menu's accessible **name** carries it (`"More options — Opened by Olumi"`), and is left undefined when the user opened it themselves.

## Real-browser readings (dev build, Chrome — the thing jsdom cannot prove)

Driven through the assistant's actual seam: `(await import('/src/stores/uiStore.ts')).useUIStore.getState().requestOverlaySurface('top_bar_menu')`.

| | reading |
|---|---|
| menu background | `rgb(255,255,255)` |
| pill ink | `rgb(63,63,62)` (`--text-body`) — **10.54:1** ✅ AA/AAA |
| pill background | `rgba(0,0,0,0)` — transparent |
| pill border | `color(srgb 0.152941 0.478431 0.615686 / 0.3)` — the info blue at 30%, **colour on the border** |
| hint ink | `rgb(110,107,107)` (`--text-light`) at 12px — **5.28:1** ✅ AA |
| box | 132×22, `visibility: visible`, `elementFromPoint` at its centre resolves **inside the badge** (not occluded) |
| menu `aria-label` | `"More options — Opened by Olumi"` |

This follows the shipped convention (DS v5 §3.2 / `DESIGN_SYSTEM.md`: outlined pills, `bg-transparent border-{colour}/30 text-text-body`) rather than inventing one. The sibling lane's 2.80:1 / 2.83:1 came from colouring the **text**; mutant **M6b** reproduces that exact failure and is caught.

Real-browser negative and takeover cases, same session:

| action | menu | badge | `data-overlay-origin` | menu aria-label |
|---|---|---|---|---|
| assistant raises | ✅ | ✅ | `assistant` | `More options — Opened by Olumi` |
| Escape | ✅ closed | ✅ gone | — | — |
| **user** opens via kebab | ✅ | **✅ absent** | `user` | **`null`** |
| click outside | ✅ closed | ✅ gone | — | — |

## Test strategy

New spec `TopBar.overlayAttribution.spec.tsx` — **12 tests, asserted collected BY NAME** (`Tests 12 passed (12)`), run with `VITE_SUPABASE_URL`/`VITE_SUPABASE_ANON_KEY` exported. RED-first at pristine: **8 failed / 4 passed of 12**.

The truth condition is a **conjunction** — *this bar's surface is raised* **AND** *the assistant raised it*. Either conjunct alone admits a state where the sentence would be a lie, so both are pinned.

**The style guard reads the stylesheet from disk.** vitest runs `css: false`, so a CSS-module lookup returns a key-echoing Proxy and a `className` assertion passes against a class that generates **no CSS** — the defect a sibling PR shipped tonight (`text-text-muted`). `?raw` does not escape it either: the stub intercepts `*.module.css` and `topBarCss.length` came back as the **string** `"length"`. The guard reads the file and proves it can see a presence (length > 1000, contains `.dropdownMenu`) before claiming an absence.

## Mutation kit — 8/8 bit, each on the expected test

Throwaway worktree at an **absolute** path outside the repo root. **Isolation proven by WRITING**, not by locating: distinct inodes (`578916560` vs `578928433`), a sentinel appended to the worktree copy left the source SHA-256 **identical**, and restores came from a **pristine `git archive` tar**, never from the other copy. Applied-check scoped to `src/`; **control read exactly 0**. Every run asserted collected == 27 (12 mine + 15 sibling) or hard-errored.

| mutant | expected | result |
|---|---|---|
| **M1** delete the badge render | capability RED | 🔴 6 RED incl. the capability test |
| **M2** loosen the ORIGIN conjunct (`=== 'assistant'` → `!== null`) | user-negatives RED, **capability GREEN** | 🔴 3 RED — all three "user" tests; capability stayed green |
| **M3** loosen the IDENTITY conjunct (`=== 'top_bar_menu'` → `!== null`) | foreign-surface RED | 🔴 2 RED (mine + the sibling's own foreign test) |
| **M4** fabricate a rationale in the copy | copy test RED | 🔴 1 RED, only that test |
| **M5** drop the `aria-label` | AT test RED | 🔴 1 RED, only that test |
| **M6a** delete the `.overlayOriginBadge` CSS rule | style guard RED | 🔴 1 RED |
| **M6b** colour the TEXT instead of the border | style guard RED | 🔴 1 RED |
| **M7** latch the origin so the badge OUTLIVES the fact | takeover test RED | 🔴 1 RED |
| **M8** drop `showMenu &&` in TopBar | — | 🟢 **DEMONSTRATED EQUIVALENT** |

**M2 + M3 are the discriminating pair (trap 19).** M2 loosens the guard for *all* origins and REDs the user tests while the assistant test stays **green** — sensitivity to the *named* origin, not merely to *something*. M3 loosens it for a *different surface* and REDs the foreign-surface test. Neither alone shows binding; the pair does.

**M8 is demonstrated equivalent, not asserted equivalent** (trap 13c: a survivor is a claim either way). Reason: the badge sits structurally inside `{isOpen && …}`, so `showMenu &&` in TopBar is defence-in-depth that makes the conjunction explicit at the reading site, not the load-bearing gate. Kept deliberately — `OVERLAY_SURFACE_IDS` has one member today, so `=== 'top_bar_menu'` and `!== null` agree on every value the enum can produce; the moment a second surface is lifted they stop agreeing.

## Gates

- `pnpm run typecheck` (`scripts/ci/typecheck-gate.sh`) — ✅ **PASSED**, coverage 3345/3385, ratchet 618 files / 2487 errors **== baseline 2487**. No `--update-baseline`.
- `npx eslint` on the three changed files — **0 errors**. The single warning (`TopBar.tsx:115`, unused `no-console` disable) is **pre-existing**: verified present at `60c4b082`.
- Neighbouring suites (`src/components/layout`, `src/stores`, `applyV5State.assistantSurfaces`, `OutputsDock.overlayOcclusion`) — ✅ **100/100**, no collateral.

## The other two menus — checked, and deliberately NOT lifted

`LeftSidebar`'s lens menu and `UserAvatarMenu` still hold their open-state in component-local `useState` and cannot be assistant-raised. **Lifting one is not small at this tip and I have not done it.** Both coordinate with the kebab over the `menu:exclusive` window event; lifting one means it must simultaneously honour the event *and* the one-slot store, and `uiStore`'s own header records that a second lifted surface immediately activates the raise-into-an-occupied-slot hazard `requestOverlaySurface` guards. `src/stores/uiStore.ts` is also explicitly **not this lane's file**. Converting that discovery into scope is this programme's most repeated failure, so it stays a finding.

### DIFF
--- a/src/components/layout/KebabMenu.tsx
+++ b/src/components/layout/KebabMenu.tsx
@@ -20,6 +20,7 @@
 import { useState, useCallback } from 'react'
 import {
   MoreVertical,
+  Sparkles,
   Pencil,
   Download,
   Upload,
@@ -37,8 +38,32 @@ import { useSettingsStore } from '../../canvas/settingsStore'
 import { useCanvasStore } from '../../canvas/store'
 import styles from './TopBar.module.css'
 
+/**
+ * The complete sentence this menu is allowed to make about its own origin.
+ *
+ * WHO, and HOW TO TAKE IT BACK — never WHY. There is no reachable reason at
+ * this tip: `ui_directive.note` is the only free-text carrier on the 0.39.0
+ * wire and nothing in src/ reads it, so any rationale rendered here would be
+ * invented. A fabricated "why" on the provenance channel is worse than no
+ * badge at all, because this is the one channel whose entire purpose is
+ * truthfulness. If a reason ever becomes available, it arrives as a new
+ * field and this constant grows a parameter — it does not get guessed.
+ *
+ * The dismissal hint is a VERIFIED fact, not reassurance: TopBar's Escape
+ * handler runs `closeMenu()` unconditionally on origin, so Escape genuinely
+ * takes the surface back from the assistant.
+ */
+export const OVERLAY_ATTRIBUTION_TEXT = 'Opened by Olumi'
+export const OVERLAY_DISMISS_HINT = 'Esc to dismiss'
+
 interface KebabMenuProps {
   isOpen: boolean
+  /**
+   * Who raised THIS menu — `'assistant'` only when the AI put it on screen.
+   * Null whenever the menu is the user's own (or not raised at all), so the
+   * attribution cannot outlive the fact it describes.
+   */
+  raisedBy?: 'user' | 'assistant' | null
   onToggle: () => void
   onClose: () => void
   onStartRename: () => void
@@ -49,6 +74,7 @@ interface KebabMenuProps {
 
 export function KebabMenu({
   isOpen,
+  raisedBy = null,
   onToggle,
   onClose,
   onStartRename,
@@ -129,7 +155,44 @@ export function KebabMenu({
         </Tooltip>
 
         {isOpen && (
-          <div className={styles.dropdownMenu} role="menu">
+          <div
+            className={styles.dropdownMenu}
+            role="menu"
+            // The visible badge below is a plain <div>: inside `role="menu"`
+            // only menuitems are reliably announced, so a screen-reader user —
+            // the person MOST disoriented by a menu that appears unprompted —
+            // would get nothing. The menu's accessible NAME carries it instead,
+            // and is left undefined (falling back to the trigger) when the user
+            // opened the menu themselves, so nothing is announced that is not
+            // true.
+            aria-label={
+              raisedBy === 'assistant'
+                ? `More options — ${OVERLAY_ATTRIBUTION_TEXT}`
+                : undefined
+            }
+          >
+            {/* PROVENANCE, not decoration. A surface the assistant raised must
+                say so, inside the surface itself — the answer sits where the
+                question ("why did this open?") is asked, rather than somewhere
+                the user has to hunt for it. Deliberately non-interactive, no
+                animation, no focus steal, no colour alarm: a user who did not
+                notice the menu open should not be startled by the explanation
+                of it. */}
+            {raisedBy === 'assistant' && (
+              <div className={styles.overlayOriginRow} data-testid="overlay-origin-badge">
+                <span className={styles.overlayOriginBadge} data-testid="overlay-origin-label">
+                  <Sparkles size={11} aria-hidden="true" />
+                  {OVERLAY_ATTRIBUTION_TEXT}
+                </span>
+                <span
+                  className={styles.overlayOriginHint}
+                  data-testid="overlay-origin-dismiss-hint"
+                >
+                  {OVERLAY_DISMISS_HINT}
+                </span>
+              </div>
+            )}
+
             {/* Decision group */}
             <div className={styles.dropdownMenuLabel}>Decision</div>
             <button
--- a/src/components/layout/TopBar.module.css
+++ b/src/components/layout/TopBar.module.css
@@ -291,6 +291,50 @@
   color: var(--text-light, #6E6B6B);
 }
 
+/* ── Attribution for an assistant-raised menu ──────────────────────────────
+   Shown ONLY while this menu is on screen because the assistant put it there.
+   Calm by construction: no animation, no fill, no focus, nothing interactive —
+   a user who did not notice the menu open must not be startled by the
+   explanation of it, while a user who wonders "why did that open?" finds the
+   answer at the top of the thing that opened.
+
+   DS v5 §3.2 / DESIGN_SYSTEM.md: pills are OUTLINED only —
+   `bg-transparent border-{colour}/30 text-text-body`. The colour is carried on
+   the BORDER; the ink stays --text-body. Colouring the text is what produced a
+   sibling lane's 2.80:1 and 2.83:1 failures against WCAG AA. Measured here:
+   --text-body #3F3F3E on the menu's white surface = 10.5:1, and the hint's
+   --text-light #6E6B6B = 5.1:1 at 12px (both AA; --text-light is already the
+   audited choice for .dropdownMenuLabel in this same menu). */
+.overlayOriginRow {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 8px;
+  padding: 6px 10px 8px;
+  margin-bottom: 2px;
+  border-bottom: 1px solid var(--border-default, #EEE6D8);
+}
+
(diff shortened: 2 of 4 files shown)


Talchain:
## SUPERSEDED — closing. The capability shipped, on the surface this PR's own body identified as the better target.

Re-derived at UI staging tip `37f626b6` (17 Aug).

**The attribution capability is live.** `src/canvas/components/AssistantOpenedNotice.tsx` carries `NOTICE_TEXT = 'Opened by Olumi'` and is mounted at `src/canvas/components/OutputsDock.tsx:2391` (`{effectiveIsOpen && <AssistantOpenedNotice />}`, imported at `:144`), with `OutputsDock.assistantOpenedAttribution.spec.tsx` pinning it. That is the `open_panel` / `open_section` path — the one this PR's body called out as *"the reachable instance of exactly the harm this lane was scoped to fix"*, deliberately left to other lanes. Those lanes took it.

**This PR's own target seam is still dark.** `requestOverlaySurface` has **zero production call sites** at the tip — every reference is `src/stores/uiStore.ts` (declaration + impl) or a test (`uiStore.overlaySurface.test.ts`, `TopBar.overlaySurface.spec.tsx`, `applyV5State.assistantSurfaces.test.ts`, `OutputsDock.assistantOpenedAttribution.spec.tsx`). Contrast control, same sweep, same globs: `forceActivateOutputTab` = **6 production files**. So the probe discriminates, and the answer is real absence, not blindness. `data-overlay-origin="assistant"` therefore remains unreachable from the wire, exactly as this PR reported at `60c4b082` — the schema still has no overlay verb and no overlay target kind.

**Why that means closing rather than merging.** Merging would add a TopBar/KebabMenu badge that cannot render on a real turn: display polish on a seam no producer can feed, which is the dark-launch pattern the programme's standing rules refuse. The honesty work this PR did — copy pinned as a literal, the badge asserted to contain nothing else, the accessible name carrying it inside `role="menu"`, the refusal to synthesise a reason from `ui_directive.note` — is worth reusing verbatim if and when the overlay verb is ratified into `olumi-schemas` with a CEE emit and an `applyV5State` dispatch. That is the enabling change this PR correctly declined to make, and it remains unmade.

Nothing is lost: the branch keeps the work, and the user-visible half of the capability is already in front of users on the surface that can actually reach it. Closed as superseded.

PR #662, opened 2026-08-10
F1 draft-state truth: keep a rendered model through unproven terminal failure; honest staged progress

Talchain:
## Pillar

PR1 / acceptance sentence 1 — Olumi makes clear what it used, inferred and left out, including about its own state. This is the first thing anyone opening a link experiences.

## The two measured defects

**DEFECT 1 — the 504 lie after a real save** (deployed staging, 8 Aug; `olumi-poc-independent-review-2026-08-09.md`): GRAPH_READY at 96.981 s (21n/43e, committed server-side), terminal COMPLETE 504 UPSTREAM_TIMEOUT at 125.202 s — the UI removed the rendered graph and said no model was drafted, while a grounded follow-up found the canonical model under the same scenario.

**DEFECT 2 — the silent wait.** Derived at `60c4b082` first (per brief): drafting narration (PROGRESSIVE_STAGES), render-at-GRAPH_READY, settling narration (SETTLING_STAGES) and the no-GRAPH_READY 500 failure report (CEE recovery + retry, tested against the live 8 Aug capture) **already exist**. The genuine residual gap: COACHING_READY was never consumed — the copy kept claiming "waiting on the values and coaching" after the coaching pass had landed (~28 s on the measured journey).

## What changed

**Terminal-failure reconciliation** (`consumeStreamedDraftTurn.ts`, `useConversation.ts`):
- `discardPreview` renamed `terminalError` — the old name was an instruction the caller obeyed unconditionally.
- Removal now happens ONLY when the terminal payload **proves** the commit failed (`draft_graph_commit_failed` — CEE route-v2's "Nothing was written" envelope, fresh-journey P0 diagnosis §1.4). The client cannot poll server state (no status route; guest RLS), so the payload's own commit marker is the cheapest authoritative check it holds.
- Every other ≥400 after GRAPH_READY **keeps the model**, marks the draft `unsettled` (run gate shut; `shouldPersistGraphForScenario` keeps the UI from ever persisting it), renders ONE honest notice — `DRAFT_FAILED_MODEL_KEPT_NOTICE`: *"Drafting failed after your model reached the canvas, so its coaching and settled values never arrived. The structure you can see is real, but we can't confirm from here whether this draft saved. Start a new draft to get a model with settled values."* — with the Start-a-new-draft chip (the affordance that genuinely works; F3), and suppresses the generic failure bubble + send-failure banner whose claims the held frame refutes. The user bubble reads delivered: GRAPH_READY is proof the server processed the message.
- The no-GRAPH_READY 500 class is untouched: plain failure + CEE recovery + retry chip (already live and pinned by `useConversation.streamedDraft500Recovery.spec.tsx`).

**Honest staged progress** (`draftStore.ts`, `AIInputBar.tsx`, `DraftLoadingAnimation.tsx`):
- COACHING_READY now sets an identity-guarded store flag through the real stream; the composer switches to `SETTLING_AFTER_COACHING_STAGES` ("Your model is on the canvas. Waiting for the settled values…") — it stops claiming coaching is outstanding without claiming its content arrived (the enum includes `failed_degraded`). No fabricated percentages, no time estimates.
- All new copy governed by `narrationHonesty.invariant.spec.ts` via the derived `*_NOTICE` / `NARRATION_TABLES` manifests.

## Evidence

- **RED-first**: `useConversation.streamedDraftTerminalError.spec.tsx` failed 12/15 at pristine `60c4b082` (kept-model, unsettled-phase, single-notice, delivery and no-banner assertions all red; the commit-failed removal guard green then and now). Real SSE frames through the real stream parser; only network seams mocked.
- **Mutants** (throwaway worktree outside the repo root, isolation proven by sentinel write + differing inodes, HEAD-relative restores, applied-check scoped to src/ with control asserting exactly zero, trailing control green; 108 tests collected every run): 10/10 BITTEN — including the discriminating pair `predicate-always-true` (kept-model tests red) vs `predicate-always-false` (commit-failed removal red), proving the reconciliation binds to the commit-failure marker in both directions, and `fix-revert-always-discard` (12 red) proving the suite sees the original defect.
- **Gates at tip**: `pnpm run typecheck` PASSED (ratchet 2487 = baseline) · `pnpm run lint` PASSED · conversation+v5+stores+components dirs: 373 files / 4922 tests passed, exit 0.

Do NOT merge — adversarial review to follow.

### DIFF
--- a/src/canvas/components/AIInputBar.tsx
+++ b/src/canvas/components/AIInputBar.tsx
@@ -23,7 +23,11 @@ import {
   type AddOptionCanvasTargets,
   type AddOptionChange,
 } from '../conversation/addOptionRequest'
-import { messageForElapsed, messageForSettling } from './DraftLoadingAnimation'
+import {
+  messageForElapsed,
+  messageForSettling,
+  messageForSettlingAfterCoaching,
+} from './DraftLoadingAnimation'
 import { useDraftStore, draftStreamPhaseFor, draftStreamInFlight } from '../stores/draftStore'
 
 export type AIInputBarVariant = 'strip' | 'docked-tab' | 'floating' | 'first-use' | 'welcome'
@@ -176,6 +180,12 @@ export const AIInputBar = memo(
     const currentScenarioId = useCanvasStore((s) => s.currentScenarioId)
     const draftStreamPhase = useDraftStore((s) => draftStreamPhaseFor(s, currentScenarioId))
     const isSettling = draftStreamPhase === 'settling'
+    // F1 (honest staged progress): once the owning turn's COACHING_READY frame
+    // has landed, the settling copy must stop claiming coaching is
+    // outstanding. Scoped by the SAME ownership derivation as the phase — the
+    // flag is only meaningful while this scenario's own draft is settling, so
+    // it is read as a conjunct of `isSettling`, never bare.
+    const coachingLanded = useDraftStore((s) => s.draftStreamCoachingLanded)
     const isGenerating = isThinking && (nodeCount === 0 || isSettling)
 
     // ── ROADMAP 2.134: THE STOP CONTROL ─────────────────────────────────────
@@ -207,7 +217,13 @@ export const AIInputBar = memo(
       // lands: the settling table's thresholds are measured from the render, not
       // from the start of the turn. Sharing the turn's clock would put the
       // escalated settling line up immediately on every single draft.
-      const resolve = isSettling ? messageForSettling : messageForElapsed
+      // `coachingLanded` likewise restarts it when COACHING_READY lands — its
+      // table is licensed by that frame, and its own clock starts with it.
+      const resolve = isSettling
+        ? coachingLanded
+          ? messageForSettlingAfterCoaching
+          : messageForSettling
+        : messageForElapsed
       if (!isGenerating) {
         setGeneratingMessage(resolve(0))
         return
@@ -218,7 +234,7 @@ export const AIInputBar = memo(
         setGeneratingMessage(resolve(Math.floor((Date.now() - start) / 1000)))
       }, 1000)
       return () => window.clearInterval(id)
-    }, [isGenerating, isSettling])
+    }, [isGenerating, isSettling, coachingLanded])
 
     useImperativeHandle(
       ref,
--- a/src/canvas/components/DraftLoadingAnimation.tsx
+++ b/src/canvas/components/DraftLoadingAnimation.tsx
@@ -121,6 +121,29 @@ export const SETTLING_STAGES = [
 ] as const
 
 /**
+ * The post-COACHING_READY table (F1 — honest staged progress). Shown only when
+ * the client HOLDS a COACHING_READY frame on top of the GRAPH_READY one, i.e.
+ * the coaching pass has landed and the turn is still running (live median:
+ * a ~1.7 s window, 59.2 s → 60.9 s — but the 8 Aug measured journey sat in it
+ * for ~28 s, reading "waiting on the values and coaching" about a coaching
+ * pass that had already run).
+ *
+ * ── WHAT THE FRAME LICENSES, AND WHAT IT DELIBERATELY DOES NOT ────────────
+ * COACHING_READY carries only a `coaching_status` ENUM — derived at the CEE
+ * producer: 'complete' | 'partial' | 'failed_degraded' — never the coaching
+ * itself, which arrives only in the terminal payload. So this line may STOP
+ * claiming coaching is outstanding (the pass has landed, whatever its status),
+ * but it may NOT claim coaching "has arrived" or "is ready" — for the
+ * 'failed_degraded' status that would be false, and the client does not
+ * word-switch on the enum because a status-keyed claim about content it has
+ * not seen would still be a guess. Dropping the claim is the only wording
+ * true for every enum value.
+ */
+export const SETTLING_AFTER_COACHING_STAGES = [
+  { afterSeconds: 0, message: 'Your model is on the canvas. Waiting for the settled values…' },
+] as const
+
+/**
  * The terminal honest lines for a draft whose values will NOT settle in this
  * session (`draftStreamPhase === 'unsettled'`). Two, because there are two
  * genuinely different causes and one sentence cannot state both truthfully.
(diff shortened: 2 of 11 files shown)


Talchain (commented):
**ROUND-1 ADVERSARIAL REVIEW — VERDICT: APPROVE** (with two required follow-up rows, neither introduced by this PR). Same-account PR, so this is posted as a comment; treat it as the independent review of record.

Reviewed at head `f8288f50dddb1eedfb3377afdfb9ff912bf2fd07` (asserted after fresh blobless clone; merge-base `60c4b082`). All commands run with `VITE_SUPABASE_URL`/`VITE_SUPABASE_ANON_KEY` exported; the PR's six spec files collected by name: 15 + 44 + 20 + 21 + 8 + 55 = 163 tests, all green.

---

## 1. The predicate's domain — terminal-shape table, derived at the CEE producer's bytes (route-v2.ts @ `f9ce8c90`, draft-graph-dispatch.ts)

Structural derivation first: `dispatchDraftGraph` has TWO try blocks. `handleDraftGraph` (the pipeline, which streams GRAPH_READY) throws from the FIRST; the commit (`commitDirectAnswer`) lives in the SECOND, whose catch returns `commitPerformed: false` → the route's `turn_commit / draft_graph_commit_failed` 500. Every route-authored draft failure is HTTP **500** (`reply.code(500)` on both paths) — so a **504** terminal is by construction middle-box-authored, which is exactly the measured 8 Aug class where the commit survived.

| Terminal shape after GRAPH_READY | Author | Commit truth at producer bytes | PR behaviour | Copy-truth verdict |
|---|---|---|---|---|
| 504 `UPSTREAM_TIMEOUT` envelope | gateway/middle box (CEE route draft errors are always 500) | Unknowable client-side; measured-common committed | keep + unsettled + notice | TRUE |
| 504 `PROXY_UPSTREAM_TIMEOUT` body | Netlify/Render proxy | Unknowable | keep + unsettled + notice | TRUE |
| 500 `turn_commit` / `draft_graph_commit_failed` | CEE (`!dg.commitPerformed`) | Server's own "nothing was written" | remove + plain failure + retry chip | TRUE — removal states the truth |
| 500 `draft_graph_pipeline` / any of the ten mapped reasons (`draft_graph_cee_graph_invalid`, `_cee_timeout`, `_cee_llm_validation_failed`, `_cee_llm_upstream_error`, `_cee_rate_limit`, `_cee_internal_error`, `_cee_validation_failed`, `_cee_service_unavailable`, `_pipeline_status_<n>`, `_pipeline_threw`) | CEE route catch | **NOT committed at current bytes** — the throw precedes the commit try; CEE itself marks `draft_graph_pipeline_threw_after_preview` + `draft_loss` disclosure | keep + unsettled + "can't confirm … saved" | **UNDER-CLAIMS** (see F2) — honest wording, weaker than derivable |
| 200, no extractable graph, preview standing (F4) | any | n/a | unsettled, cause `stream_loss` → "connection dropped" copy | PRE-EXISTING minor mislabel (F3) |
| ≥400 with no GRAPH_READY | any | n/a | unchanged plain failure path | TRUE |

GRAPH_READY-then-validation-failure **is wire-possible** — route-v2's own comment block names "CEE_GRAPH_INVALID after GRAPH_READY has already streamed" as a lived case, and `graphPreviewEmitted()` exists precisely to disclose it.

**F2 (follow-up row, cross-service — the sharpest thing in the domain).** For the whole `draft_graph_pipeline` 500 family the client HOLDS a payload that, at the producer's current bytes, implies non-commit — CEE records it as draft loss and the scenario's NEXT turn will tell the user "the graph you saw was never saved", while this PR's notice said "we can't confirm from here whether this draft saved". The kept model in that class is a gated, unpersisted phantom until the next turn/reload resolves it. I judge the PR's fail-closed choice CORRECT anyway: distinguishing that class client-side means mirroring CEE's internal throw-vs-commit ordering (trap 12 — and the drift direction of that mirror is *destroying committed models*, the exact P0 this PR fixes). The honest fix is a **wire contract**: CEE stamps an explicit commit marker (e.g. `details.commit_performed: boolean`) on every terminal draft envelope, and the client removes on proof. Row it; do not solve it in this PR.

## 2. Reload adjudication

- Server committed (504 class): reload hydrates the committed settled graph via `loadScenario` — strictly better than the notice's expectation; no contradiction (the notice is dated conversation history; `draftStore` is memory-only, so no sticky `unsettled` survives reload to gate the good graph).
- Server did not commit (pipeline-500 class): reload shows an empty canvas; the notice's "can't confirm … Start a new draft" stays coherent, and CEE's own next-turn draft-loss disclosure completes the story.
- `start_new_draft` chip mints a FRESH scenario (chipDispatch.ts states this and why `retry` is not rendered), so the guidance cannot collide with CEE's continuation guard or a committed graph. Coherent in both worlds. No finding.

## 3. The unsettled gate — verified by execution

- Run gate: `canRunAnalysis` refused with the phase `unsettled` (PR spec, executed green).
- Cause not sticky: `unsettledCause` is a per-call return value; the store flag resets on `drafting`/`idle` (M5/M4 mutants below); happy-path terminal ingest releases to `idle` with the coaching flag cleared (executed).
- **F1 (follow-up row, the one real defect found — PRE-EXISTING, not this PR's):** "graph never persisted by the UI" is true at the guarded chokepoint and FALSE at one bypass. `useScenario.ts`'s unmount best-effort flush (~line 449) calls `scenarioService.saveGraphViaGatedPath` DIRECTLY, skipping `persistGraphNow` and its `shouldPersistGraphForScenario` guard — and `graphSaveTimerRef` is never nulled after firing, so the flush triggers on unmount whenever a save was EVER scheduled. Proven by execution (probe spec, signed-in harness, real draftStore): with the phase `unsettled`, the debounced autosave was correctly SUPPRESSED, then unmount wrote the kept-unsettled nodes to `scenarios.graph` — guard never consulted. Severity: a signed-in user who navigates away after a kept-model terminal error writes the structure-only preview over CEE's committed settled graph. Pre-existing (file untouched by this PR; at merge-base the same flush wrote the post-discard EMPTIED canvas over the commit, which is worse) — so this PR strictly improves the axis and is not blocked on it. Fix in a follow-up: route the unmount flush through `persistGraphNow`, and null the timer ref on fire/clear. Note `runGateCallSites.derived.spec.ts` asserts the guard's presence in `persistGraphNow` but cannot see this fourth call site (trap 13b's shape).

## 4. Trap 3b — mount path

Deployed posture derived at this head: `netlify.toml` `[context.staging.environment] VITE_FEATURE_AI_PANEL_V2 = "true"` AND `flags.ts` default ON → ChatThread (via ConversationPanel) is the mounted thread. The notice binds UPSTREAM of the flag fork (the shared `messages` array feeds both surfaces), so a flag flip cannot strand the capability the way 2.466/2.491 were stranded; the DOM half targets the deployed surface. Residual, non-blocking: no spec DERIVES the posture (the estate's `readFileSync(netlify.toml)` technique is used elsewhere, e.g. compareChangeCoherence.spec) — a one-line derived assert would make the ChatThread binding fail loud if the flag ever moves; and the emergency-rollback surface (DraftChat) does not obviously render `actionChips`, so under rollback the notice text flows but the chip may not. Observation only.

## 5–6. Mutants — run independently in an isolated worktree

Worktree at `/private/tmp/pr662-mutants-*` (outside the repo root, detached at the head SHA); isolation PROVEN by write: sentinel appended in the worktree did not appear in the clone, inodes distinct (`583401371` vs `583640710`); deps installed and resolved inside the worktree; applied-check per mutation scoped to `src/` asserting exactly one named file; HEAD-relative restores; clean-tree assert before/after each; trailing pristine control 100/100 GREEN.

9/9 bitten, each on the named assertion:
- M1 predicate never proves (`return false`): 3 RED (commit-failure recognition + removal pin).
- M2 predicate always proves (`return true`): 12 RED (keeps + proves-nothing corpus).
- M3 drop `onCoachingReady` wiring in useConversation: 1 RED — the real-stream seam test (the consumer hook pre-existed unused at merge-base; the PR's caller wiring is what this pins).
- **M4/M4b — the disclosure item, run in BOTH directions:** guard→always-records REDs "IGNORES a frame from a turn that does not own the phase"; guard→never-records REDs "records the landing for the OWNING turn only" (+ the survives-unsettled pin). Different assertions in each direction — the draftStreamOwnership.spec pair genuinely discriminates; the lane's honesty disclosure about its cruder M8 application is accepted.
- M5 flag survives new drafting turn: 1 RED (the leak pin).
- M6 cause always `stream_loss`: 7 RED (notice keying, delivery, banner).
- M7 delivery proof dropped: 2 RED (user bubble DELIVERED, both payload shapes).
- M8 FIX REVERT (`terminalErrorKeepsModel = false`): 12 RED across both spec files — reverting the fix turns the suite red (trap 11 satisfied); matches the lane's claimed 12-failed pristine RED.

## 7. Gates at the head + CI

- `pnpm typecheck`: PASSED — coverage 3345/3385 loaded, ratchet 2487 = baseline 2487 exactly.
- `pnpm lint`: exit 0 — eslint clean; rules-of-hooks ratchet 232 exactly at baseline.
- Wide suite (`npx vitest run`, full, clean tree — probe removed first): 1590/1594 files passed, **23,096 tests passed, 2 failed, 66 skipped**. The 2 reds are both wall-clock perf budgets in `EdgeEditPopover.perf.spec.tsx` ("within 16ms") — a file this PR does not touch — measured while my mutant kit was competing for CPU (49.9 ms observed under load). Re-run in isolation at the head: **15/15 GREEN**, renders at 2–5 ms. Adjudication: load-induced flake, zero PR-attributable failures; corroborated by CI's own 4-shard Full Test Suite green on this exact SHA.
- CI check-runs queried for the EXACT SHA `f8288f50…`: 13 total, `status=completed` on all 13, running=0, zero non-success conclusions (incl. Staging Gate, Full Test Suite 4 shards + summary, TypeScript + Lint, Production Build, Security Audit, CEE Contract Validation). `mergeable_state: clean`.

## Diff surprises

None — every hunk maps to the described fix. Two notes: (a) the commit-failed class marks the user bubble `failed`/`unconfirmed` via the pre-existing typed-error path, which is honest (the turn genuinely did not commit); (b) `terminalProvesCommitFailed`'s four-container duck-walk cannot false-positive from the route's `postStageExtras` merge (the allowlist carries no `reason` key), and the `turn_commit` envelope carries no extras — so the dangerous direction (wrongly removing) is not constructible from the current producer.

**Follow-ups to row (both outside this PR's scope): (1) unmount-flush persistence bypass in useScenario.ts — the fix is small and should land promptly; (2) CEE explicit commit-state marker on terminal draft envelopes, retiring the pipeline-500 under-claim; (3, minor) the F4 200-no-graph class inherits "connection dropped" copy that is false for a delivered 200 — a fourth cause string when convenient.**

PR #663, opened 2026-08-10
UI: retire the win-frequency gap across the results panel + two persistence truth fixes

Talchain:
Truth fixes established by executed evidence (Codex audit, 10 Aug; the PR #662 review). **Do not merge without adversarial review.**

## THE CLASS — no user-facing surface states the percentage-point gap between win frequencies

A difference of two Monte-Carlo estimates carries more uncertainty than either estimate does; printed as a bare integer with no interval it reads as the most precise number on screen while being the least reliable. **The option's OWN probability is the statistic.** Four surfaces stated it; all four are closed.

| Surface | Was | Now |
|---|---|---|
| `buildV7Headline` / `V7Hero` | "Leads by 40 points" under a correct own-probability headline | names the runner-up and states **its own** probability; silent when the headline has no magnitude, no rival is measured, or two rivals **tie** at the top |
| `OptionCards` (2 arms) | "Behind by N percentage points" | **deleted** |
| `OptionNode` (canvas) | "Close call: within N percentage points" | "Close call with the leading option" — signal kept, quantity gone |
| `certaintyCopy` (2 sites) | "{winner} currently leads **by N points**" / "{winner} leads **by N points**" | **deleted**; the softening survives |

**Licence to delete, derived at each seam — never assumed.** `OptionCards`: both arms require `option.winProbability != null`, the *same* condition that renders the card's `win-pct-{id}` readout, so the own probability is already there. `certaintyCopy`: the panel already shows the winner's own probability (the ring's `overrideScore`, captioned `COMPARATIVE_COPY.label`) **and** `aheadHeadline`'s own F4 note had already adjudicated this question — a magnitude-free comparative sentence is the correct behaviour here, so replacing the gap with a magnitude would contradict a ruling recorded in the file being changed.

**Every multi-site surface changed all its sites together.** `OptionCards` runner-up + other-non-winner; `certaintyCopy` softened + confident. A half-fix leaves the quantity on screen for whichever users hit the other branch.

**`OptionNode`:** naming the leader is entitled — `closeCallGapPp` returns null unless `verdict.hasLeadingOption`. The 5pp window and 1pp floor are untouched, so the qualifying-run set is provably unchanged; only the sentence moved.

### ⚠ Correction to an earlier claim in this PR
Round 2 reported `certaintyCopy`'s suffix as *"still on screen"*. **That was an overclaim** — a live rendering consumer in *code* generalised to a live *rendered* surface without deriving the deployed flag posture. `decisionConfidenceElement` is referenced at exactly one site (`ResultsBody.tsx:508`), inside `{!isAnalysisHeroPanelEnabled() && …}`, and `netlify.toml` bakes `VITE_FEATURE_ANALYSIS_HERO_PANEL="1"`. **The panel mounts on the flag-OFF arm only, so that suffix was never on the surface staging serves.** The fix is defence in depth on live code one flag move from every user, and its test is labelled as such rather than dressed up as a live-surface guard.

### The durable guard
`ResultsBody.winFrequencyGapAbsence.spec.tsx` asserts **all three** retired forms absent across everything `ResultsBody` renders, with a **positive control** injecting the phrase through a real render input (an option label) proving the probe can *see* it. Plus a derived sweep in `certaintyCopy.spec.ts`: no gap value — 12 magnitudes x 4 tiers x 3 readiness states, with a count assertion so the loop cannot pass by running zero times — emits a point-gap clause on **any** branch.

Two preconditions in the flag-OFF case earned their place **by failing first**: the panel must be MOUNTED, and the run must be ON the softened branch. The fixture carried no `verdict`, so `NO_CLAIM_VERDICT` returned "the analysis did not put an option forward" — the absence assertion would have passed against a headline that never mentions a leader.

### Producers
- **`decisionVerdict.gapPp` — render-dead** (zero production readers; only two test files).
- **`certaintyCopy.winProbabilityGap` — now render-dead too**: still computed and passed, no longer read. Deleting both is a clean follow-up, deliberately not folded in.

---

## FIX 2 — freshness discarded CEE's verdict at an equal timestamp
`computed_at <= prev.computedAt` returned the previous object, but `computed_at` stamps the **analysis**, not the verdict: CEE re-evaluates against the current graph and re-sends the same analysis with a new verdict and hash. Witnessed on staging — CEE returned `stale` with a new hash, the store kept `fresh` with pre-edit hashes. Now rejects only **strictly** older; `sameVerdict` no-ops genuine echoes.

## FIX 3 — the unmount graph flush bypassed the persistence gate
`persistGraphNow` declares itself *"the ONE write code path"*; the unmount cleanup called `saveGraphViaGatedPath` directly, so a signed-in user navigating away wrote a kept-unsettled preview **over CEE's committed graph**. Now routed through the choke point. **Second limb:** `graphSaveTimerRef` was cleared but never nulled, so an already-**fired** debounce read as "pending" — nulled at fire time and after the flush.

---

## Evidence

**RED-first per surface**, measured before each source edit, by name.

**Mutants** — throwaway worktree outside the repo root; isolation proven by **inode comparison + sentinel write test**; restores from a pristine archive outside both trees; clean-tree assertion before *and* after each; applied-check scoped to `src/`; **collected-count asserted every run**; trailing control.

| Round | Mutant | Result |
|---|---|---|
| 1 | `M1` hero gap restored | 9 RED incl. both mount postures |
| 1 | `M5` weakest rival named | **1 RED — the identity test only** (pair with `M1`) |
| 1 | `M6` tie guard deleted | 1 RED |
| 1 | `M2` `<` → `<=` | **RED equal+changed / GREEN equal+identical** |
| 1 | `M3`/`M4`/`M7` | RED (c) / RED (e) / RED (b)(c)(e) — reproducing the pristine signature |
| 2 | `A1` behind-by, **runner-up arm only** | RED that arm; **other arm GREEN** |
| 2 | `A2` behind-by, **other arm only** | RED that arm; **runner-up GREEN** |
| 2 | `A3` own-% readout deleted | RED the identity pin; **absence guard GREEN** |
| 2 | `B1`/`B2` node quantity / window | RED node tests only / RED window test only |
| 2 | `C1` hero regression | RED hero + guard; no card/node test trips |
| 3 | `D1b` soft site restored | 14 RED — certaintyCopy, caveatGuarantee, the flag-OFF guard |
| 3 | `D2b` confident site restored | 18 RED — certaintyCopy, caveatGuarantee, HeroFooterAlignment |

`A1`/`A2` and `D1b`/`D2b` prove each multi-site surface is pinned **per site**. `A3` proves the absence guard and the identity pin measure different things. **No surface's mutant trips another surface's tests.** `D1b`/`D2b` were re-cut after the first attempt failed by *throwing* rather than emitting the clause — `referenceErrors=0` is asserted, because a mutant that fails for the wrong reason proves nothing.

**Gates at `cef39dba`**, in the required check's step order (`lint → typecheck → tests`): lint **0 errors**, hooks ratchet 232/15 exactly at baseline · typecheck **PASSED**, coverage 3347/3387, ratchet **2487 = baseline 2487** · selftest **18/18** · build OK · `src/components/results` **252 files / 4,471 tests** green; earlier rounds swept `src/canvas` + `src/hooks` (**1,115 files / 16,412 tests**).

**CI on the exact head: 13/13 green**, including the required Staging Gate and all four shards.

## Scope
20 files — **6 source, 14 spec**. Nothing else. The Research CTA was checked and is **already retired** (ROADMAP 2.816, `researchCtaRetired.spec.tsx` pins both deployed arms) — the brief's premise was stale; no change needed or made.

### DIFF
--- a/src/canvas/nodes/OptionNode.tsx
+++ b/src/canvas/nodes/OptionNode.tsx
@@ -432,7 +432,16 @@ export const OptionNode = memo((props: NodeProps) => {
   const viewMode = useCanvasStore(state => state.viewMode)
   const isDetailed = viewMode === 'expert'
 
-  // Graph v2 Task 4: close-call gap in percentage points. Non-zero only when
+  // Graph v2 Task 4: close-call gap in percentage points.
+  //
+  // ⭐ PREDICATE ONLY SINCE 2026-08-10 — the number is NEVER RENDERED. It
+  // decides two things: whether the qualitative "Close call" marker shows, and
+  // whether the extra "What would change this?" chip is offered. The 5pp
+  // window and the 1pp floor are kept exactly as they were, so the set of runs
+  // that qualify is provably unchanged by this retirement; only the sentence
+  // moved. Do not reintroduce it into copy.
+  //
+  // Non-zero only when
   // this option is a non-leader within 5pp of the leader's win probability.
   // Prefers report.robustness.recommended_option_id; falls back to max scan if
   // missing. Returns null when not in close-call territory or pre-analysis.
@@ -1248,10 +1257,26 @@ export const OptionNode = memo((props: NodeProps) => {
 
         {/* Graph v2 Task 4: close-call line — only when within 5pp of leader.
             Renders ABOVE the existing "Behind:" reason so users see both the
-            closeness signal and the causal explanation. */}
+            closeness signal and the causal explanation.
+
+            ⭐⭐ THE QUANTITY IS RETIRED (2026-08-10); THE SIGNAL IS NOT.
+            This read "Close call: within N percentage points" — the
+            percentage-point difference between two Monte-Carlo win
+            frequencies, which is the banned statistic: less reliable than
+            either estimate it is built from, yet printed as a bare integer
+            with no interval. The tie-ness signal it carried is genuinely
+            useful and stays, as a QUALITATIVE marker with no number.
+
+            Naming the leading option is entitled here and nowhere borrowed:
+            `closeCallGapPp` returns null unless `verdict.hasLeadingOption`,
+            so this line cannot render on a turn where the producer withheld
+            the designation. And the reader is not left without a number — the
+            node states this option's OWN win probability directly above
+            (`COMPARATIVE_COPY.phrase`), which is the statistic the ratified
+            rule licenses. */}
         {closeCallGapPp != null && (
           <p className={`${typography.nodeLabel} text-warning mt-0.5 m-0`}>
-            Close call: within {closeCallGapPp} percentage point{closeCallGapPp !== 1 ? 's' : ''}
+            Close call with the leading option
           </p>
         )}
 
--- a/src/canvas/nodes/__tests__/render-matrix.spec.tsx
+++ b/src/canvas/nodes/__tests__/render-matrix.spec.tsx
@@ -439,7 +439,7 @@ describe('Render matrix — OptionNode × view × phase', () => {
     },
   })
 
-  it('Standard post non-leader, gap 3pp: shows "Close call: within 3 percentage points" and keeps "Behind:" line', () => {
+  it('Standard post non-leader, gap 3pp: shows the qualitative close-call marker and keeps "Behind:" line', () => {
     applyStore(closeCallTopology(3))
     vi.mocked(useNodeDisplayMetadata).mockReturnValue({
       sensitivityRank: null,
@@ -452,11 +452,16 @@ describe('Render matrix — OptionNode × view × phase', () => {
       isResultsMode: true,
     } as any)
     renderOption({})
-    expect(screen.getByText('Close call: within 3 percentage points')).toBeDefined()
+    // ⭐ SUPERSEDED 2026-08-10: was 'Close call: within 3 percentage points'.
+    // The tie-ness SIGNAL is valuable and stays; the percentage-point gap is
+    // the banned statistic and is gone. The node already states this option's
+    // own win probability directly above.
+    expect(screen.getByText('Close call with the leading option')).toBeDefined()
+    expect(screen.queryByText(/percentage point/i)).toBeNull()
     expect(screen.getByText(/Behind:/)).toBeDefined()
   })
 
-  it('Standard post non-leader, gap 1pp: pluralisation drops the trailing "s"', () => {
+  it('Standard post non-leader, gap 1pp: the marker still fires at the narrow end of the window', () => {
     applyStore(closeCallTopology(1))
     vi.mocked(useNodeDisplayMetadata).mockReturnValue({
       sensitivityRank: null,
@@ -469,12 +474,17 @@ describe('Render matrix — OptionNode × view × phase', () => {
       isResultsMode: true,
     } as any)
     renderOption({})
-    expect(screen.getByText('Close call: within 1 percentage point')).toBeDefined()
+    // ⭐ SUPERSEDED 2026-08-10: this asserted the singular 'point' form. With
+    // no number rendered there is no pluralisation left to pin — what remains
+    // worth pinning is that a 1pp gap is still INSIDE the close-call window.
+    expect(screen.getByText('Close call with the leading option')).toBeDefined()
+    expect(screen.queryByText(/percentage point/i)).toBeNull()
   })
 
-  it('Standard post non-leader, sub-percent gap (0.4pp): floor at 1pp, never reads "0 percentage points"', () => {
-    // Gap of 0.004 → Math.round → 0; floor clamps to 1pp so the line never
-    // reads the awkward "within 0 percentage points" phrasing.
+  it('Standard post non-leader, sub-percent gap (0.4pp): still inside the window, and states no quantity', () => {
+    // Gap of 0.004 → Math.round → 0; the 1pp floor keeps the predicate
+    // non-null so the marker still fires. The floor no longer has a phrasing
+    // job — the rendered marker carries no number at all.
     applyStore({
       viewMode: 'standard',
       phase: 'post',
@@ -507,8 +517,8 @@ describe('Render matrix — OptionNode × view × phase', () => {
       isResultsMode: true,
     } as any)
     renderOption({})
-    expect(screen.getByText('Close call: within 1 percentage point')).toBeDefined()
-    expect(screen.queryByText(/within 0 percentage point/)).toBeNull()
+    expect(screen.getByText('Close call with the leading option')).toBeDefined()
+    expect(screen.queryByText(/percentage point/i)).toBeNull()
   })
 
   it('Standard post close-call: "What would change this?" chip is added alongside "What would make this lead?"', () => {
@@ -541,7 +551,10 @@ describe('Render matrix — OptionNode × view × phase', () => {
       isResultsMode: true,
     } as any)
     renderOption({})
-    expect(screen.queryByText(/Close call:/)).toBeNull()
+    // Bound to the marker that actually renders — the old /Close call:/ pattern
+    // stops matching once the colon-and-number form is gone, which would make
+    // this absence assertion pass by testing nothing.
+    expect(screen.queryByText(/Close call/i)).toBeNull()
     expect(screen.queryByText('What would change this?')).toBeNull()
     // The standard "What would make this lead?" chip is still present.
     expect(screen.getByText('What would make this lead?')).toBeDefined()
--- a/src/canvas/nodes/__tests__/residualComparative.optionNode.spec.tsx
+++ b/src/canvas/nodes/__tests__/residualComparative.optionNode.spec.tsx
@@ -209,7 +209,7 @@ const CLOSE_CALL_PERMITTED = {
 
 const CLOSE_CALL_NODES = OPTION_NODES.slice(0, 2)
 
-describe('OptionNode — "Close call: within N percentage points" (found, not dispatched)', () => {
+describe('OptionNode — the close-call marker (was "Close call: within N percentage points"; the quantity retired 2026-08-10, the signal kept)', () => {
   beforeEach(() => {
     vi.mocked(useNodeDisplayMetadata).mockReturnValue(resultsMetadata(0.47))
   })
@@ -217,12 +217,20 @@ describe('OptionNode — "Close call: within N percentage points" (found, not di
   it('WITHHELD: no distance-to-the-leader line', () => {
     withStore(CLOSE_CALL_WITHHELD, CLOSE_CALL_NODES)
     renderNode(RUNNER_UP_ID, 'Standardise on Dell XPS')
-    expect(screen.queryByText(/Close call: within/)).toBeNull()
+    // Bound to the marker that actually renders: /Close call: within/ stops
+    // matching once the colon-and-number form is retired, so it would pass by
+    // testing nothing.
+    expect(screen.queryByText(/Close call/i)).toBeNull()
   })
 
   it('PERMITTED: the line renders (over-suppression control)', () => {
     withStore(CLOSE_CALL_PERMITTED, CLOSE_CALL_NODES)
     renderNode(RUNNER_UP_ID, 'Standardise on Dell XPS')
-    expect(screen.getByText('Close call: within 3 percentage points')).toBeDefined()
+    // ⭐ SUPERSEDED 2026-08-10: was 'Close call: within 3 percentage points'.
+    // The withheld-turn ENTITLEMENT this describe block exists to pin is
+    // unchanged; only the quantity has gone. The percentage-point gap between
+    // two win frequencies is retired from every user-facing surface.
+    expect(screen.getByText('Close call with the leading option')).toBeDefined()
+    expect(screen.queryByText(/percentage point/i)).toBeNull()
   })
 })
--- a/src/canvas/store/__tests__/analysisFreshness.spec.ts
+++ b/src/canvas/store/__tests__/analysisFreshness.spec.ts
@@ -50,11 +50,58 @@ describe('deriveAnalysisFreshnessUpdate', () => {
       computed_at: '2026-06-23T09:00:00.000Z',
     })
     expect(older).toBe(p) // older ignored
-    const equal = deriveAnalysisFreshnessUpdate(p, {
+  })
+
+  /**
+   * ⭐ CORRECTED 2026-08-10. This block used to assert that an EQUAL-timestamp
+   * payload is ignored, which positively enshrined a live defect: `computed_at`
+   * stamps the ANALYSIS, not the verdict, so CEE re-evaluates freshness against
+   * the CURRENT graph and re-sends the SAME analysis with a NEW verdict and a
+   * NEW `current_graph_hash`. Witnessed on staging: after a graph edit CEE
+   * returned `freshness: 'stale'` with a new `current_graph_hash` at an
+   * unchanged `computed_at`, and the store discarded it — keeping `fresh` and
+   * the stale hashes, so the panel told the user the analysis reflected a model
+   * it did not.
+   *
+   * The ordering rule is now STRICTLY older. Equal timestamps are decided by
+   * the echo guard below (`sameVerdict`), which is the check that actually
+   * answers "is this the same verdict?" — the timestamp only ever answered
+   * "is this the same analysis?", a different question.
+   */
+  it('equal computed_at + IDENTICAL verdict: still a no-op (same reference — the echo guard, not the clock, decides)', () => {
+    const p = prev({
+      freshness: 'fresh',
+      computedAt: '2026-06-23T12:00:00.000Z',
+      graphHashAtRun: 'h-run',
+      currentGraphHash: 'h-run',
+    })
+    const echoed = deriveAnalysisFreshnessUpdate(p, {
+      freshness: 'fresh',
+      computed_at: '2026-06-23T12:00:00.000Z',
+      graph_hash_at_run: 'h-run',
+      current_graph_hash: 'h-run',
+    })
+    expect(echoed).toBe(p)
+  })
+
+  it('equal computed_at + CHANGED verdict: CEE’s re-evaluation is APPLIED, hashes and all', () => {
+    const p = prev({
+      freshness: 'fresh',
+      computedAt: '2026-06-23T12:00:00.000Z',
+      graphHashAtRun: 'h-run',
+      currentGraphHash: 'h-run',
+    })
+    const reEvaluated = deriveAnalysisFreshnessUpdate(p, {
       freshness: 'stale',
       computed_at: '2026-06-23T12:00:00.000Z',
+      graph_hash_at_run: 'h-run',
+      current_graph_hash: 'h-edited',
     })
-    expect(equal).toBe(p) // equal ignored (no change)
+    expect(reEvaluated).not.toBe(p)
+    expect(reEvaluated?.freshness).toBe('stale')
+    expect(reEvaluated?.graphHashAtRun).toBe('h-run')
+    expect(reEvaluated?.currentGraphHash).toBe('h-edited')
+    expect(reEvaluated?.computedAt).toBe('2026-06-23T12:00:00.000Z')
   })
 
   it('applies a newer payload', () => {
--- a/src/canvas/store/analysisFreshness.ts
+++ b/src/canvas/store/analysisFreshness.ts
@@ -154,10 +154,25 @@ export function deriveAnalysisFreshnessUpdate(
 
   const computedAt = nonEmptyString(o.computed_at)
 
-  // Order by computed_at: ignore a strictly-older (or equal) payload when both
-  // timestamps are present. When the new payload has no computed_at we cannot
-  // order it, so we fall through to the content check below.
-  if (prev?.computedAt && computedAt && computedAt <= prev.computedAt) {
+  // Order by computed_at: ignore a STRICTLY-older payload when both timestamps
+  // are present. When the new payload has no computed_at we cannot order it, so
+  // we fall through to the content check below.
+  //
+  // ⭐ CORRECTED 2026-08-10. This read `computed_at <= prev.computedAt`, and the
+  // equal arm silently discarded CEE's authoritative verdict. `computed_at`
+  // stamps the ANALYSIS, not the verdict: CEE re-evaluates freshness against the
+  // CURRENT graph and re-sends the SAME analysis with a NEW verdict and a NEW
+  // `current_graph_hash`. Witnessed on staging — after a graph edit CEE returned
+  // `freshness: 'stale'` with a new `current_graph_hash` at an unchanged
+  // `computed_at`, and this branch threw it away, leaving the store on `fresh`
+  // with the pre-edit hashes. The panel then told the user the analysis
+  // reflected a model it did not.
+  //
+  // The timestamp only ever answered "is this the SAME ANALYSIS?". Whether it is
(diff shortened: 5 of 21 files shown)


Talchain (commented):
## Round-1 adversarial review of the COMPLETE PR — **CHANGES REQUIRED**

Reviewed whole at head `cef39dba9f58ac0874561151a67b573951f6619d` (re-asserted via `gh api`; `mergeable_state: clean`), in a fresh blobless clone at a unique `/private/tmp` path with `HEAD` asserted equal to the SHA. `VITE_SUPABASE_URL`/`VITE_SUPABASE_ANON_KEY` exported before every run; every spec count asserted by name.

Five of the six fixes are correct, well-pinned and verified by execution. **One blocking defect: fix 1's replacement subline introduces a false ordering claim in a state where the code it replaced was silent** — on the ungated, live V7 hero.

---

# 🔴 BLOCKING — F1: `runnerUpSubline` dropped the non-positive-gap guard that `leadSubline` had

`buildV7Headline.ts` — the retirement of `gapPoints` also retired, unintentionally, the guard that came with it.

The old code: `leadSubline(points)` returned `null` when `points <= 0`. Because `gapPoints = round((winner − runnerUp) * 100)`, that single predicate **silently covered the case where the designated winner is not the win-probability maximum**.

The new `runnerUpSubline` checks only for a tie *among rivals*. It never compares the top rival against the winner. Result, **proven by execution** (probe run at both tips, same input):

| | headline | subline |
|---|---|---|
| pre-PR (`944799c1`, = the currently deployed staging commit) | `Option A came out ahead in 40% of simulated scenarios` | `null` — **silent** |
| this PR (`cef39dba`) | `Option A came out ahead in 40% of simulated scenarios` | **`Next: Option B, 55%`** |

"Next:" is an ordering claim. It asserts Option B is the runner-up while stating a number 15 points **higher** than the leader named directly above it. That is precisely the class of adjacent-contradictory sentence this PR exists to remove, and it is a **regression against the deployed build**, not a pre-existing condition.

### Reachability — bounded at the producer, not from a fixture

This is not a fixture artefact. `determineWinnerSelection` (`useResultsSectionData.ts:87-96`) returns the producer's id **verbatim, with no comparison against any rival's win probability**; the argmax scan is only the fallback when the backend supplies nothing. `recommendedOption` is then `allOptions.find(o => o.isRecommended)`, keyed to that id.

And the estate says so itself, in the module that computes the very verdict the hero gates on — `src/lib/decisionVerdict.ts:305-309`:

> *"Keeping them apart matters: **PLoT may recommend an option that is not the win-probability argmax**, and a leader-minus-rival subtraction would then **go NEGATIVE and read as a tie**. The two options are still separated; the producer has simply pointed at the other one."*

Note what that paragraph also settles: because `separation` is measured on `top1 − top2` (the *actual* top two), `hasLeadingOption` is **true** in this state — so the `noLeadingOption` early-return does **not** gate it. The further apart the two options are, the more certainly the verdict permits the claim, and the more contradictory the rendered pair becomes.

Nothing downstream mitigates it: `V7Hero.tsx:128` renders `model.subline` behind a bare truthiness check, and `V7TopMatter` mounts **unconditionally** in `ResultsBody` (`:353`, "additive, passthrough, no flag") — this is a live surface on both flag postures.

### Why the suite did not see it

Three sibling sites in this estate guard exactly this condition — the old `leadSubline` (`points <= 0`), `hingeAwareDescription`'s `gapPct > 0 ? … : 'Statistically tied…'` else-arm, and `DecisionConfidencePanel.winProbabilityGap` (`gapPct > 0 ? gapPct : undefined`). The new subline is the only one of the four without it.

Every case in the PR's own corpus places the rival **below** the winner (`0.7/0.3`, `0.6/0.29`, `0.6/0.2·0.2`, `0.99/0.001`). The value class the contract admits — a rival at or above the winner — is absent from the corpus, so a fully green suite is consistent with the defect. That is CLAUDE.md trap 22 / 13d in its exact shape: *an invariant written with the same asymmetry as the code it tests*, and *a corpus that omits a class the contract admits cannot certify the code over that class*.

### Asked for

1. Restore the guard the retirement dropped — `if (first.winProbability >= winProbability) return null` — which reproduces `leadSubline`'s `points <= 0` semantics exactly and returns the surface to silence in the state it was silent in before.
2. A corpus case for the **opposite-direction twin** of each existing case (rival ≥ winner, including exact equality), shown RED before the guard and GREEN after — sourced from the producer's output domain, not from the author's head.
3. A mutant proving the new guard bites.

---

# 🟠 SHOULD FIX (non-blocking)

### F2 — licence 4, ground 1 is **false on the goal-preferring branch**

The comment in `certaintyCopy.ts:177-179` (repeated verbatim in the spec docblock) states the deletion is licensed because *"the panel ALREADY shows the winner's own probability where this lede renders — the confidence ring's `overrideScore`, captioned `COMPARATIVE_COPY.label`."*

Derived at the bytes (`DecisionConfidencePanel.tsx:178-186`), `ringClaim` **prefers the goal probability**:

```
isFiniteProbability(goalProbability)
  ? { value: goalProbability, caption: GOAL_ANCHOR_COPY.label(...) }   // ← wins whenever a target was set
  : isFiniteProbability(winProbability)
    ? { value: winProbability, caption: COMPARATIVE_COPY.label }
    : null
```

So on any run carrying a success threshold — the *common* case, and the one the ring's own header calls the preferred branch — the ring shows the **goal** number under the **goal** caption, and the winner's own win probability is not on the panel at all. Ground 1 is true only on the fallback branch.

**This does not invalidate the deletion.** Ground 2 does stand: I verified the F4 adjudication is real at the bytes (`certaintyCopy.ts:200-214`) and says exactly what is claimed — *"Paul's ruling DEMOTES the comparative number below the goal number, so a magnitude-free comparative sentence on this surface is the CORRECT behaviour, not a gap."* The two grounds were offered as independent, and the independent one holds. But an overstated justification written into a comment is what the next session inherits as settled fact (trap 20). Please correct both copies to say the ring shows *the winner's own headline quantity, goal-preferred*, or drop ground 1 and rest on the adjudication.

### F3 — the panel-wide guard's `GAP_CLAIM` also bans a **sanctioned** string

`GAP_CLAIM = /leads?\s+by\s+\d+\s+points?/i`, asserted absent across everything `ResultsBody` renders. The **goal** arm still legitimately emits `Leads by N points` via the surviving `leadSubline` (`buildV7Headline.ts:251`) — deliberately, and the PR documents that scoping decision. Verified: `"Leads by 40 points"` matches `GAP_CLAIM`.

The guard passes only because its fixture carries no `goalProbability`/`goalThreshold`. A future lane adding goal data to this fixture gets a **false RED on correct behaviour**. The in-file "SCOPE, STATED EXACTLY" block is true as far as it goes (I verified both its claims — `ResultsBody` genuinely does not compose `OptionNode`, and the `certaintyCopy` suffix is genuinely covered on the separate flag-OFF case), but it does not disclose this sanctioned exception. Please either narrow the regex to the comparative arm or add the goal-arm exception to the scope block with a case pinning it.

### F4 — the file's own verbatim-sourcing invariant is now false

`buildV7Headline.ts`'s header asserts *"Every string form below is sourced VERBATIM from a live production headline surface, so 'Headline copy matches production strings' (spec row 3 done-when) holds"*, followed by an enumerated list. `Next: {label}, {pct}` is a **novel** string and was not added to the list. The header was updated for the goal-arm note but not for the invariant it breaks.

---

# ✅ VERIFIED GOOD (executed evidence)

**Flag posture — both halves verified independently, and the lane is right.**
- Single reference site: I grepped *every* `DecisionConfidencePanel` reference (49 files). Exactly one render site — the element is defined at `ResultsBody.tsx:294` and mounted at `:508`, inside `{!isAnalysisHeroPanelEnabled() && …}`. All other non-test hits are comments. `buildCertaintyCopy` has exactly **one** production caller.
- Deployed posture derived from the **bundle**, not from YAML (trap 18): crawled 87 JS chunks transitively from `staging--olumi.netlify.app` (5.88 MB, 4/5 positive controls present — the fifth string legitimately no longer ships). `chunks/flags-DBpusaUf.js` bakes **`VITE_FEATURE_ANALYSIS_HERO_PANEL:"1"`**. The panel is off by default on staging. ✅ Framing correct.
- ⚠ One nuance for the record: `makeFlag` gives **localStorage precedence over env** and `'0'`/`'false'` forces OFF, so the panel *is* reachable on staging via a documented local override. "Dark under the default deployed posture" is exact; "not served" is slightly stronger than the mechanism supports. Safe direction either way.
- The surfaces that must be live for closure **are** live: `V7TopMatter` mounts unconditionally (`:353`); `OptionCards` sits in Section 2 gated only by `!isSingleOption`; `OptionNode` is a canvas node. All three retired strings confirmed present in the deployed bundle (`ReactFlowGraph-5gNC_h2B.js`) — `Behind by …`, `Close call: within …`, `Leads by …` — i.e. these were live-surface fixes, not dark ones.

**Licence 2 (OptionCards) — holds at the seam.** `hingeAwareDescription` has no consumer outside `OptionCards`; its output and the `win-pct-{id}` readout are rendered by the same `OptionCard` for the same option. `win-pct-{id}` is gated on exactly `option.winProbability != null` (`:727`), inside the card header with no compact/variant branch — the same strict precondition as both retired arms. I looked for the counter-case the mandate asked for and there is none. `''` renders nothing (`description ? … : null`, `:750`) — no fallback string sneaks in.

**Panel-wide absence guard — not vacuous, and both preconditions bite.** Proven by mutation:
- Removing the flag-force + both preconditions ⇒ the DEFENCE-IN-DEPTH case **PASSES on the deployed posture where the panel never mounts**. The preconditions are load-bearing exactly as claimed.
- Mistyping `PP_CLAIM` ⇒ the **positive control REDs**. It reaches the same `container.textContent` read and the same regex as the absence assertions.

**Freshness pair — the whole claim, verified.** Mutating `<` back to `<=` REDs **exactly one** named test — *"equal computed_at + CHANGED verdict: CEE's re-evaluation is APPLIED, hashes and all"* — while equal+identical (no-op) stays GREEN (26 passed). The old spec did positively enshrine the bug (it asserted a `freshness: 'stale'` payload was "ignored (no change)").

**Unmount-flush limbs — pinned separately, by different named cases.**
- Restoring the gate bypass ⇒ REDs `(c) unmount with a PENDING (never-fired) debounce must not write the unsettled graph`.
- Removing the fire-time null ⇒ REDs `(e) a debounce that already FIRED AND WROTE is not re-flushed at unmount`.

Different cases, different assertions, each leaving the other four green. Every case pins its own precondition in-test via `expect(shouldPersistGraphForScenario(SCENARIO)).toBe(…)`, and `(a)`/`(d)` are genuine over-suppression controls — `(d)` proves the gate-open path still flushes exactly once. No path persists a kept-unsettled graph.

**Test-hygiene disclosures — all three audited and accurate**, and the disclosure itself is the right standard:
- `ownedLeaderClaim.optionCards` over-suppression control did positively assert the retired sentence; rebound to `getByTestId('win-pct-…')` — bound by identity, not by a value another element could satisfy. ✅
- `LEADER_LANGUAGE`'s `Behind by N percentage points` row **is** now unconditionally true. Kept as a comeback guard with the loss written down rather than dressed as coverage. ✅ Honest.
- `/Close call:/` and `/Close call: within/` negatives **would** have passed by testing nothing; rebound to `/Close call/i`. ✅
- Swept the other changed specs for **undisclosed** vacuity. The `certaintyCopy` rewrite replaces per-case suffix pins with a derived sweep (12 gaps × 4 tiers × 3 readiness = 144, with `expect(asserted).toBe(…)` guarding a zero-iteration loop) — real. `caveatGuarantee`/`HeroFooterAlignment` rebinds use anchored `/Option A currently leads$/`, which stays discriminating. Minor, undisclosed: four cases in the retired-suffix block still pass `winProbabilityGap` into a function that no longer reads it, so those inputs are now inert — same class as the disclosed `LEADER_LANGUAGE` loss, and covered by the new sweep. Worth a one-line note.

**Residual sweep.** No user-facing win-frequency-gap string remains on the results panel. `gapPp` confirmed zero production readers (written in `decisionVerdict.ts`, never read to render). Remaining `* 100` gap arithmetic is predicate-only at all four sites. Two unrelated "percentage points" strings exist off-panel (`ContestedEdgeCard` EVOI impact; `sandbox-guide` `CompareState` before/after delta) — different quantities, outside the ratified rule as scoped.

**Mutation hygiene.** Throwaway worktree at an absolute path **outside** the repo root. Isolation proven by **writing**, not locating (trap 9g): distinct inodes (`585258736` vs `585352069`), sentinel landed only in the worktree, source SHA-256 unchanged, restore left the tree clean. All restores `git checkout HEAD -- <path>` (trap 9h). Applied-check scoped to `src/` with the pristine control reading **exactly 0** (the untracked `node_modules` symlink never entered the count — trap 12e). Mutation **aim asserted** (the changed line printed) before believing any result. Cross-limb independence confirmed. One caveat disclosed: an `M3` copy-surface mutant was killed by a 10-minute timeout before its restore line ran; I restored it immediately and re-asserted `src/` clean at HEAD, so no later measurement was taken against a polluted tree — but that mutant produced **no usable result** and I am not claiming one.

**Gates at the head.**
- Typecheck gate **PASSED** — coverage 3347/3387 tracked files; ratchet **2487 errors / 618 files, baseline 2487**. Equal, not merely green.
- rules-of-hooks ratchet **PASSED** — 232 known violations across 15 files, exactly matching baseline.
- CI check-runs queried on the **exact SHA**, filtering `.status=="completed"` first (trap 24b): **13 total, 13 completed, running 0, failed 0**.
- New/changed specs collect non-zero by name: `ResultsBody.winFrequencyGapAbsence` **7**, `analysisFreshness` **27**, `useScenario.unmountFlushPersistenceGate` **5**, `certaintyCopy` **51**, `buildV7Headline` **17**, `buildV7Headline.goalLeader` **18**, `render-matrix` **58**, `OptionCards` **50**, `HeroFooterAlignment.matrix` **17**.
- Wide suite: **1596 files, 23,179 tests — 23,111 passed, 66 skipped, 2 failed**, both in `OutputsDock.conversationSingleton.spec.tsx`. **Stated precisely, because this is a claim type that is easy to overstate:** that file is **not in the PR diff**, it passes **12/12 in isolation at this head**, and CI is green on the exact SHA. So it is not attributable to this PR on the evidence I have — a singleton-invariant spec is exactly the shape that goes order- or resource-dependent under a fully parallel run, and my machine was under contention. I did **not** run the base's full suite, so I am **not** claiming it proven pre-existing; I am claiming it not attributable, and naming the gap.

---

## Summary

The persistence and freshness halves (fixes 5 and 6) are the strongest work here — genuinely discriminating pairs, preconditions pinned in-test, controls that would catch over-suppression. The three copy retirements on the live surfaces (2, 3, and the deletion half of 1) are correctly licensed and correctly scoped, and the flag-posture self-correction is right at the deployed bytes.

The blocker is narrow and fixable: **retiring `gapPoints` also retired the guard that rode on it**, and the replacement asserts an ordering the numbers can contradict — in a state this codebase explicitly documents as reachable, on a surface no flag gates. Restore the guard, add the opposite-direction twins, and this is ready.

Talchain (commented):
## Round-4 final confirmation — **APPROVE**

Head `67099b1efa936f56a41d3f7d2314f43258423b96`, verified identical from **both** `gh api` and `git ls-remote refs/pull/663/head`; `mergeable_state: clean`; still **20 files, no new files** beyond the two specs added in round 1. Round-4 delta is 5 files / +220 −31.

**Items 1 and 2 hold.** Approving.

---

### 1. F1 — the restored guard, and the monotone claim

The four-row table reproduces **exactly** as stated, executed at this head:

| input | subline |
|---|---|
| 0.40 / 0.55 inverted | `null` |
| 0.50 / 0.50 tie | `null` |
| 0.404 / 0.400 sub-point | `null` |
| 0.55 / 0.40 twin | `Next: Option B, 40%` |

**The rounded-lead choice was right and my bare `>=` would have been wrong.** I ran the same four inputs against `944799c1` itself: it is `null`, `null`, `null`, `"Leads by 15 points"`. The sub-point row is the one that matters — `round((0.404 − 0.400) × 100) = 0`, so the deployed build was silent there, and a bare `>=` comparison would have started emitting a sentence where the shipped product says nothing. The lane was right to decline the instruction and say why.

**The monotone claim, tested adversarially rather than accepted.** I looked for any input where this head emits a subline `944799c1` suppressed, and there is exactly one class: the rival filter tightened from `typeof === 'number'` to `isFiniteProbability`, so a **non-finite** rival that was the old argmax is now excluded. Executed, winner 0.40 with rivals {B: +Infinity, C: 0.10}:

- `944799c1` → `null` (Infinity sorted first; `round((0.40 − Inf) × 100) = -Infinity ≤ 0`)
- `67099b1e` → `Next: Option C, 10%`

So the claim is not *universally* true — but it is true over the entire **producer-reachable** domain, and the counterexample runs the safe way on both counts:

1. **Not reachable.** `winProbability: prob.win_probability` is a direct passthrough of parsed report JSON (`useResultsSectionData.ts:1617`), and JSON cannot transport `Infinity` or `NaN`. The value is constructible only in a hand-written fixture.
2. **The new behaviour is strictly more honest there.** `Next: Option C, 10%` beneath a 40% headline is true — C really is the top finite rival. The old silence was an accident of a garbage value poisoning a subtraction. And the `NaN` variant is sharper still: `944799c1` returns **`"Leads by NaN points"`** on the same shape, i.e. the deployed build can render a literal `NaN` to a user. This head returns the honest sentence. The filter tightening quietly fixed that.

**Verified: monotone over every input the producer can emit; the sole exception is unreachable and improves on the deployed build.** ✅

### 2. E1/E2 — the discriminating pair, both run with line-level aim asserted

**E1** (guard removed — aim: both `const leadPoints = …` and `if (leadPoints <= 0) return null` deleted) REDs **exactly three**, by name:
- `F1: a rival ABOVE the designated winner gets NO subline` → `expected 'Next: Option B, 55%' to be null`
- `F1: a rival EQUAL to the winner gets no subline either` → `'Next: Option B, 50%'`
- `F1: a lead too small to survive rounding gets no subline` → `'Next: Option B, 40%'`

Twin **GREEN**. `buildV7Headline.goalLeader.spec.ts` and `ResultsBody.winFrequencyGapAbsence.spec.tsx` both **PASS** — no cross-surface trip. ✅

**E2** (`<= 0` → `< 0`, single-line aim) REDs **only equality and sub-point**; the inversion case stays GREEN (`round(-15) = -15 < 0`, still suppressed). ✅

**The lane's reading holds, and it is the right reading.** E1 alone is satisfied by any guard whatever — including a `< 0` one that would leak both boundary cases. E2 is what pins the boundary at `<= 0`. The two mutants fail on *different assertions*, which is what makes it a pair rather than two runs of the same probe.

### 3. F2 — accurate at the bytes, corrected in both places

Re-derived at this head (`DecisionConfidencePanel.tsx:178-186`): `ringClaim` takes `goalProbability` with `GOAL_ANCHOR_COPY.label` when finite, falling back to `winProbability` with `COMPARATIVE_COPY.label` only when no goal figure exists. **Ground 1 was indeed false** — on any target-bearing run the winner's own win probability is absent from that panel.

Corrected in **both** locations (source comment and spec docblock), and both say plainly *"the FIRST WAS FALSE"* with the reason, rather than quietly dropping it — *"a false rationale left in a comment is how the next lane inherits a wrong premise"*. Correct, and the right instinct. Ground 2 re-verified real at the bytes (`certaintyCopy.ts` F4 note, *"would have added a live claim the ruling does not want"*), and the deletion stands on it alone. ✅

### 4. F3 — the self-evidencing assertion is real, and coverage did not relax

The goal-data case asserts the sanctioned sentence **present** (`Leads by 50 points`, from 0.9 − 0.4) and then asserts `expect(subline.textContent).toMatch(GAP_CLAIM)` — executable proof that the previous panel-wide assertion **would have RED on a correct render**. Exactly the right way to record it. `PP_CLAIM` stays panel-wide and is still asserted against the whole container including the hero. ✅

I re-ran my own banned-form corpus against the scoped guard, injecting each form as the comparative subline. All three RED:

| injected form | result |
|---|---|
| `Leads by 40 points` | 4 failed |
| `Ahead by 40 points` | 2 failed |
| `Behind by 40 percentage points` | 3 failed |

Note the middle row — I expected that one to be the gap, since the hero subline is no longer checked against `POINTS_CLAIM`. It is caught anyway, because the subline is *additionally pinned by identity* (`toHaveTextContent('Next: {label}, 31%')`) in two cases, which is a stronger binding than a shape regex. **Coverage of the banned class is intact.** ✅

### 5. F4 — provenance list fixed; the replacement invariant is **still inaccurate** (non-blocking)

The provenance list is now correctly labelled as *"the forms that were inherited, not an invariant over all of them"*, and the `Next:` string is disclosed as new copy with no production antecedent. That half is right.

But the replacement invariant — *"every NUMBER rendered is read from the results store and never derived into a new quantity"* — **is false of this module.** The goal arm renders `goalLeadPoints`, which is `Math.round((leaderValue − runnerUp) * 100)` (`selectGoalLeader.ts:223`) — a derived difference — as `Leads by N points` at `buildV7Headline.ts:185`. **This PR's own new F3 test proves it by execution:** goal probabilities 0.9 and 0.4 render `Leads by 50 points`, and `50` exists nowhere in the store.

This is the F4 defect class recurring one turn later — an over-broad invariant swapped for another over-broad invariant. It is a comment-accuracy issue with no product consequence, so it does **not** block. A one-line follow-up: scope it to the comparative arm and name the goal arm's derived quantity as the deliberate exception (which the file already documents everywhere else).

### 6. Gates on the exact head

- CI **13/13**, filtering `.status=="completed"` first: **completed 13, running 0, failed 0**. Includes **`Staging Gate` ✅**, `TypeScript + Lint` ✅, `Full Test Suite` shards 1–4 all ✅.
- Typecheck gate **PASSED** — coverage 3347/3387; ratchet **2487 / 618 files, baseline 2487**. Equal.
- rules-of-hooks ratchet **PASSED** — 232 across 15 files, exactly baseline.
- Specs collect non-zero **by name**, 8 files / **145 tests**, all green: `buildV7Headline` **21** (17→21, the four F1 cases), `ResultsBody.winFrequencyGapAbsence` **8** (7→8, the F3 goal case), `certaintyCopy` **51**, `buildV7Headline.goalLeader` **18**, `analysisFreshness` **27**, `useScenario.unmountFlushPersistenceGate` **5**, `ownedLeaderClaim.optionCards` **9**, `ownedLeaderClaim.surfaces` **6**.
- `OutputsDock.conversationSingleton.spec.tsx`: unchanged at this head, still not in the diff. CI's four Full Test Suite shards are green here, which corroborates the round-1 call — I reported it as not-attributable rather than proven pre-existing, and nothing has changed that.

Mutation hygiene throughout: worktree at an absolute path outside the repo root, isolation proven by **sentinel write** (distinct inodes `585748820` / `585753581`, source SHA-256 unchanged), HEAD-relative restores, applied-check scoped to `src/` with the pristine control at exactly 0, aim printed per mutant, clean tree re-asserted after every one.

---

## Verdict: **APPROVE**

F1 is properly closed, and closed better than I specified — the lane reproduced the defect at its own head, confirmed the producer bound independently, and then declined the bare `>=` I ordered because it would have broken monotonicity at the sub-point boundary. That correction is the substantive contribution of this round, and it was right.

The E1/E2 pair is a genuine discriminating pair, not two probes of the same thing. F2's correction states what was wrong instead of hiding it. F3 turned a latent false-RED into executable evidence about its own scoping.

One residual, non-blocking: **F4's replacement invariant is still broader than the module satisfies** (the goal arm renders a derived number). Worth a one-line follow-up, not a round 5.

PR #664, opened 2026-08-10
P0: the provenance receipt rendered a DIFFERENT decision's brief

Talchain:
**Do not merge — adversarial review follows.**

## The defect

Witnessed on deployed staging (UI `944799c1`, CEE `1cbe757`, 10 Aug 21:05–21:33 UTC). The surface titled **"What you gave me, and what I did with it"** — whose entire job is to tell the user what Olumi did with *their* brief — rendered **a different decision's brief**, verbatim, under "What you gave me", with a subtitle counting *that* brief's figures ("27 of 28"). It persisted through reset canvas → new brief → draft → analysis → an edit turn. **Only a hard page reload corrected it.** Two unrelated decisions both reporting "27 of 28" is what exposed it.

It has a data-leak shape: on a shared session this renders one person's strategic brief inside another's decision. **Bound:** the store is plain in-memory Zustand (no `persist`), so it cannot cross page loads or profiles — the exposure is one page session, e.g. an unreloaded handover on a shared screen. That bound is a property of the current implementation, not a guarantee; persisting this store later would widen it to a genuine cross-user leak.

## Mechanism (at the bytes)

`serverGraphHydration` writes `contextIntegrityStore` **only** on `result.status === 'graph'`. Every other cold-read outcome — `absent`, `notReadable`, `unavailable`, `refused`, `unusable`, and the non-UUID `skipped` — returns **before** that write. Nothing clears the store either: `reset()` had **zero production callers**, and `DECISION_CONTEXT_CLEAR` — the canonical per-decision clear set, which does carry the sibling 2.312 `serverGraphIdentity` token for exactly this reason — is a partial of `useCanvasStore` and cannot reach a separate store. `useServerGraphHydration` attempts **once per scenario id**, so it never self-corrects. The previous decision's content simply stayed, and a reload fixed it because the store re-initialises empty.

The store had carried a `scenarioId`; it was removed as write-only. That reasoning (trap 10) was right in general and wrong here — the answer was to give the field a **reader**, not delete it. The removal note argued staleness was "answered UPSTREAM", but the upstream guard answers *"is this in-flight response for the scenario the user is still on?"* and runs only on the `'graph'` path, while the render needs *"does the content I am about to show belong to the scenario on screen?"* Two questions under one name (trap 21), differing exactly where the defect lives.

## Reachable set

| Transition | Verdict | How established |
|---|---|---|
| Reset canvas → new decision | **LEAKY** | witnessed live + traced |
| Any 2nd+ decision in one page session, cold read not `graph` | **LEAKY** | traced (same mechanism; reset not required) |
| Scenario switch A→B, B's read `absent`/404/503/401/429/transport-fail | **LEAKY** | traced |
| Local draft (non-UUID id → `skipped`) | **LEAKY** | traced |
| In-flight window during a slow A→B switch | **LEAKY** | traced (now also closed by this fix) |
| Scenario switch A→B where B's read returns `graph` | safe | traced — the write overwrites |
| Virgin profile, **first** decision of a page session | safe | traced — store starts empty |
| Virgin profile, **second** decision, no reload | **LEAKY** | traced — *not* reset-specific |
| Hard reload | safe | witnessed live |

**Not tested by execution:** every row marked *traced* is a state-lifecycle derivation, not a driven browser witness. I did not drive staging; the live evidence is the original witness's. The reset-canvas row is the only one with a live witness.

## The fix

Bind content to its decision and assert it **at the point of render**, so a store nobody cleared cannot defeat it:

- `contextIntegrityStore` carries `scenarioId`, **required** on every write — a new writer cannot omit it by accident.
- `serverGraphHydration` records the scenario the payload came back for.
- `V7WhatIWasGivenSection` renders only on a **positive** identity match, otherwise nothing — the truthful reading of "we have nothing for this decision". It must be positive, not `!==`: `!==` passes when either side is null, which is exactly what `resetCanvas` leaves behind (`currentScenarioId: null`).

Keying the content is what makes clearing unnecessary: an unkeyed clear must be remembered at every transition (trap 12), whereas content carrying its own identity fails safe at the point of use. Twelve effective lines of behaviour change.

## Evidence

**RED-first at pristine `944799c1`**, by name — 51 collected, 3 failed:
- `THE P0: content recorded for another scenario is NOT rendered on this one`
- `content recorded with NO scenario id is never rendered`
- `a POSITIVE match is required — an unknown current decision renders nothing`

**Mutants** — throwaway worktree outside the repo root, isolation proven by *writing* a sentinel and asserting the source was unchanged (inodes compared, trap 9g), HEAD-relative restores, applied-check scoped to `src/`, control at exactly 0, **aim asserted per mutation**:

| # | Mutation | Applied | Result |
|---|---|---|---|
| M0 | control, unmutated | **0** | GREEN 52/52 |
| M1 | identity gate removed | 1 | **RED** — 3 identity tests |
| M2 | gate unconditionally `true` | 1 | **RED** — 3 identity tests |
| M3 | `typeof` conjunct dropped | 1 | **RED** — the null===null test |
| M4 | writer drops `scenarioId` | 1 | **RED** — 36 tests |
| M5 | *unrelated* empty-state guard broken | 1 | **RED on its own test only; identity tests GREEN** |

M5 is the discriminating partner: it proves the identity tests bind to the identity gate specifically, not to any change in the component.

**Two findings the kit forced, both fixed here:**
1. **M3 initially SURVIVED.** No case put *both* ids at null with content on file, so a bare `recordedScenarioId === currentScenarioId` (`null === null` → true) passed the whole suite. Now pinned by `BOTH ids absent is NOT a match`.
2. **The new gate SHADOWED an existing test.** `renders nothing only when there is genuinely nothing on file` began passing via the identity gate rather than the guard it names (trap 13b) — it left the store empty, so `scenarioId` was null and the gate suppressed first. Repaired to record *this* decision with no content, so the empty-state mutant reddens it again.

**Gates, in the required check's step order** (`lint → typecheck → guards`):
- lint: **0 errors**; hooks ratchet **exactly at baseline** (232 violations / 15 files)
- typecheck ratchet: **2487 = baseline 2487**; coverage 3345/3385
- `ci:guard:zustand`, `ci:guard:starters`, `check:vendor`, `ci:guard:schemas-version`: all pass
- full suite, sharded 4×: **23,103 passed · 66 skipped · 0 failed**

UI suites run with `VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` exported; the target spec's own collected count asserted by name (51 pristine → 52 fixed), never inferred from the suite total.

## Scope

This defect only. **The counts-on-reload finding does NOT share this store** — "N worth checking" comes from `StrengthenPanel` via `strengthenStore`, which never reads `contextIntegrityStore`. It is left untouched. It *is* the same defect class, though, and worth its own row: `strengthenStore` has no `scenarioId` and its `_reset` has **zero production callers** (complete manifest: all six non-declaration hits are in `__tests__`). I did not establish causation for the witnessed 5→1 by execution.

No file in this change overlaps the CEE brief-audit/state-query lane — the cause is entirely UI-side and CEE was not touched.

### Known limitation, stated plainly

Where the gate suppresses, the receipt is now **absent** for that session rather than wrong, and hydration will not retry (once per scenario id). A reload still produces the truthful receipt. Truthful absence beats confident wrongness, but closing that gap — re-hydrating when the server graph appears — is a separate change and is not attempted here.

### DIFF
--- a/src/canvas/hydrate/serverGraphHydration.ts
+++ b/src/canvas/hydrate/serverGraphHydration.ts
@@ -148,7 +148,12 @@ export async function hydrateCanvasFromServer(
   // stored AS a null: the surface renders it as "we cannot tell you", never as
   // an empty list, which on a brief we demonstrably lose content from would be
   // a new and more damaging lie than the silence it replaces.
+  // `scenarioId` is the scenario this content DESCRIBES, and the surface refuses
+  // to render unless it matches the live one. It is the requested id, not
+  // `currentId`: the two are equal here (the guard above returned otherwise),
+  // and the requested id is the one the payload actually came back for.
   useContextIntegrityStore.getState().setContextIntegrity({
+    scenarioId,
     briefText: result.briefText,
     manifest: result.notModelled,
   })
--- a/src/canvas/stores/contextIntegrityStore.ts
+++ b/src/canvas/stores/contextIntegrityStore.ts
@@ -25,30 +25,62 @@ import { create } from 'zustand'
 import type { NotModelledManifest } from '../../adapters/cee/notModelled'
 
 /**
- * ── WHY THERE IS NO `scenarioId` HERE ──────────────────────────────────────
- * There was one, and its comment claimed it "guards against a stale read from a
- * previous scenario being shown against the current one". It guarded nothing:
- * it was WRITE-ONLY. `serverGraphHydration` set it and no consumer ever read it
- * — `V7WhatIWasGivenSection` selects `briefText` and `manifest` and nothing
- * else. A field with zero readers cannot fork identity and cannot gate a
- * render, so its blast radius was zero by construction (CLAUDE.md trap 10),
- * and a comment asserting a guarantee nobody enforces is worse than no field:
- * the next reader trusts it and stops looking.
+ * ── WHY `scenarioId` IS BACK, AND WHY IT MUST HAVE A READER ────────────────
+ * This field existed, was removed as write-only, and its removal shipped a
+ * P0: the receipt rendered A PREVIOUS DECISION'S BRIEF under "What you gave
+ * me", and survived reset-canvas → new brief → draft → analysis → edit. Only a
+ * page reload cleared it (this store is in-memory, so a reload re-inits it).
  *
- * The staleness question itself is real but is answered UPSTREAM, before this
- * store is written: `serverGraphHydration` compares the requested scenario
- * against `useCanvasStore.currentScenarioId` and returns `'skipped'` on a
- * mismatch, so a hydration for a scenario the user has left never reaches
- * `setContextIntegrity` at all. If that ever stops being true, the fix is a
- * guard with a reader and a test — not a field that records the answer and
- * throws it away.
+ * The removal note argued the staleness question was "answered UPSTREAM" by
+ * `serverGraphHydration` comparing the requested scenario against
+ * `useCanvasStore.currentScenarioId`. THAT WAS FALSE, and the reason is
+ * CLAUDE.md trap 21 — two questions under one name:
+ *
+ *   · The upstream guard answers *"is this in-flight RESPONSE for the scenario
+ *     the user is still on?"* It only runs on the `'graph'` path.
+ *   · The render needs *"does the content I am about to SHOW belong to the
+ *     scenario on screen?"*
+ *
+ * They differ precisely where the defect lives. `setContextIntegrity` is
+ * reached ONLY on `result.status === 'graph'`; every other cold-read outcome
+ * (`absent`, `notReadable`, `unavailable`, `refused`, `unusable`, and the
+ * non-UUID `skipped`) returns BEFORE the write. A freshly-minted scenario
+ * reliably answers `absent` — CEE has no graph for it yet at the moment
+ * `useServerGraphHydration` fires — so nothing is written, nothing is cleared,
+ * and the previous decision's brief simply stays. The hook attempts ONCE PER
+ * SCENARIO ID, so it never self-corrects.
+ *
+ * The removal reasoning about write-only fields (trap 10) was right in general
+ * and wrong here: the answer was not to delete the field but to GIVE IT A
+ * READER. `V7WhatIWasGivenSection` now refuses to render unless this id
+ * positively matches `useCanvasStore.currentScenarioId`, so a stale store
+ * cannot reach the screen even when nothing clears it. Keying the content is
+ * what makes clearing unnecessary: an unkeyed clear must be remembered at every
+ * transition (the hand-maintained mirror, trap 12), whereas content that
+ * carries its own identity fails safe at the point of use.
  */
 export interface ContextIntegrityState {
+  /**
+   * The scenario this content describes. `null` = nothing recorded.
+   *
+   * ⚠ HAS A READER, AND MUST KEEP ONE. `V7WhatIWasGivenSection` gates its
+   * entire render on this matching the live scenario. If a future change drops
+   * that comparison, the P0 above returns silently — the pinning spec is
+   * `V7WhatIWasGivenSection.spec.tsx` → "never renders another decision's
+   * brief".
+   */
+  scenarioId: string | null
   /** The brief as the user wrote it, byte-verbatim. `null` = none persisted. */
   briefText: string | null
   /** CEE's manifest. `null` = we were told nothing. NEVER "nothing dropped". */
   manifest: NotModelledManifest | null
+  /**
+   * `scenarioId` is REQUIRED, deliberately: content this store cannot attribute
+   * to a decision is content the surface must never show, and making the caller
+   * state it means a new writer cannot omit it by accident.
+   */
   setContextIntegrity: (input: {
+    scenarioId: string | null
     briefText: string | null
     manifest: NotModelledManifest | null
   }) => void
@@ -56,12 +88,14 @@ export interface ContextIntegrityState {
 }
 
 const EMPTY = {
+  scenarioId: null,
   briefText: null,
   manifest: null,
 } as const
 
 export const useContextIntegrityStore = create<ContextIntegrityState>((set) => ({
   ...EMPTY,
-  setContextIntegrity: ({ briefText, manifest }) => set({ briefText, manifest }),
+  setContextIntegrity: ({ scenarioId, briefText, manifest }) =>
+    set({ scenarioId, briefText, manifest }),
   reset: () => set({ ...EMPTY }),
 }))
--- a/src/components/results/v7/V7WhatIWasGivenSection.tsx
+++ b/src/components/results/v7/V7WhatIWasGivenSection.tsx
@@ -32,6 +32,15 @@
  * those read as "everything made it in", which on a brief we demonstrably lose
  * content from is a worse lie than the silence it replaces.
  *
+ * ── THE IDENTITY GATE ──────────────────────────────────────────────────────
+ * This surface renders ONLY when the context-integrity store's `scenarioId`
+ * positively matches the live `currentScenarioId`. It shipped without that
+ * check and rendered A PREVIOUS DECISION'S BRIEF, verbatim, under "What you
+ * gave me" — the trust anchor telling the user something false about their own
+ * input, with a cross-decision data-leak shape on any shared session. See the
+ * store's header for the mechanism. The gate lives at the point of RENDER, not
+ * only at the point of write, so that a store nobody cleared cannot defeat it.
+ *
  * ── MOUNT ──────────────────────────────────────────────────────────────────
  * Hosted in the UNCONDITIONAL `v7-top-group`. `ResultsBody` forks on
  * `analysisHeroPanel` and this estate has twice shipped a feature dark by
@@ -42,6 +51,7 @@ import { useState } from 'react'
 import { ChevronDown } from 'lucide-react'
 
 import { typography } from '../../../styles/typography'
+import { useCanvasStore } from '../../../canvas/store'
 import { useContextIntegrityStore } from '../../../canvas/stores/contextIntegrityStore'
 import type { NotModelledItem } from '../../../adapters/cee/notModelled'
 import { ClampToggle } from './ClampToggle'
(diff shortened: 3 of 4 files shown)


Talchain (commented):
## Round-1 adversarial review — **APPROVE**

Independent verification at `c455f065aa70c10a3daa7043a5d78570619eb96e` (asserted from **both** `gh api` and `git ls-remote`), fresh blobless clone at a unique `/private/tmp` path, `VITE_SUPABASE_*` exported for every run. The diagnosis holds at the bytes, the gate is correct, and the two self-disclosed instrument defects are genuinely repaired — I proved the second one in **both** directions.

One correction to the record is required before merge (PR body only, no code change), plus two rows to file. Neither blocks.

---

### 1. The diagnosis — verified, including the two load-bearing claims

**The write path.** `serverGraphHydration.ts` reaches `setContextIntegrity` at :155 only past the `status !== 'graph'` switch (:109–126) and the non-UUID return (:97–99). `absent`/`notReadable`/`unavailable`/`refused`/`unusable`/non-UUID-`skipped` all return first. Confirmed.

**Nothing clears it.** Complete manifest, `rg -uu` over the whole tree less `node_modules`/`.git` (binary-safe, trap 17): **19 references total**. `reset()` has exactly **two** callers, both `__tests__` (spec :124, :132) — **zero production callers**. One production writer, one production reader. `DECISION_CONTEXT_CLEAR` (`store.ts:1233`) is a plain `useCanvasStore` state partial and **does** carry `serverGraphIdentity: null` — so the very set that clears the sibling token structurally cannot reach a separate store. Corroborating: `importCanvas` reaches *other* stores by explicit call (`useDraftStore`, `useAnalysisSnapshotStore`, `useComparisonStore`); `useContextIntegrityStore` was never added to any of them.

**Once per scenario id.** `useServerGraphHydration.ts:38,42–43` — `attemptedRef` short-circuits. Never self-corrects. Confirmed.

**⭐ The deletion history — confirmed, and the reasoning inversion is real.** `e8a2fc57` (#638), point 3, in its own words: *"`contextIntegrityStore.scenarioId` deleted. It was WRITE-ONLY … A zero-reader field cannot guard anything (trap 10) … The guard is NOT reimplemented: the staleness question is already answered upstream."* Exactly as this PR describes. The general trap-10 reasoning was sound; the remedy was inverted — the field needed a **reader**, not deletion.

**⭐ Trap 21 — both predicates read at the bytes, and they differ as claimed.** The upstream guard (`serverGraphHydration.ts:130–137`) is `currentId !== null && currentId !== undefined && currentId !== scenarioId` — it answers *"is this in-flight response still for the scenario the user is on?"*, and it sits **after** the `status !== 'graph'` return, so it never runs on the path where the defect lives. Note the sharper point: **that guard is itself a negative test that passes when `currentId` is null** — i.e. it would not have caught the reset case even had it run. The render needed a different question and a positive answer.

---

### 2. The positive-match gate and the null/null hole — verified

Gate is `typeof recordedScenarioId === 'string' && recordedScenarioId === currentScenarioId` — a positive equality, not `!==`. I constructed the null/null case via **M3** (below): dropping the `typeof` conjunct REDs exactly `BOTH ids absent is NOT a match — the null===null hole stays shut`, **failing by assertion** (`expected <section …> to be null`), not by throwing. The hole is genuinely pinned.

Two edges I checked and cleared: empty-string ids are unreachable on both sides (store side passes `UUID_RE`; canvas side is `row.id`/`scenario.id`/`crypto.randomUUID()`/`isUUID`-guarded/null), so `'' === ''` cannot produce a false match. All hooks precede the early return; the hooks ratchet confirms no new conditional hooks.

---

### 3. ⭐ The self-introduced trap-13b — repair proven in BOTH directions

This is the strongest evidence in the review. With the empty-state guard mutated (**M5**):

| Spec form | `renders nothing only when there is genuinely nothing on file` |
|---|---|
| **post**-repair (records THIS decision, no content) | **RED** |
| **pre**-repair (store left empty) | **GREEN — 52/52** |

So the shadowing was real (the test was vacuous, passing via the new gate rather than the guard it names), and the repair genuinely restores its discrimination. A single direction would have shown neither.

---

### 4. Mutants — re-run independently

Throwaway worktree at an **absolute** path outside the repo root. Isolation **proven by writing**, not by locating (trap 9g): inodes differ (`585769045` vs `585997438`), a sentinel appended to the worktree left the source SHA-256 **byte-identical**. Restores from a pristine **archive** outside both trees — never `git checkout --`, which restores from the index (trap 9h). Applied-check scoped to `src/`, aim asserted before believing any result.

| # | Aim asserted | Applied | Result |
|---|---|---|---|
| M0 | control, unmutated | **0** | GREEN 52/52 |
| M3 | `typeof` conjunct dropped → bare equality | 1 | **RED — the null===null test only**, by assertion |
| M5 | unrelated empty-state guard neutered (identity gate confirmed untouched) | 1 | **RED on its own test only; all 5 identity tests GREEN** |

**M5 is a genuine discriminating partner** — it proves the identity tests bind to the identity gate specifically rather than to any change in the component. Trailing control after restore: clean tree, 52/52. Main clone untouched throughout (`git status --porcelain src/` = 0).

The five identity-gate tests were asserted collected **by name** (5/5), never inferred from the suite total.

---

### 5. ⚠ The reachable set — one row is OVERCLAIMED, and I could not drive staging

**I did not obtain a driven browser witness.** Every page-reading browser tool returned `Policy check temporarily unavailable` for the whole session (~12 attempts, multiple tabs, `get_page_text`/`read_page`/`find`/`screenshot`/`javascript`; `navigate` succeeded, reads never did). That is an infrastructure outage on my side, **not** evidence either way. **The non-reset reproduction remains unwitnessed, and this review does not raise the P0's severity on driven evidence.**

What I could establish about the **deployed artefact** — I crawled the served chunk graph from the entry asset (87 chunks fetched; `what-i-was-given` present in exactly one, `/assets/ReactFlowGraph-BHo7vSOi.js`). At `54030415`, the served bytes are:

```js
const Ine = { briefText: null, manifest: null }          // no scenarioId field at all
Mne = Et(e => ({ ...Ine,
  setContextIntegrity: ({briefText:t, manifest:n}) => e({briefText:t, manifest:n}),  // no scenarioId
  reset: () => e({...Ine}) }))

function Une({onSendMessage:e}={}) {
  const i = Mne(e => e.briefText), s = Mne(e => e.manifest)
  if (null === i && null === s) return null              // the ONLY gate
```

That is deployed-bytes confirmation that the store carries no identity and the render's sole gate is the empty-state one — one rung below a journey witness, but a real derivation about what users are served.

**⭐ The overclaimed row: "Virgin profile, second decision, no reload — LEAKY".** My trace says this is **not** reachable as stated. Scenario ids are **lazily minted at turn-send** (`useConversation.ts:2937–2944` and `:4278–4285`), so a brand-new decision's single hydration attempt fires *before* CEE has a graph for it → `absent` → **nothing written**. On a strictly virgin profile with one continuous canvas mount and no reload or re-navigation, the store is **never populated**, so the receipt is *absent*, not wrong.

The governing precondition is cleaner than the row list, and I suggest replacing the table's framing with it:

> **The leak requires the store to have been populated at least once in the page session — which happens only on a `graph` cold read — followed by a decision whose cold read is not `graph`.**

Population happens on **boot into an existing decision** (`ReactFlowGraph.tsx:1460–1463` restores `currentScenarioId` from the persisted id, and hydration then reads a graph CEE already has), or after a route unmount/remount, which releases `attemptedRef` (`useServerGraphHydration.ts:38,71`) and lets a decision CEE now knows be re-read.

This **narrows** the enumeration; it does not soften the P0. The witnessed reset-canvas path stands, and the commonest real shape — reload or return into an existing decision, then start or switch to another — is squarely leaky. But per trap 20, a traced row generalised in the record is what every later session inherits, so please correct that one row (and mark the table rows still `traced`, since no live drive was performed by either of us).

---

### 6. Judgement on the disclosed limitation — accept, and row it

**Accept for this PR; do not add re-hydration here.** The reasoning that decides it: the suppression is **co-extensive with having nothing truthful to show**. Content is written *for* a decision, so whenever a truthful receipt exists for the live decision the gate passes and nothing is suppressed. There is no reachable state where correct content for the current decision sits in the store and is hidden. So the limitation is not "the receipt can go permanently missing" — it is "a decision whose graph lands on CEE *after* its single hydration attempt shows no receipt until a reload".

That is a **missing improvement, not a lie**, and truthful absence is the right ordering. Re-hydration needs a retry trigger keyed to the server graph appearing, which races the boot path the lane rightly declined to invent inside a P0 fix. **Row it; don't block.**

---

### 7. The sibling — class confirmed, and it is worse than stated

Manifest verified exactly as claimed: `useStrengthenStore.getState()._reset()` has **six non-declaration hits, all in `__tests__`** (`ownedLeaderClaim.strengthen.spec.tsx:312`, `StrengthenContainer.spec.tsx:42`, `analysis-hero/interaction.spec.tsx:47`, `switchProbabilityPresence.spec.tsx:138`, `strengthenStore.reconcile.pin.spec.ts:40`, `strengthenStore.spec.ts:25`), plus the declaration/impl at `strengthenStore.ts:63,236`. `StrengthenState` carries **no `scenarioId`**; it never reads `contextIntegrityStore` (zero hits); the copy is `strengthenCopy.ts:8`. **Same defect class — confirmed. Different store — confirmed.** Causation for the witnessed 5→1 is correctly left unproven; I did not attempt it and it should stay unproven.

**⭐ One thing to add to that row:** `strengthenStore` persists to **`sessionStorage`** (`strengthen.lifecycle.v1`, `:66–89`), whereas `contextIntegrityStore` is plain in-memory. **The one remedy that cleared this P0 — a hard reload — does not clear the sibling.** Its exposure window is therefore strictly wider in duration, and the row should say so.

---

### 8. Gates — run at the head in the required job's step order

Required check derived, not assumed: branch protection on `staging` requires **`Staging Gate`**, which needs `tsc` · `typecheck-selftest` · `vitest` (4 shards) · `vitest-summary` · `build`. The `tsc` job runs **Lint first**, then typecheck, then guards (trap 22e) — run in that order:

- **Lint: 0 errors** (1264 warnings, non-failing) · hooks ratchet **exactly at baseline, 232 violations / 15 files**
- **Typecheck ratchet: 2487 = baseline 2487**; coverage 3345 / 3385 tracked files — gate PASSED
- `ci:guard:zustand` (relevant — this PR adds a selector), `ci:guard:starters`, `check:vendor`, `ci:guard:schemas-version`: **all pass**
- Target spec **52/52**, identity-gate tests asserted by name (5/5)
- Blast-radius regression, `src/components/results` + `src/canvas/hydrate` + `src/canvas/stores`: **277 files / 4804 tests, 0 failures**, no collect-time failures
- **CI on the exact SHA** `c455f065`: `total_count = 13`, **running = 0**, completed-and-not-success = **none**, `Staging Gate` success; `mergeable_state: clean` (not DIRTY)

Full diff read across all four files; nothing undisclosed.

---

### Verdict

**APPROVE.** Twelve effective lines, correctly aimed: the gate is positive, the null/null hole is pinned by a test that a mutant proved was missing, and the self-introduced shadowing is repaired with discrimination demonstrated in both directions. The lane disclosed both of its own instrument failures — that is the standard.

Before merge, one **record** correction (no code): fix the "virgin profile, second decision" row per §5 and keep the untested rows marked `traced`. Then two rows to file: **re-hydration when the server graph appears after the single attempt** (§6), and **`strengthenStore` scenario keying, noting the sessionStorage persistence that defeats the reload escape hatch** (§7).

PR #682, opened 2026-08-13
P0: remove the PLoT bearer from the client bundle; add a value-shape sentinel

Talchain:
## Lead

**A live PLoT bearer is served to every visitor of staging right now**, as a 140-byte public asset (`/assets/plotAuthHeaders-Bazgbw-s.js`, credential `sha256:e10e8e9d`). It is not an orphan awaiting cache expiry — it is a **live static import referenced 15 times across 14 of the 92 chunks** of the currently-deployed build. Every visitor's browser fetches it automatically as part of normal page load.

**This PR stops future publication. It does not un-publish what is already out.** Rotation is mandatory and is Paul's action — see `SECURITY-PLOT-BEARER-ROTATION.md` in this PR.

> **Update, 14 Aug 2026 — rebased onto `828ff369` and extended.** Three things changed since the first round, each described in its own section below: the two absolute-base escape hatches are now **retired rather than normalised**; the proxy has gained the **upstream path allowlist** it shipped without; and the bundle proof now carries a **before/after control**. One premise this PR previously stated has been refuted and is corrected in place.

## Is the credential still in the emitted bundle after the change?

**No — proven by rebuilding and scanning the output, and now against a BEFORE control.** Both builds were run with `VITE_PLOT_BEARER`, `VITE_PLOT_ENGINE_URL` and `VITE_CEE_DRAFT_BASE` **all SET to bypassing values**, so this measures the mechanism rather than an empty environment.

| Probe | BEFORE (`828ff369`) | AFTER |
|---|---|---|
| **synthetic bearer VALUE in `dist/`** | **1 file** | **0** |
| **host literal `plot-lite-service-staging.onrender.com`** | **4 files** (3 JS chunks) | **0** |
| **`plotAuthHeaders` chunk emitted** | **1** | **0** |
| live credential by SHA-256 prefix (749 tokens) | 0 | 0 |
| the *name* `VITE_PLOT_BEARER` | **0** | 1 (a prose comment in `_redirects`) |
| *self-test* `/bff/engine` (must be > 0) | 6 | 7 |
| *self-test* a baked Supabase value (must be > 0) | 2 | 2 |

**The row that matters:** the *name* reads **0 in the BEFORE build — the one where the credential is genuinely present.** Vite compiles the name away and inlines the value, so `grep VITE_PLOT_BEARER dist/` was a confident false all-clear for the defect's entire life. That is now demonstrated with a control rather than asserted.

**Scope stated precisely:** the hash probe reading 0 in *both* columns is not evidence that the probe works — this lane never possesses the live value and cannot build a positive control for it. Its only claim is that no token in either bundle hashes to the compromised prefix. The load-bearing evidence is the **synthetic bearer** row, which does have one.

## The two escape hatches are now RETIRED, not normalised

**This corrects a claim the earlier version of this PR made.** It said `plotSameOrigin.ts` normalises the two absolute-base escape hatches "so those variables may stay set without routing around the boundary."

**That was true only of the values the variables were measured to hold.** `toSameOriginPlotBase` rewrites a base whose host is in the PLoT family and passes **every other absolute base through untouched** — correct for the normaliser, and the v2 spec's own contrast control asserted exactly that with `https://elsewhere.example.test/api`. So a dashboard value pointing at any host that is not literally PLoT's still took `/v2/run` — the primary analysis path — clean off-origin and clean around the credential boundary, **and the suite agreed**, because the pass-through was written as desired behaviour of the normaliser rather than examined as a property of the run path.

Bounding an override is not the same as not having one. Both reads are gone:

- `src/adapters/plot/v2/adapter.ts` — `VITE_PLOT_ENGINE_URL` read deleted; the base is the caller's, normalised.
- `src/adapters/cee/client.ts` — `VITE_CEE_DRAFT_BASE` read deleted; `CEE_DRAFT_ENGINE_BASE = '/bff/engine/v1/cee'`.

Pins bind to the **fetched URL**, plus comment-stripped source pins with survival controls, because under test both old fallbacks were already relative and no behavioural fixture could tell an unset read from an absent one.

### Refuted premise: the draft-graph client is NOT unmounted

The standing verdict was RETIRE, on the grounds that its product call sites are unmounted. **Three of the four chains are indeed dead** — `AIClarifierChat`'s gate has no `true` setter anywhere in `src/`; `DraftChat` is unmounted by `aiPanelV2` (default ON) *and* short-circuited by `VITE_ENABLE_ORCHESTRATOR_V2` in `[build.environment]`; `useRetryDraft`'s button needs a store value only `DraftChat` writes.

**The fourth is reachable with no env var set.** `PayloadLabTab` → "Run Draft" calls `draftModel` directly. `debugPanelVisibility.ts:20` reads `VITE_APP_ENV || 'development'`, and `'development'` is an allowed env, so the gate reduces to a URL parameter: `?diag` on any route. The panel mounts at the app shell **outside `<Routes>`**, so `DevRoutesGuard` does not apply.

So retiring the *client* would have deleted a reachable surface. Retiring the *env read* achieves the credential objective without that. Whether that debug surface should be reachable in production is a separate question with a separate owner — it is not a credential defect and is not decided here.

## The proxy shipped without a path allowlist — it has one now

`plot-proxy.ts` was **the only one of five credential-injecting edge proxies with no upstream path bound**. `cee-proxy`, `orchestrator-proxy`, `collab-proxy` and `isl-proxy` each 404 an off-list path and never let the key leave the edge (disposition item 13). This one forwarded **any** path under `/bff/engine/*` with the bearer attached — every browser an authenticated caller across PLoT's whole surface. Narrower than a published credential, and the same blast-radius shape: moved to the edge rather than removed.

**19 anchored literals**, derived from the UI's call sites, matched before `PLOT_AUTH_TOKEN` is read, with a traversal check ordered first because the dynamic segments are deliberately permissive (`[^/]+`) so real run/template ids never false-404.

**The list is derived, not hand-maintained.** It has to be a literal inside the edge function — Deno, no `src/` imports — so the derivation lives in a guard instead: `plot-proxy-path-allowlist-derived.spec.ts` re-derives every browser→PLoT call site from source each CI run and fails **both ways**: a new call site nobody allowed, and an entry whose caller is gone. One-directional coverage proves the two copies *agree* and can never notice the list is too wide.

Unresolvable call sites are **RED and listed**, never skipped — a scanner that silently drops a shape it cannot read is an absence probe with no positive control.

Two scanner defects were found while building it, both of which manufactured false absence and would have **deleted live routes** under a green suite: a `[^,)]+` first-argument regex truncating at the paren inside `encodeURIComponent(id)`, and identifier bindings resolving to the *first* match in the file, so `http.ts`'s three `const endpoint` sites all reported `/v1/run`.

## The boundary

`netlify/edge-functions/plot-proxy.ts` serves `/bff/engine/*` and `/engine/*`, injecting `PLOT_AUTH_TOKEN` server-side. **Inert until Paul sets it**: it forwards unauthenticated, marks the response `X-Plot-Proxy-Credential: absent`, and never embeds a value.

**What Paul must set:** `PLOT_AUTH_TOKEN` in the Netlify UI site's environment variables (**not** `VITE_`-prefixed), **before this merges**. Then delete `VITE_PLOT_BEARER`, `VITE_PLOT_ENGINE_URL` and `VITE_CEE_DRAFT_BASE`. Then rotate. Full order, verification commands and unknowns in the runbook.

## Residual seam of the same class — reported, not fixed

Ten further env-resolved PLoT bases exist (`VITE_PLOT_PROXY_BASE` × 9, `VITE_BFF_BASE` × 1, in `lib/config.ts`, the five `adapters/plot/v1/*` resolvers, two canvas hooks and `SafeMode.tsx`). **All ten default to `/bff/engine`, and neither variable is known to be set to a bypassing value** — materially unlike the two retired here, one of which was *measured SET* to PLoT's absolute origin. Latent rather than live, ten files wide, with local-dev override semantics to preserve. Rowed for a separate lane rather than expanded into this one.

Even if one were pointed at PLoT's origin, the authenticated surface is now bounded by the allowlist, and a cross-origin call carries no credential because the browser holds none.

## The sentinel

`scripts/ci/assert-no-bundle-credentials.mjs` scans the **emitted bundle** for credential-shaped **values** and fails. Wired into the **Staging Gate** build job — the only check `staging` branch protection requires (re-derived at this tip: `["Staging Gate"]`) — not merely `ci.yml`, which is advisory.

Public-by-design values are separated by reading the artefact, never by widening the pattern until it stops firing: a **JWT is decoded** and its `role` claim read (`anon` passes, `service_role` fails loud); a **content digest** is identified by the key it is assigned to, because a SHA-256 and a hex secret are byte-indistinguishable. The digest exemption has a **discriminating twin**: the same value under a non-digest key still fails.

A **runtime self-test** runs before any verdict — a positive canary must be caught and its credential-free twin must not — so a detector that has silently stopped matching cannot read as a clean bundle.

## Two corrected premises from the first round (unchanged, still true)

1. **`netlify.toml:207` was the DEAD twin.** The live rule was `public/_redirects:25`, which Netlify processes first. A change made only to `netlify.toml` would have done nothing. Both are now removed in favour of the edge function.
2. **The allow-list did not silently drift — it explicitly sanctioned the credential** via `knownExposed.VITE_PLOT_BEARER: { valueStillInlined: true }`, and the guard **printed a 🚨 about it on every run, then exited 0**. A loud log that returns zero is not a gate.

## Gates

- Typecheck gate: **PASSED** — `2486` against baseline `2486`.
  - The rebase auto-merged row edits from both sides of the baseline while keeping one side's `# count=` header, leaving it internally inconsistent (`count=2485`, rows summing to 2487) and hard-failing the gate. Pristine `828ff369` passes at 2485, proven in a control worktree, so the break was the merge artefact. Regenerated delta vs staging: **`+2 plot-proxy.ts`** (cee/collab/orchestrator-proxy each carry exactly 2 for the same Deno-globals reason) and **`−1`** for a pre-existing error this PR fixes. `2485 − 1 + 2 = 2486`.
- Credential sentinel against the built `dist/`: **PASSED**, self-test green, 102 files / 6,145,589 bytes scanned.
- **Mutation kit: 8/8 bitten, 0 survivors, 0 unapplied**, controls exactly 0 before and after, trailing control green, collected counts stable at 66. Isolation proven by a **write test** (source SHA unchanged), not by locating the worktree.

Four instrument defects were found and are written up in the evidence directory, because each produced a clean, plausible, wrong answer and none was visible by inspection — including a first kit run that reported three **false survivors** because the worktree's HEAD predated the spec under test (34 collected instead of 45).

Evidence: `olumi-docs/PHASE0-EVIDENCE-2026-07-28/credential-migration-682-2026-08-14/`

## Watch after deploy

**Drafting.** Its base was the one genuinely cross-origin call and now traverses the proxy. Historical latency ~56s; the absolute base existed to bypass a ~28s proxy timeout. If drafting starts timing out, that is the cause, and the fix is server-side — never a credential back in the client.


Talchain:
## SUPERSEDED — closing. The whole of this PR is live on staging, landed by #723 + #730.

Re-derived independently today (17 Aug) against **deployed UI `37f626b6`** (`/version.json`, matching `origin/staging` tip exactly), not against this branch's premises.

### The credential is gone from the served bundle

Closed crawl from the entry asset: **84 chunks, 5,915,740 bytes**, closure reached at round 8 (completeness sweep: every chunk-shaped name referenced anywhere in the graph was fetched).

| Probe | Deployed `37f626b6` |
|---|---|
| `plot-lite-service-staging.onrender.com` | **0 files** |
| `plotAuthHeaders` | **0 files** |
| `VITE_PLOT_BEARER` / `PLOT_AUTH` / `VITE_PLOT_AUTH` | 0 files *(and this row proves nothing on its own — Vite compiles the name away; see below)* |
| opaque secret-shaped literals (`[A-Za-z0-9]{48,}`) | **5, all `captureSha256:` content digests** — none matches the compromised prefix `e10e8e9d` |
| `Authorization` literals | 4 files, all Supabase-SDK / user-session-token / collab-token — **no baked credential** |
| `/assets/plotAuthHeaders-Bazgbw-s.js` | no longer served — 200 is the **SPA fallback** (`content-type: text/html`, body is `index.html`), not the chunk. The runbook's own warning that *status code alone proves nothing here* is respected: discriminated by body, not by code. |

**Controls, as this PR's own doctrine requires.** Positive: `/bff/engine` = 6 files/14 hits · `cee-staging.onrender.com` = 1/6 · `Authorization` = 4/12 · and the opaque-token scanner was proven sighted by injecting a synthetic 64-char token into a copied chunk (caught, `sha256=6666b638`). Contrast: `plot-lite-service-FABRICATED.onrender.com`, `VITE_PLOT_ZZQNOTREAL`, `zzq-fabricated-bearer-token` = **0 each**. Scope limit stated: the bundle carries **zero `eyJ` literals**, so the JWT-decode branch of the scan has no positive control here and makes no claim.

### The boundary is live and enforcing

| Unauthenticated request to the UI origin | Result |
|---|---|
| `/bff/engine/v1/templates` | 200, `x-plot-proxy-credential: injected`, `x-render-origin-server: Render` |
| `/bff/engine/v1/limits` | 200 + `injected` |
| `/bff/engine/v1/health` | 200 + `injected` |
| `/bff/engine/v1/zzq-fabricated-route` | **404 at the edge** — no credential header, never reached Render |
| `/bff/engine/v2/run` | **404 at the edge** |

The allowlist is real and discriminating (contrast pair above), `PLOT_AUTH_TOKEN` is set and injected server-side, and the browser holds no credential.

### The analysis path

The 14 Aug finding that the shipped bundle carries a full PLoT client building `/v2/run` **no longer holds at this tip.** `/v2/run` survives in exactly 2 chunks: a URL→service *classifier* (`types-*.js`) and a DebugPanel display default (`"/plot/v2/run"`). `runV2` / `runV2Analysis` / `plotClient` / `engineBase`: **0 files.** `/v1/cee/draft-graph`: **0**. The draft client now hardcodes `"/bff/engine/v1/cee"` when `window` is defined; `limits`, `templates`, `health` and `pre-analysis-sensitivity` are all relative `/bff/engine/*`. `OutputsDock`'s previously ungated `runV2Analysis()` drop-through is now an explicit refusal. Non-test `import.meta.env` reads of `VITE_PLOT_BEARER` / `VITE_PLOT_ENGINE_URL` / `VITE_CEE_DRAFT_BASE` in `src/`: **zero** (regex positive-controlled at 13 hits for `VITE_SUPABASE_URL` / `VITE_PLOT_PROXY_BASE` / `VITE_BFF_BASE`).

### This PR versus the tip, file by file

25 of 27 paths present; the 3 absent are the two this PR *deleted* (`src/lib/plotAuthHeaders.ts` + its spec — deletion achieved) and `runV2.plotBearer.spec.ts`, whose subject #730 retired outright. Of the 25 present, **13 are byte-identical to this branch**, including `plotFetch.ts`, `plotSameOrigin.ts`, `client.ts`, `public/_redirects`, the sentinel, its `.d.mts`, the env allowlist, the rotation runbook and both CI workflows. Every file that differs, differs because **the tip is ahead**:

- `plot-proxy.ts` — the tip's allowlist is **narrower**: this branch still allows `/version`, `/v1/run`, `/v1/run/:id/cancel`, `/v1/run_bundle`, `/v1/stream`, `/v1/validate`, `/v1/diff`, `/v1/analysis/pareto`, `/v1/analysis/sequential`, `/v1/explain/policy`, `/v1/suggest/edge-function` and `/v2/run`. All retired by caller-deletion at the tip.
- `adapters/plot/v2/adapter.ts` — tip is **370 lines shorter** (the HTTP transport section is gone).
- Both allowlist guard specs — tip adds +129 and +43 lines.
- Typecheck baselines — tip carries 135 and 53 **fewer** entries.
- `netlify.toml` / `package.json` — unrelated drift the tip owns (schemas `0.40.0` → `0.46.0`, analysis-hero flag retirement).

The sentinel is wired into **both** gates at the tip, not just one: `staging-full-tests.yml:555` (the Staging Gate — the single required check) **and** `ci.yml:446`.

### Nothing is lost by closing, and one thing is still owed

There is no salvageable delta. Merging this branch would *widen* a boundary the tip has since narrowed, and it is `DIRTY`/`CONFLICTING` against a base ~60 commits ahead.

**The rotation remains outstanding and is Paul's alone.** Removing the credential from the bundle does not un-publish what every visitor already fetched. `SECURITY-PLOT-BEARER-ROTATION.md` is at the tip, unchanged since it landed in #723, and still reads *"the rotation is NOT done and is Paul's action"* — including its §2 step 4 warning that the value is recorded as shared with CEE, so PLoT must not be rotated without updating CEE in the same window. This session verified the code boundary only; it did not and must not touch rotation.

Closed as superseded by #723 (`7ad0cefc`) and #730 (`6c009997`).

PR #705, opened 2026-08-14
feat(canvas): the element the answer was grounded on lights up on the canvas (hop 4b)

Talchain:
**LANDS TOGETHER WITH → Talchain/olumi-assistants-service#965.** That PR puts `_grounded_selection`
on the wire; this one consumes it. **#965 merges first** (the field must exist), this immediately after.
Neither ships alone: a resolved field with no consumer is this estate's dominant defect, and a consumer
with no field is inert.

## What a user now SEES

Select a node → ask naturally → the answer comes back grounded in that element and its analysis outputs
(hop 4, already merged) → **that node is marked on the canvas as the thing this answer is about**, riding
the canvas's existing "the AI is talking about this" treatment.

Ask the same question with **nothing selected** and **nothing marks** — on a turn that carries no sidecar
at all, byte-identical to every pre-hop-4b turn.

## The chain, file:line

| Hop | Where |
|---|---|
| Demoted to `__additive__` | `src/v5/responseParser.ts:217-235` (existing, unchanged) |
| Read + re-validated | `src/v5/groundedSelection.ts` `readGroundedSelection` |
| ⚠ THE CONSUMER CALL SITE | `src/v5/applyV5State.ts` — `store.setGroundedFocus?.(readGroundedSelection(response))` |
| Store slice + marks | `src/canvas/store.ts` — `groundedFocus` + `setGroundedFocus` |
| Rendered | `src/canvas/nodes/BaseNode.tsx:77` → `:249` — existing `ring-4 ring-info/60 ai-highlight-pulse` |
| Honesty notice | `src/canvas/components/GroundedFocusNotice.tsx`, mounted `ReactFlowGraph.tsx` beside `FocusModeChip` |

**The carrier is proven by two working consumers, not by inspection.** The question that kills slices like
this is *does the non-enumerable sidecar actually reach the reader?* — any spread on the way loses it.
`applyV5State` has **exactly one** production call site (`useConversation.ts:4736`, receiving
`target.response` from `{ ...useCanvasStore.getState() }` — a spread of the live store, which is what
carries `setGroundedFocus` through), and **two existing live sidecar readers already read `__additive__`
off that same object**: `extractReasoningSidecar` (`:5058`) and `extractAnswerShapeSidecar` (`:5070`).

## Design constraints that are structural, not incidental

- **Nothing writes `highlightedEdges`.** `FocusModeChip` renders *"Showing paths from X to goal"* on
  `highlightedEdges.size > 0` alone. A grounding involves **no path**, so writing that set would make the
  chip state something untrue. This is **#694's neighbourhood-focus honesty precedent transposed**, and it
  is guarded by an **additive** mutant (add the edge write ⇒ the guard REDs), not satisfied by accident.
- **The AI never moves the user's pointer.** `selection` is untouched; `groundedFocus` is its own slice and
  the two are allowed to diverge.
- **The clear path is unconditional.** Every turn writes the slice, so an ungrounded turn clears the
  previous answer's marks — stale attention is worse than none. It clears only marks this slice owns, so it
  cannot wipe an applied-edit pulse.
- **No new visual language, and no second focus authority** (trap 21): the marks ride the existing
  `highlightedNodes` channel. Two emphasis styles on one canvas is a worse outcome than none.
- **No scroll-jacking.** Highlight in place; the canvas is never moved under the user.

## `not_in_model` and `could_not_check` must not collapse — and needed a COMPONENT

The spec asked for the two states to be "visibly different". **With node marking alone they are not: both
mark nothing.** Honouring it required a surface — `GroundedFocusNotice` — whose entire reason to exist is
that discrimination. `could_not_check` says *"I couldn't read your model to check what you selected"* and
claims nothing about the element either way; `not_in_model` renders **nothing at all**, because the absence
of a mark IS the honest answer there.

## Verification

**RED-first at pristine** (implementation reverted to `staging`, my specs retained, in a throwaway worktree
**outside** the repo root): **16 failed / 11 passed**, named signatures —
`AssertionError: expected [] to deeply equal [ 'node-engineer-salary' ]`,
`TypeError: useCanvasStore.getState(...).setGroundedFocus is not a function`.
The 11 that pass at pristine are the fail-closed and negative-control cases, which are **vacuous without
the wiring** and meaningful only with it — stated rather than counted as coverage.

**Isolation proven by WRITING, not by locating** (trap 9g): a sentinel appended to the worktree left the
source clone's `git status` empty, and inodes differ (`623869068` vs `624333135`). Restores are
HEAD-relative (trap 9h). Leading **and** trailing controls: applied-check scoped to `src/` reads **exactly 0**,
27/27 green both times.

**Six mutants, all biting, with DISTINCT red sets** (uniformity would be evidence about the probe, trap 20):

| Mutant | Effect | Result |
|---|---|---|
| **m1 wiring (F6)** | delete the `setGroundedFocus?.(...)` call site | **9 RED** |
| m2 edges (additive) | write `highlightedEdges` in the grounding | **1 RED** — the chip-honesty guard |
| m3 collapse | notice speaks for `not_in_model` too | **2 RED** — incl. the discrimination pair |
| **m4 identity** | mark a FIXED id instead of the grounded one | **6 RED** |
| m5 clear | ungrounded turn stops clearing | **1 RED** |
| m6 mount (trap 3b) | put the notice behind a flag | **1 RED** |

⭐ **m4 is the discriminating RED/GREEN pair**, not a single biting mutant: *"marks the node the answer was
grounded on"* stays **GREEN** under a fixed-id implementation while *"marks a DIFFERENT node"* **REDs**. A
single biting mutant proves sensitivity to *something*; only the pair proves sensitivity to the **named
object** (trap 19).

⚠ **A harness defect caught mid-run, disclosed:** the first battery returned **identical empty output for
every mutant**. Cause was the documented zsh trap — an unquoted `$SPECS` is not word-split, so vitest
received one giant argument, found no test files and **exited 0**. Re-run with explicit arguments and a
**hard error on zero collected**. The first battery's results were void and are not reported above.

**Gates at my tip:** `pnpm typecheck` (`scripts/ci/typecheck-gate.sh`) — **PASS**, 618 files / 2485 errors,
**exactly baseline, no drift**. It caught a real error in my own spec (a missing `anchorPosition`), fixed at
the source and **not** absorbed into the baseline. Both spec files asserted to collect **by name**: 27 tests,
non-zero (trap 2b — a new spec collecting zero is invisible to every aggregate). UI suites run with
`VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` exported.

**Typecheck baseline note (instrument property):** the identity baseline moved by exactly one line — a
PRE-EXISTING `TS2345` whose identity string embeds an *elided property count* (`... 254 more ...` →
`... 256 more ...`). Adding any store action shifts it, manufacturing a phantom "1 appeared / 1 disappeared"
pair. Totals unchanged. **Every store-slice PR will produce this phantom pair.**

**Trap 3b** — this estate has shipped the same badge dark twice by testing a component the deployment does
not mount. The notice's mount path is asserted directly: it sits in the same **unconditional** chip rail as
`FocusModeChip`, with a positive control on the chip itself. **Scope stated precisely: that is a structural
claim about `ReactFlowGraph.tsx`, not a render proof of it.**

⚠ **jsdom cannot prove visibility** (trap 3). Every jsdom claim here is about DOM presence or store state.
The visibility claim is made in a browser, separately, and is reported with its own scope.

## Ladder — stated honestly

CODE EXISTS ✅ · TESTED ✅ · DEPLOYED ❌ · MOUNTED ❌ · WIRE-WITNESSED ❌ · JOURNEY-WITNESSED ❌

**The amended acceptance is NOT met by these two PRs.** They complete the **visible half** — the model
context focuses on screen. The answer-side wire witness (that the answer text demonstrably reasons about
the element) is a post-merge witness against the deployed build, ordered separately. #960's CEE/wire half is
already journey-witnessed; this is the half a user can see.

⚠ **Paired follow-up, sequencing:** register row **2.1219** — when a selection resolves to nothing, the
model silently adopted the leader in 3 of 6 fabricated-id turns, and **a visible focus makes that worse on
screen**. Its fix is a disclosure tail at CEE's compose seam, which is **#964's** (open at my tip), so it is
**not** in this slice. It should land close behind.

---

# UPDATE — a HIGH defect was found and fixed; the body above predates it

## The capability did not hold on its own headline journey

An adversarial capability review refuted this slice's ownership claim at the bytes. The marks rode
`highlightedNodes`, on the reasoning that it is already the canvas's "the AI is talking about this"
channel. It is — **and it is shared**:

- **Eleven other writers** use it (`appliedEditPulse`, `HighlightContext`, `useHighlightDispatch`,
  `PreAnalysisHealth`, `TransitionCard`, `CompareState`, node components). A wholesale replace trampled them.
- **`appliedEditPulse.flush` REPLACES the set (`:84`) and then EMPTIES it unconditionally at
  `PULSE_DURATION_MS` (`:90`).** So on any turn that both applied a patch and carried a selection —
  **"change THIS one", the flagship case** — the grounded marks were overwritten ~100ms later and erased at
  ~2.1s, with `groundedFocus.nodeIds` still reporting them and nothing reading it to restore them.
  **The store disagreed with the pixels.**
- The clear guarded on whether **this slice** was non-empty, not on whether the marks on screen were still
  the ones it wrote — so an ungrounded turn could wipe another writer's marks, **the exact thing its own
  comment claimed it could not do.**

An independent probe confirmed the same defect on four timelines, including a **cross-turn** one that needs
no coincidence at all (edit, then ask a grounded question within 2s), and proved the obvious fix does not
work: reordering the calls produces a **byte-identical timeline**, because the grounded write is synchronous
while the flush is +100ms and the clear +2000ms.

## The fix: its own channel (option (a))

`setGroundedFocus` now writes **only its own slice**. `BaseNode` reads `groundedFocus.nodeIds` **alongside**
`highlightedNodes` and applies the **same** `ai-highlight-pulse` treatment — one emphasis language on the
canvas (the spec's instruction), two independent pieces of state, **no lifetime negotiation**. That is what
"single authority" actually requires: owning your own slice, not writing someone else's.

Proven by revert-mutant: restoring the clobbering writes REDs **exactly** the three new ownership guards and
nothing else.

## The suite was structurally blind to all of it — the more important finding

Every test reset `highlightedNodes` to empty in `beforeEach` and none exercised a patch-bearing turn, so a
wholesale replace and an ownership-respecting write were **indistinguishable to it**. 19/19 green on a turn
class that structurally excluded the defect. Five cases added, all from outside the author's head: another
writer's marks survive a grounded turn; they survive an ungrounded turn; the grounding survives a
same-turn pulse past its 2s clear; **the cross-turn timeline**; and an **edge-only** patch, which needs no
timer at all because `flush()` calls `setHighlightedNodes(nodeIds)` unconditionally and an empty node list
actively clears.

## The notice's honesty assertion was defeatable, and was defeated

It was a positive phrase match plus a **hand-written list of forbidden words** (trap 12). Rewriting the copy
to *"I couldn't read your model to check what you selected — that element has been removed from your model"*
satisfied the phrase, dodged all three words, and stayed **GREEN** — while making **the exact false claim
this component exists to prevent.** The whole sentence is now **bound exactly**; replaying that mutation now
REDs **both** guards.

## Still open — handed off, not silently carried

1. **The mount assertion is source-structural and defeatable.** The same dark outcome written as a ternary,
   or as a gate on the enclosing rail, would survive it, and its "positive control" only proves the file was
   read — not that the probe can discriminate a gated mount (trap 13e). **It should be replaced by a
   render-based mount assertion.** Its scope is stated honestly in the file, but the guard is weaker than a
   reader would assume.
2. **Ghost ids.** The wire spec's grounded ids do not exist in `store.nodes`, so it asserts store state
   rather than user-visible focus. Under the separate-channel architecture a ghost id is **inert at the
   pixel layer** (BaseNode can only mark a node that renders), so this is a weaker claim than a defect —
   but the spec should use ids that exist on the canvas, and `appliedEditPulse`'s fail-closed canvas filter
   is worth adopting for the same reason.
3. **Producer-side, unsettled (trap 16-inverse):** whether CEE emits `_grounded_selection` on the same turn
   as an applied patch is a question this lane did **not** settle, and it is not claimed either way. The
   cross-turn timeline needs no such coincidence, so reachability does not hinge on it.

## Gates at the new tip `2b39b704`

- `pnpm typecheck` — **PASS**, 618 files / 2485 errors, **unchanged**
- 34 tests, both files asserted to collect **by name**
- CI: **13/13 SUCCESS, `mergeable = CLEAN`** (includes the full suite across 4 shards, so these specs ran in CI)

### DIFF
--- a/scripts/ci/typecheck-baseline-identities.txt
+++ b/scripts/ci/typecheck-baseline-identities.txt
@@ -540,7 +540,7 @@
 2	src/canvas/conversation/useConversation.ts	TS2339	Property 'graph_state' does not exist on type 'TurnRequestPayload'.
 1	src/canvas/conversation/useConversation.ts	TS2339	Property 'message' does not exist on type 'TurnRequestPayload'.
 2	src/canvas/conversation/useConversation.ts	TS2339	Property 'selected_elements' does not exist on type 'TurnRequestPayload'.
-1	src/canvas/conversation/useConversation.ts	TS2345	Argument of type '{ currentResultsHash: null | string; backfillGoalThreshold: (analysisReady: { goal_node_id?: string; goal_threshold_raw?: null | number; goal_threshold_unit?: null | string; goal_threshold_cap?: null | number; } | null) => void; ... 254 more ...; clearClarifierPreview: () => void; }' is not assignable to parameter of type 'V5ApplicatorStore'.
+1	src/canvas/conversation/useConversation.ts	TS2345	Argument of type '{ currentResultsHash: null | string; backfillGoalThreshold: (analysisReady: { goal_node_id?: string; goal_threshold_raw?: null | number; goal_threshold_unit?: null | string; goal_threshold_cap?: null | number; } | null) => void; ... 256 more ...; clearClarifierPreview: () => void; }' is not assignable to parameter of type 'V5ApplicatorStore'.
 2	src/canvas/conversation/useConversation.ts	TS2352	Conversion of type 'OrchestratorResponseEnvelopeV2' to type 'Record<string, unknown>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
 1	src/canvas/conversation/useGraphEditEvents.ts	TS6133	'op' is declared but its value is never read.
 1	src/canvas/conversation/useGraphEditEvents.ts	TS6133	'ops' is declared but its value is never read.
--- a/src/canvas/ReactFlowGraph.tsx
+++ b/src/canvas/ReactFlowGraph.tsx
@@ -93,6 +93,7 @@ import { DocumentsManager } from './components/DocumentsManager'
 import { ProvenanceHubTab } from './components/ProvenanceHubTab'
 import { ConnectPrompt } from './components/ConnectPrompt'
 import { FocusModeChip } from './components/FocusModeChip'
+import { GroundedFocusNotice } from './components/GroundedFocusNotice'
 // EdgeLabelToggle moved to CanvasToolbar for cleaner UI
 import { LimitsPanel } from './components/LimitsPanel'
 import { BottomSheet } from './components/BottomSheet'
@@ -2154,6 +2155,12 @@ const ReactFlowGraphInner = memo(function ReactFlowGraphInner({ blueprintEventBu
         style={{ top: 'calc(var(--topbar-h, 0px) + 1rem)' }}
       >
         <FocusModeChip />
+        {/* Hop 4b: the "I couldn't read your model" notice. Shares the chip
+            rail deliberately — it is the same class of transient, canvas-level
+            status, and the two are mutually exclusive in practice (a grounded
+            turn whose model could not be read has no path highlighting).
+            Renders nothing for every other grounding state. */}
+        <GroundedFocusNotice />
       </div>
 
       {/* Influence Explainer - shown when triggered from TopBar dropdown */}
--- a/src/canvas/components/GroundedFocusNotice.stories.tsx
+++ b/src/canvas/components/GroundedFocusNotice.stories.tsx
@@ -1,0 +1,76 @@
+/**
+ * GroundedFocusNotice — the honesty pair, in a real browser.
+ *
+ * jsdom can prove the element is in the DOM. It cannot prove it is visible,
+ * laid out, or legible (CLAUDE.md trap 3), and the acceptance condition for
+ * this slice is that `could_not_check` and `not_in_model` are VISIBLY
+ * different. That claim can only be made here.
+ *
+ * ⭐ EACH STORY DRIVES THE REAL `setGroundedFocus` ACTION rather than writing
+ * the store slice directly, so what renders is the production path — a story
+ * that stubbed the slice would prove the component and say nothing about the
+ * action that feeds it.
+ */
+// `@storybook/react-vite`, not `@storybook/react`: this repo is on Storybook
+// 10 and only the former is installed. Every existing story imports the
+// latter and carries a baselined TS2307 for it — this one does not add a
+// seventh.
+import type { Meta, StoryObj } from '@storybook/react-vite'
+import { useEffect } from 'react'
+
+import { GroundedFocusNotice } from './GroundedFocusNotice'
+import { useCanvasStore } from '../store'
+import type { GroundedUnresolved } from '../../v5/groundedSelection'
+
+function Harness({
+  unresolved,
+  nodeIds = [],
+}: {
+  unresolved: GroundedUnresolved | null
+  nodeIds?: string[]
+}) {
+  useEffect(() => {
+    useCanvasStore
+      .getState()
+      .setGroundedFocus(unresolved === null ? null : { nodeIds, unresolved })
+  }, [unresolved, nodeIds])
+
+  return (
+    <div style={{ padding: '2rem', minHeight: 160 }}>
+      <GroundedFocusNotice />
+    </div>
+  )
+}
+
+const meta: Meta<typeof Harness> = {
+  title: 'Canvas/GroundedFocusNotice',
+  component: Harness,
+}
+export default meta
+
+type Story = StoryObj<typeof Harness>
+
+/** The graph could NOT be read. The notice must SPEAK — and must not say
+ *  "not found", which is the claim this component exists to prevent. */
+export const CouldNotCheck: Story = {
+  args: { unresolved: 'could_not_check' },
+}
+
+/** The graph WAS read and does not contain the element. Marking nothing is
+ *  honest, so the notice renders nothing — no empty box, no stray border, no
+ *  reserved space. Compare against `CouldNotCheck`: that visible contrast is
+ *  the acceptance condition. */
+export const NotInModel: Story = {
+  args: { unresolved: 'not_in_model' },
+}
+
+/** Everything the user pointed at resolved. The node marks are the answer; the
+ *  notice stays silent. */
+export const FullyResolved: Story = {
+  args: { unresolved: 'none', nodeIds: ['node-engineer-salary'] },
+}
+
+/** An ordinary turn, never grounded on a selection. Silent. */
+export const NotGrounded: Story = {
+  args: { unresolved: null },
+}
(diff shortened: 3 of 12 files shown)


Talchain:
**Reconciliation note (17 Aug) — SCHEDULE. Your blocking dependency merged; this is now a landed field with no consumer.**

`Talchain/olumi-assistants-service#965` **merged 15 Aug 01:29Z**, so `_grounded_selection` is on the wire. At UI staging tip `37f626b6` the consumer half is absent: `groundedSelection` = **0** production files, `GroundedFocusNotice` = **0**, `groundedOn`/`grounded_on` = **0** (globs excluding `__tests__`/`*.spec.*`; contrast controls in the same sweep: `grounding` = 19 production files, `grounded` = 32, so the probe discriminates). Seven of this PR's twelve paths do not exist at the tip, including `src/v5/groundedSelection.ts` and `GroundedFocusNotice.tsx`.

This PR's own body called the failure mode exactly right — *"a resolved field with no consumer is this estate's dominant defect"* — and that is now the live state, on the producer side of the pairing it warned about. Kept open and scheduled, not closed: this is unlanded capability, not stale work. It needs a rebase (branch is `DIRTY` against a base ~60 commits ahead) and a wire witness that the sidecar actually arrives, since the producer has been live for two days unconsumed.

Talchain:
Superseded by #747 (merged), which lands the grounded-selection consumer against the current tip.

Why a fresh build rather than this branch: #705's canvas-mark design predates #732-#746, which rewrote every surface it mounted into. #747 departs on three structural grounds derived at the tip — the notice is per-message so transcript history stays true (a canvas mark is one global slot and would sit beside every older answer); under node-marking the two unresolved enum states both mark nothing, so a second component was needed anyway; and #747 writes no store state at all, avoiding a third writer on `highlightedNodes`.

#705's anchors all still existed at the tip — this is a design supersession, not rot. Producer contract re-derived at CEE `bf4a1d28` (closed enum, key-absent-when-ungrounded, empty element_ids honest). #747 verdict: MERGE-CLEAN after adversarial review (11 mutants, discriminating pair confirmed, 28-shape malformed battery through the real chain, one-authority delegation to the existing `resolveElementLabel`).

PR #706, opened 2026-08-14
collab: the owner can cite the evidence that moved the model (0.41.0 UI half)

Talchain:
The UI half of **P2 — collaboration evidence/challenge → attributed model consequence**. Pairs with CEE [#966](https://github.com/Talchain/olumi-assistants-service/pull/966).

## What was dark

CEE can already **verify** a citation (binding (f)) and **stamp** it onto the graph. **Nothing could send one**: the apply button posted the two ids it always had, so the verified-and-stamped half was unreachable through the product's own click path — the estate's chronic failure #1, build it and never plug it in.

## What a user gets

On the owner's reveal, where evidence exists on a target, the apply now reads:

> Your model will record **Ada's note** as the reason for this change.
> **[ Use Grace's 0.85 ]**

…and the model records exactly that. **The value applied is Grace's; the reason cited is Ada's** — the asymmetry this hop exists for.

## The product rule: cite only when the choice is not a guess

| evidence on the target | sentence | claim |
|---|---|---|
| 0 | — | none |
| **1** | names it | claims **that** id |
| 2+ | — | **none** |

The stamp asserts the model changed *because of* that evidence. With two or more notes, picking one would be **the product inventing the owner's reason** and writing it into the graph as a server-verified fact — the fabrication class every binding in `apply-verification.ts` exists to refuse, arriving through the client instead of the wire. Honestly uncited beats dishonestly cited; the explicit picker is the follow-up.

And one note is not automatically *the reason* either. What makes the single case honest is not the arithmetic — it is that **the affordance discloses the citation before the click**, so the owner consents to a stated fact.

### ⚠ Record corrected (review)

This body previously said *"the sentence and the claim come from one call"*. **That was false** — there were **five** call sites, and a code comment claiming they were "not a second computation" was exactly what it denied being. The property did hold, but **for a reason the argument never stated**: `citableEvidenceFor` is pure over a render-stable `disagreement`. That is far weaker than the sentence implied and **would have survived a change that broke the real invariant**.

Now made true by construction: **one `const citation` per target**, read by the sentence and by every button's claim. The claim is true for the reason it gives.

### ⚠ F1, fixed — and why ten green tests could not see it

The sentence sat **inside `row.responses.map()`** while its key and data are per **target**, so two answerers rendered it **twice under a duplicate testid** — and `getByTestId` *throws* on a duplicate, making it a hard failure, not a cosmetic one.

It survived because **every case in the corpus had exactly one response, while the disagreement fixture beside it declared `answering_participants: 2`**. A reveal exists *because* people disagree: one answerer is the **degenerate** case, and the corpus was built entirely out of it (trap 22 — the class the author did not imagine). The fixture now takes **two answerers by default**.

Hoisting also separated the sentence from the `r.value !== null` gate beside each button, so `canApply` restores it: a target where everyone declined renders no buttons, and a promise about a change nobody can make is the same over-claiming this slice exists to end.

## ⚠ Corrected premise — the guard written after this exact defect was GREEN

`factorValueEditWireCarriage.spec.ts` exists because `applied_from` once shipped dark, and it derives its expected field set from the published Zod schema. At the bumped pin **all 14 tests passed while the adapter dropped `evidence_event_id`.**

The walk is **depth-1**: it sees the top-level `applied_from` key, which the adapter does carry, and is blind to a member added *inside* it. `applied_from` is picked rather than spread (deliberately — a spread would 422 the turn), and **a pick is a mirror by construction**. The walk now goes one level deeper with its own positive control, and REDs at pristine naming the dropped member.

## Fail closed on a malformed citation — for a *different* reason than the ids beside it

Dropping a `round_id` produces a **lie** (a colleague's number stamped "Set by you"). Dropping a citation produces an **ambiguity**: CEE's stamp, its log line and the contract all rest on *absent ⇔ cited nothing*, and a client that may silently drop one makes that biconditional false everywhere downstream, unrecoverably. A refusal is visible and retryable.

## Threaded through five picks

`RevealBody.onApply` → `PanelSetupPage.handleApply` → `panelApplyHandoff` (localStorage) → `usePanelApplyDrain` → `buildFactorValueEditEvent` → `adaptFactorValueEdit`. Each was a mirror of the contract shape. The stored intent carries **an id only** — the evidence body is a person's words and its author has a name, and the PII rule keeps both out of localStorage.

## ⚠ DEPLOY ORDER — CEE #966 FIRST, THIS AFTER

`.strict()` again: a CEE below 0.41.0 refuses the whole turn rather than dropping the field. Every intermediate state is valid — the uncited path is byte-identical at every hop, and a null/absent `standing_note` or disagreement view degrades to the pre-0.41.0 behaviour rather than to a broken button.

## Gates
`pnpm typecheck` PASSED — baseline **unchanged 2485→2485**, coverage 3445/3485 · 167 files / 2287 passed across collab + the apply chain · new `panelApplyCitation.spec.tsx` **10 asserted by name**, carriage **22** · eslint clean on every touched file.

**Mutants — 6 applied, 6 bitten, 0 survivors.** Isolation proven by **writing** a sentinel (trap 9g), worktree outside the repo root and removed before the gates, applied-check asserting exactly one changed file, opening control 32 = trailing control 32.

The load-bearing evidence is the **discriminating pair**: relaxing *exactly-one* to *at-least-one* REDs **only** the two-evidence test while the one-evidence test stays green; deleting the claim at the click site REDs **only** the one-evidence test. Neither alone binds the rule — the pair proves it holds in both directions.

## ⚠ ROLLBACK IS NOT GRACEFUL — reverting CEE alone breaks owners

If CEE #966 is reverted while this stays deployed, an owner applying on a **single-evidence** target sends `evidence_event_id` to a CEE that refuses unknown members, and **the whole turn 422s**. Targets with **0 or 2+** evidence rows keep working, because they send no citation. So the failure is **partial and confusing**, not clean: the same button works on some rows and fails on others.

**Roll back this PR first, or roll back both together.** Never CEE alone.

**Do not merge — reviewed, F1 fixed; the coordinator runs the merge/deploy order.**


Talchain:
Deployed exact-head witness verdict: **FAIL; train rollback in progress**.

Exact merge/deploy: `26b2e68f8dd6faeb05fdebb11ea0026ce9b35447` over prior-good UI `1de0a0b7f24723786c43f78804b6bca66b8ab115`; paired CEE `904d57d3595d3d4403ca6a7354cc211a021ef915`.

The real browser journey passed: owner-created two-person round; blind Ada/Grace packets; Ada note; exact 0.35/0.85 elicitation; owner reveal preserved dissent; one-note disclosure stated Ada’s note would be recorded; clicking Use Grace’s 0.85 showed the bounded “being applied” copy. It then failed the load-bearing handoff: after clicking Back to your model and waiting three minutes, **no `factor_value_edit` request reached `/proxy/v5/turn`**, no citation reached the wire, and the persisted graph stayed at 0.4 / cee_inference. No fixture-built payload was used.

Evidence: `acceptance-evidence/codex-night-shift-2026-08-15/train4b-p2-ui-exact-26b2e68f/VERDICT.json`. UI rollback PR #708 is first; paired CEE rollback PR #969 is held until the old UI is serving. Please retain this PR/branch for inspection; do not layer a compensating train fix.

PR #731, opened 2026-08-16
Adopt redacted model-building notices reader

Talchain:
## What

- Vendors the exact `@talchain/schemas` `v0.45.0` package and adopts only its optional top-level `model_building_notices` field.
- Carries validated aggregate notices ephemerally on live assistant messages.
- Renders one compact, accessible, neutral `Model-building notices` strip after assistant text and before structured blocks.
- Adds strict parser, raw-wire-to-DOM, accessibility/copy, legacy/absent-field, malformed/mutant, and transcript-serializer-negative coverage.

## Reader-first safety boundary

- This is a reader-only change. The CEE producer must remain gated until this reader is merged and deployed.
- An absent field produces no notice DOM and preserves legacy behavior.
- Any malformed, unknown, unredacted, detail-bearing, duplicate, or mismatched notice data fails the whole response parse; no partial notice is presented.
- Presentation uses fixed neutral copy and grouped counts with `details_redacted` only. It never renders labels, values, raw reasons, node IDs, “you said” language, conclusions, or new actions/prompts.
- Notice data is not written to `applyV5State`, canvas/Zustand, graph, analysis readiness, receipts, compute, context, or local/Supabase transcript persistence.
- No CEE/provider/PMS/model/cache/config or producer behavior changes are included.

## Exact identities and provenance

- Target/source base: `9a8b84c6f95b7530e54480cc04950fae73c3a142`
- Re-derived live staging Netlify commit before publish: `9a8b84c6f95b7530e54480cc04950fae73c3a142`
- Head: `b719e54d63bacb82f0570c40532c8ed85571af4d`
- Tree: `6cad8fe28a433ee1773e53eca4f6e245ab33e796`
- Schemas tag/main commit: `3c6545f9115d0a392c6210b858683906232e3e7d`
- Schemas tree: `02d3d6537cbcf40646a4a40ae061c09151ff2ebd`
- Vendored tarball SHA-256: `274fc8bd28962ffee2bc8acc19da5978ca1477ab778a397c3aa069421d1843cb`

## Validation

- Focused matrix: 8 files / 62 tests passed.
- Post-fix notice DOM suite: 5 / 5 passed.
- Vendor integrity and schema-version guards passed (`0.45.0`).
- Source typecheck coverage/ratchet gate passed within the repository's exact baseline.
- Changed-file lint: 0 errors; 12 inherited warnings on pre-existing lines in `MessageBubble.tsx` / `useConversation.ts`.
- `git diff --check` passed.
[...]

PR #752, opened 2026-08-17
feat(2.1271): deliver the auto-run's provisional analysis without another turn (H4-guarded)

Talchain:
**⚠ DO NOT MERGE YET.** Paired CEE PR: `Talchain/olumi-assistants-service` **#1010** (the producer half, itself stacked on #1004). This half is cut from `staging` and is honest on its own — with an older CEE it simply never arms.

---

## What was broken

Paul's ruling, 17 Aug: *"Treat server calculation + automatic client delivery/rendering as one capability. Do not report auto-analysis as COMPLETE until it appears to the user without another turn."*

CEE's post-draft auto-run commits a `run_analysis` fact ~20s after the draft SSE stream's terminal COMPLETE frame closes the socket. No CEE route returned a scenario's analysis except a turn. So the provisional result was computed, persisted, and seen only by a user who happened to send another message.

## Three pieces

**1 · `adapters/cee/scenarioGraph.ts`** carries CEE's two new keys through:
- `analysisState` — parsed with `AnalysisStateV1Schema`, the CONTRACT's own `.strict()` discriminated union and the same schema object `applyV5State` uses. Never a local mirror of the vocabulary. A malformed verdict yields `null` = *"CEE did not answer"*.
- `analysisResult` — opaque, gated on its own `type` discriminator, handed to the one mapper that owns the block contract.

Both **REQUIRED** on the `graph` member so a consumer cannot silently forget them and `undefined` is unreachable.

**2 · `canvas/hydrate/applyScenarioAnalysisRead.ts` — THE H4 GUARD.** This is the file the PR exists for.

`applyV5State` is the TURN applier and its clear-on-absence rule is **correct for turns** (*"the field's whole contract is 'CEE stated this FOR THIS TURN', so silence must clear"*). It is wrong for a read, because the two authorities answer different questions (trap 21):

| | question | can say `running`? |
|---|---|---|
| draft turn | *did I start a provisional run?* | yes |
| graph read | *has a fact landed for this graph?* | **no** — CEE keeps no in-flight marker |

So a read taken mid-run can only answer `never_run`, and through the turn applier that flips the product from *"an analysis is running"* to *"no analysis has ever been run"* **while one is running** — with `never_run` licensing the pre-analysis affordance over it. Not an edge case: the **default behaviour on the first read**.

This applier writes a read's verdict **only on a TERMINAL kind** (`complete_current`, `complete_stale`, `blocked`, `refused` — each derived from the contract's own `.describe()` text, one line of reasoning per kind in the header) and performs **NO STORE WRITE AT ALL** for `never_run`, `unknown_degraded`, `running`, or an absent verdict.

Two further deliberate choices:
- **Results are written BEFORE the verdict.** A verdict-first order exposes a frame where `complete_current` vouches for a result not yet on screen, which `deriveAnalysisDisplayState` renders as *"Ready to analyse"* over a completed analysis. Pinned by a call-order assertion.
- **It never touches the graph.** `serverGraphHydration` stays the one ingestion authority (`draftRecovery.ts:13-19`: *"A second ingestion path here would be the two-`generateGraphHash`-twins defect all over again"*).

**3 · `canvas/hooks/useProvisionalAnalysisDelivery.ts` — the bounded wait.** Armed by the draft turn's own `running` verdict, re-arming on `started_at` IDENTITY (a second draft gets its own wait; a re-render does not restart one), and **inert on every draft where CEE scheduled nothing** — it does not poll on a timer and does not second-guess the server. 7 reads over 60s: the run is ~20s and the route's limiter is **per CLIENT IP**, shared with boot hydration and draft recovery (`read` tier = 90 rpm, derived at CEE's tip).

**H3 — on expiry it invents nothing.** A `running` claim can outlive its run (the dispatch throws, the already-analysed guard skips, a process restarts). The wait ends and **writes nothing** — no synthesised `unknown_degraded`, no verdict of the client's own. And the manual Run affordance is untouched throughout, because every gate that could disable it reads the LOCAL results status, never the wire.

Mounted in `routes/CanvasMVP.tsx` beside `useServerGraphHydration` — deliberately **not** in `useConversation` (owned by #731, and the trigger is a store verdict, so the delivery needs no conversation context and stays decoupled from the message loop).

## Verification

**G3 IS ASSERTED, NOT DESCRIBED.** The applier spec drives a read-shaped `never_run` through the **real `applyV5State`** and shows the demotion happening, next to the narrow applier declining it in the same test — so the necessity of this file cannot quietly stop being true.

**Mutants: 11 / 11 BITTEN**, in a worktree outside the repo root with **sentinel-WRITE isolation** (a sentinel written into the worktree, source asserted unchanged; inodes compared — `cp -R` can preserve APFS hard links), restores from a **pristine archive outside the tree**, applied-check **exactly 1** / control **exactly 0** on every mutant, pristine AND trailing controls green, and the collected count asserted **28 BY NAME** (both spec files named, never an aggregate).

| mutant | verdict |
|---|---|
| M1 the H4 filter admits EVERY read verdict | **BITTEN** (12 tests red) |
| M2 INVERT the filter — admit only NON-terminal kinds | **BITTEN** (22) |
| M3 admit `never_run` to the terminal set *(the brief's named H4 mutant)* | **BITTEN** (9) |
| M4 verdict before results | **BITTEN** (2) |
| M5 treat an absent verdict as a state (write `null`) | **BITTEN** (2) |
| M6 drop the results dedupe | **BITTEN** (1) |
| M7 settle the schedule on a non-terminal read | **BITTEN** (6) |
| M8 make the wait unbounded | **BITTEN** (1) |
| M9 conclude on an unreadable (404) read | **BITTEN** (1) |
| M10 snapshot the store once instead of per attempt | **BITTEN** (1) |
| M11 keep reading after delivery (budget leak) | **BITTEN** (1) |

M1/M2 are the discriminating **pair** for the guard itself, failing on different assertions.

**P4 · the mutated tree was typechecked.** The instrument is `tsc -p tsconfig.mutant-scope.json`, extending the repo's own `tsconfig.app.json` (inheriting `types: ["vite/client"]`, `strict`, `noUnusedLocals`, the path aliases) and narrowed to the three mutated modules; it counts diagnostics **in those files only** against a pristine baseline of 0, cross-validated two ways — a full `tsc -p tsconfig.app.json --noEmit` (1981 diagnostics total, **zero** in these files) and the scoped run (14, all in pre-existing transitive files, **zero** in these). 17s vs 7min, so the cheap instrument is used per the P4 cost refinement.

**Two instrument defects the controls caught, disclosed:**
1. A first ad-hoc `tsc` flag soup reported **21 phantom errors on the CLEAN tree** (missing `vite/client`, so `import.meta.env` failed) — the mandatory pristine control refused to proceed. Replaced with the project-config instrument above.
2. M1's first form (`return true`) was **type-INVALID** under `noUnusedParameters` and correctly scored VOID, not SURVIVED — the exact false-survivor class P4 exists for. Reformed as `return kind.length >= 0`.
3. The G3 test's cold dynamic import of `applyV5State` exceeded vitest's 5s default in a fresh worktree and turned the pristine control red. Fixed **at the source** with an explicit budget (separate commit), never absorbed into a baseline.

**Typecheck gate PASSES** — coverage 3452/3492 (both new files loaded, no new exception entries), ratchet within baseline. One existing fixture fixed **at the source**: `useConversation.draftRecovery.spec.ts`'s `serverGraphResult()` gains the two keys as `null`. The gate's reported 18-diagnostic pre-existing shrink is deliberately **NOT** banked here — it is churn in files this PR never touched.

## Two findings for the reviewer, from a deployed-consumer audit (P2)

A read-only audit traced every mounted consumer of `useAnalysisState()` for exactly this input (wire `running`, no local run, `hasReport === false`). Both are **pre-existing** and neither is introduced here, but this PR makes them reachable:

1. **The `running` signal is quieter than the brief assumed.** `semantic` and `displayedFreshness` both stay `'none'`, so there is **no** false "Analysis complete", no "Results may be outdated", no Rerun CTA — the P2 harm is avoided. But `AnalysisStateRegion` (the **Analysis tab**, which a fresh draft fronts) is driven by `useAnalysisRunState()`, which reads only store slices and is **blind to the wire** — so on the default tab the user sees nothing change. The visible in-flight chrome lives on the Model and Compare dock tabs. **#749 is reworking `useAnalysisRunState` into a union and is the right home for closing that**; it is not in scope here and no new display surface is added.
2. **`AnalysisRunningBanner` differences CEE's `started_at` against the browser clock with no past-direction skew guard** (`:94-96`), so a `started_at` >20s behind the client clock opens the banner on *"Still analysing…"*. Bounded by clock offset rather than by elapsed time in this flow, but worth a row.

Also noted: the Compare tab's honest empty state is replaced by a loading skeleton for the run's duration (`CompareTabBody.tsx:229-238`), which is that component's deliberate existing behaviour.

### DIFF
--- a/src/adapters/cee/scenarioGraph.ts
+++ b/src/adapters/cee/scenarioGraph.ts
@@ -56,6 +56,9 @@
  * locally; see `canvas/utils/mergeServerGraph.ts`.
  */
 
+import { AnalysisStateV1Schema } from '@talchain/schemas/boundary'
+import type { AnalysisStateV1 } from '@talchain/schemas/boundary'
+
 import { logger } from '../../lib/logger'
 import { parseNotModelled, type NotModelledManifest } from './notModelled'
 
@@ -118,6 +121,34 @@ export type ScenarioGraphResult =
       identity: ScenarioGraphIdentity | null
       /** MEASURED by CEE on the returned bytes. `false` for every real graph today. */
       layoutPresent: boolean
+      /**
+       * ROADMAP 2.1271 — CEE's composed `AnalysisStateV1` verdict for this
+       * scenario, PARSED, or `null`.
+       *
+       * ⚠ `null` IS NOT A STATE, and reading it as one is the whole hazard of
+       * this field. It means CEE DID NOT ANSWER — the build predates the key, the
+       * scenario has no graph, or the shape failed validation. A consumer must
+       * leave whatever it already believed standing. In particular it must NOT be
+       * read as evidence against an in-flight run the DRAFT TURN reported: the
+       * two authorities answer different questions (a turn answers "did I start a
+       * run?", a read answers "has a fact landed?"), and CEE keeps no in-flight
+       * marker, so mid-run this leg can only ever say `never_run`. See
+       * `canvas/hydrate/applyScenarioAnalysisRead.ts`, which is the ONLY
+       * sanctioned consumer.
+       */
+      analysisState: AnalysisStateV1 | null
+      /**
+       * ROADMAP 2.1271 — the `analysis_result` block for the fact the verdict
+       * selected, present ONLY on a `complete_current` verdict (CEE withholds it
+       * on a stale one, because those numbers describe a graph the user has since
+       * changed). `null` means no CURRENT result is being delivered — never "the
+       * analysis is empty".
+       *
+       * Deliberately typed `unknown`: the block is handed to `mapV5AnalysisToReport`
+       * — the SAME mapper the turn path uses — and a second local shape
+       * declaration here would be a mirror of the block contract.
+       */
+      analysisResult: unknown
       requestId: string | null
     }
   /** 200, `graph_present:false` — the scenario exists and has no graph yet. Normal. */
@@ -174,6 +205,48 @@ function readIdentityEnvelope(raw: unknown): ScenarioGraphIdentity | null {
   return { value, projectionVersion }
 }
 
+/**
+ * ROADMAP 2.1271 — parse CEE's verdict with the CONTRACT, never a local mirror.
+ *
+ * `AnalysisStateV1Schema` is `.strict()` at every level and its `run_state` is a
+ * discriminated union, so an unknown kind or an extra key FAILS rather than
+ * handing a consumer a shape it will read as authority — the same discipline
+ * `applyV5State` applies on the turn path, using the same schema object.
+ *
+ * A parse failure returns `null`, i.e. "CEE did not answer". That is deliberately
+ * the same value as absence here, and it is safe ONLY because the single consumer
+ * treats `null` as "leave what you believed standing" rather than as a state.
+ */
+function readAnalysisState(raw: unknown): AnalysisStateV1 | null {
+  if (raw === null || raw === undefined) return null
+  const parsed = AnalysisStateV1Schema.safeParse(raw)
+  if (parsed.success) return parsed.data
+  logger.warn('scenario_graph.analysis_state_invalid_shape', {
+    issueCount: parsed.error.issues.length,
+  })
+  return null
+}
+
+/**
+ * The `analysis_result` block, gated on its own discriminator.
+ *
+ * Not validated further on purpose: `mapV5AnalysisToReport` is the block
+ * contract's one consumer and restating its shape here would be a mirror of it
+ * (trap 12). But the TYPE TAG is checked, so a future CEE key landing on this
+ * name cannot be forwarded to a mapper written for a different block.
+ */
+function readAnalysisResultBlock(raw: unknown): unknown {
+  if (raw === null || raw === undefined || typeof raw !== 'object') return null
+  const type = (raw as { type?: unknown }).type
+  if (type !== 'analysis_result') {
+    logger.warn('scenario_graph.analysis_result_unexpected_type', {
+      receivedType: typeof type === 'string' ? type : typeof type,
+    })
+    return null
+  }
+  return raw
+}
+
 function parseOk(body: unknown): ScenarioGraphResult {
   if (body === null || typeof body !== 'object') return { status: 'unusable' }
   const b = body as Record<string, unknown>
@@ -215,6 +288,17 @@ function parseOk(body: unknown): ScenarioGraphResult {
     notModelled: parseNotModelled(b.not_modelled),
     identity: readIdentityEnvelope(b.graph_identity_hash),
     layoutPresent: b.layout_present === true,
+    // ROADMAP 2.1271 — PARSED, NOT TRUSTED, and by the SAME `.strict()`
+    // discriminated-union schema `v5/applyV5State.ts` uses on the turn path. A
+    // malformed verdict yields `null` ("CEE did not answer") rather than a shape
+    // a consumer would read as authority. There is deliberately no local mirror
+    // of the vocabulary here.
+    analysisState: readAnalysisState(b.analysis_state),
+    // Handed through opaque: the ONLY reader is `mapV5AnalysisToReport`, which
+    // already owns the block contract. Presence is gated on the discriminator
+    // being the one block type this leg may carry, so a future CEE key cannot
+    // arrive here as an unlabelled object.
+    analysisResult: readAnalysisResultBlock(b.analysis_result),
     requestId,
   }
 }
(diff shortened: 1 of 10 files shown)


Talchain:
## MERGE — both blocking changes cleared, and the open question is answered with a precise boundary

### F1 — the dedupe was dead in production; now reachable and bound to the real store

**Proven at runtime, not inferred:** `'currentResultsHash' in useCanvasStore.getState()` = **false**, `typeof` = `undefined`. Positive controls `setAnalysisStateV1`/`resultsComplete` = `function`; contrast control `currentResultsHashXYZ` absent. The probe discriminates.

Fixed with `readProvisionalApplyStore()` — **no cast** (so drift now REDs `tsc`) and **no spread**. ⚠ **The spread would itself have been a defect:** the mutant proved it handed the applier **266** store members, including the graph slices the applier's own header assigns to `serverGraphHydration`.

New spec bound to the **real** `useCanvasStore` — 7 collected. Under M-F1 (the shipped double-cast restored): **5 failed / 2 passed**, including `expected { outcome: 'applied' } to deeply equal { outcome: 'alreadyHeld' }` — direct proof the dedupe was dead.

**⚠ And a gap closed beyond the brief:** nothing pinned that the *delivery* used the builder — every spec injected its own `getStore`, **which is exactly how this defect shipped**. `getStore` is now optional with `readProvisionalApplyStore` as the default, so production has one path. Mutant **M-H** (helper left perfectly correct, core bypasses it) is caught **only** by the new production-path test.

### F2 — rebased and re-measured on the contract that will actually ship
Clean rebase, 0 conflicts, **85** commits behind (not 82). Pin moved **0.46.0 → 0.48.0**. All specs re-run on that pin: `applyScenarioAnalysisRead` 20 · `useProvisionalAnalysisDelivery` 8 · `useConversation.draftRecovery` 7 · new 7 / 7 / 5. None zero-collected.

### The open question — **FAIL, and the boundary is sharper than expected**

Settled in the DOM, not statically:

| state | result |
|---|---|
| `complete_stale`, no report, **no prior verdict** | notice null-renders — no misleading string |
| `refused`, no report | banner slot is `refusal`; the freshness strip is structurally unreachable |
| **`complete_stale`, no report, user already holds a freshness verdict** | **"Model changed since this analysis. Re-run to update." renders with `body_presentation: hidden`** |

So the user is told to **re-run an analysis whose results are not on screen**. **Attribution control:** without the delivery, same store, `run_state=never_run`, banner `none`, notice absent — **the delivery is the cause**. Positive and contrast controls both fire, so the two absences are real rather than blindness.

**The string is true but unanchored** — it refers to an analysis the user cannot see. That is the shape **CEE #1068** should be judged against, and it is not a fabrication.

### Three corrected premises
1. **85** commits behind, not 82.
2. **`useAnalysisRunState()` is NOT blind to the wire** — that guard is deleted; it reads `composed.runStateKind` directly (`useAnalysisRunState.ts:247`). And `copy/freshnessReasons.ts` does not exist at this tip; the copy is `FRESHNESS_COPY.stale`.
3. **The 0.48.0 `superRefine` rules ARE reachable on the read leg** — `scenarioGraph.ts:222` uses `AnalysisStateV1Schema.safeParse`. Probed: well-formed parses (control); a cross-field violation is rejected on `blocked_unusable`. **"Unreachable" was wrong; "cannot cause a bad write" is right** — rejection → `null` → `notYet/no_verdict` → keep polling → H3 writes nothing. Fails safe.

### A premise declined rather than silently complied with
**`resultsSource` was deliberately NOT changed.** The field answers *what caused the analysis*, not *how it reached the client* (`store.ts:901`; sole consumer renders "Updated from conversation", `OutputsDock.tsx:3114`). The run **was** scheduled by the draft turn. `'direct'` would be a lie **and** suppress the indicator, so results would appear unannounced; a third value needs a product decision on that copy. Reasoning pinned at the call site so it is not re-litigated.

### `READ_TERMINAL_RUN_STATE_KINDS` now fails loud on growth
A declared non-terminal twin plus a partition assertion against the contract's own exported `ANALYSIS_RUN_STATE_KINDS`. M-E (drop a classified kind, simulating contract growth) REDs it; M-F (move `refused` across the line) REDs 4 tests.

### Mutants — 8 run, all bite
Applied-check scoped to `src/`, leading **and** trailing controls exactly 0. Isolation proven by **writing** a sentinel (source sha unchanged, worktree sha changed, inodes differ), worktree outside the repo, restores via `git checkout HEAD --`. One instrument failure caught: **M-F first read as a silent survivor because zsh does not word-split an unquoted spec list** — re-run properly it bites.

### Gate at head `b42589b1`
14 checks, **`running=0`**, required **`Staging Gate` = success**. Sole failure is `Visual Regression (advisory)`, which is not in the gate's `needs`.

**Pre-existing red, recorded not chased:** `aiPanelTranche1.spec.tsx > Item 14 — SuggestedChips focus ring` fails identically at pristine staging `5113f5a0`, same counts.

⚠ **CEE #1068 must not merge until this is DEPLOYED**, and its residual list must name `complete_stale` and `refused` — now with the boundary above.

PR #773, opened 2026-08-18
fix(olumi): converge every Ask-Olumi action on one surface, and stop log language reaching the user

Talchain:
# B3 · Olumi convergence — one Olumi surface, and no log language in it

Base `staging` @ `dd089a50`. Lane B3 of the first-send UX gate. **Phase 2 (non-removal
convergence) only — no visible content block is removed; the Phase-1 KEEP/CUT/REHOME table is in
the lane report and awaits Paul's veto.**

---

## P-COMPLIANCE BLOCK — read **v4, P1–P9**

| Rule | Evidence |
|---|---|
| **P7** producer semantics from the PRODUCER | `humaniseWireToken` deliberately has **no lookup table** — a token→word map would be a hand-maintained mirror of a producer enum and every entry a guess at semantics I did not derive. It changes casing and separators only, so it cannot assert a meaning the producer did not have. The one *semantic* judgement — "an entity id is never display text" — is derived from `friendlyOperation.RAW_ID_PATTERN`, whose header states it is kept in sync with the CEE producer's own `patch-summary.ts:ENTITY_ID_LEAK_RE`. |
| **P2** checked through the MOUNTED consumer | Mount path derived at this tip: `AppPoC.tsx:951/956` → `CanvasMVP.tsx:245` → `ReactFlowGraph.tsx:2105` → **`OutputsDock.tsx:3201 → OlumiTabBody → ConversationPanel`** (docked, always mounted) and **`ReactFlowGraph.tsx:2451 → FloatingOlumiPanel → ConversationPanel`**. Gate is `aiPanelV2`: `defaultValue: true` (`src/flags.ts:357`) **and** `VITE_FEATURE_AI_PANEL_V2="true"` (`netlify.toml:57`) → ON in every deployed context. Every render assertion executes the real component through that path; `DraftChat` is the flag-OFF rollback host and was deliberately **not** used as the consumer under test. |
| **P1** one seam BEYOND the guard | The guard is `humaniseWireToken`. The seam past it is **the render site's null-handling**: a caller that wrote `?? raw` would reinstate the leak with the guard fully intact. So every fix asserts at the DOM — `card.innerHTML` as well as `textContent`, because `TargetRefPill` also put the raw id in an `aria-label`, i.e. past the visible-text seam and into the accessibility tree. |
| **P5** claims grounded in canonical persisted state | n/a — nothing here asserts anything about the user's model, session or history. The one claim added ("This build cannot apply that kind of change yet") is about **this build's own capability**, which is the code path making it. |
| **P6** no manufactured obligations | n/a — no readiness/repair semantics touched; nothing new is required of the user. |
| **P8** every affordance has a pinned acceptance path | The only affordance changed is the guidance strip's action button. Its face was previously **empty** for an unmodelled wire type — an affordance with no readable answer at all. It now always carries a verb, and `GuidanceStrip.spec.tsx` pins that the label is non-empty and never the raw token. |
| **P4** mutation harness typechecked the mutated tree | **Mandatory control first:** the pristine tree passes `pnpm run typecheck` (563 files / 2306 errors). Three mutants of the type-risky shape (two `if (false)` orphan candidates + one widened assignment) were applied **together** and the gate still read **2306** — i.e. type-VALID, so no verdict is a false survivor. Restore control after: exactly **0** dirty files. Note the risk class P4 exists for (a false SURVIVOR) is not reachable here: **10/10 mutants were BITTEN.** |

---

## 1 · Every Ask-Olumi action overrode the surface the user had chosen

`revealOlumiSurface()` is **THE** convergence primitive — six direct call sites plus
`withOlumiReveal`, which wraps every guidance-store send, so Graph, Analysis and Inspector all
arrive here. It did two things unconditionally:

```ts
forceActivateOutputTab('olumi')   // claim the DOCK
focusFloating()                   // focus the FLOATING panel
```

Those fight each other, and the fight is visible. Claiming the docked Olumi tab makes
`dockHostsOlumi` true, and `FloatingOlumiPanel` returns `null` the moment it does
(`FloatingOlumiPanel.tsx:856`, `yieldToDockedOlumi`). So:

1. **A user working in the floating Olumi window who clicked "Ask Olumi" on a node had that window
   closed** and the conversation relocated to the dock. The product overrode their surface choice —
   the exact opposite of converging on one interaction model.
2. `focusFloating()` then focused the panel that was about to unmount, while the surface that
   actually took over — the docked composer in `PersistentInputStrip` — got **no focus at all**,
   because nothing had ever registered a channel for it. The action "worked" and the user still had
   to click into the box before typing.

**Now:** reveal the surface the user already has; claim the dock only when no Olumi composer is on
screen.

⭐ The test for "is a floating/hero composer on screen?" is **`focusFloating()`'s own return
value**. Both floating surfaces register their focus channel under exactly the condition that makes
them paint (`FloatingOlumiPanel.tsx:553`, `FirstUseComposer.tsx:283`) and deregister on cleanup, so
**a registration is the surface's own statement that it is visible** — taken from the surface, not
modelled here.

⚠ **A first draft of this did model it**, deriving visibility from the node count, the persisted
dock-open state and the active tab. Two things were wrong with it, and both are recorded because
they are the interesting part: it was a **second copy of a rule the surfaces already implement**
(trap 12 — the copies disagree the first time the rail rule moves), and it made this module import
`OutputsDock`, closing a real cycle **`revealOlumi → OutputsDock → guidanceStore → revealOlumi`**.
The repo's own typecheck ratchet caught it as **+1 error in a file I had never touched**
(`src/lib/auth/accessValidation.ts` TS7006) — proven pre-existing-clean by a pristine control
worktree. Reverted; the shipped version imports nothing new but the docked channel.

`dockedOlumiFocus.ts` is a **separately named** second channel rather than a widening of
`useFloatingFocus`: the two surfaces answer different questions and folding them into one registry
means the last surface to mount silently wins (trap 21). `PersistentInputStrip` registers it **only
in composer mode**, so the channel exists exactly when there is a textarea to focus — a channel that
reports success while focusing a null ref is worse than none, because the caller stops looking.

## 2 · Raw wire tokens rendered as user copy

Five surfaces put a producer token straight into the DOM, every one of them at the moment the
product is telling the user something went wrong:

| Was | Now |
|---|---|
| `VALIDATION_FAILED: Patch validation failed` | the producer's sentence; the code moves to `data-rejection-code` |
| `Unsupported operation: add_node` | "This build cannot apply that kind of change yet (add node)." |
| proposal pill `add_edge` | `Add edge` |
| proposal pill `fac_x7` — **visible label AND `aria-label`, i.e. read aloud** | omitted; the change's own sentence carries the meaning |
| unsurfaced-block pill `v5_flip_analysis` | omitted; `data-block-type` + the DEV warn keep it for operators |
| `[item]` mid-sentence (a scrub that read as a broken template) | `this element` |

Plus a runtime hole: `GuidanceStrip`'s `actionLabel` switch had **no `default`**, so an unmodelled
wire `primary_action.type` produced the aria-label `"undefined: <title>"` and an **empty button
face**. TypeScript could not see it — the union made the switch exhaustive at compile time, which is
exactly why it survived review. `primary_action.type` arrives on the wire, so the union is a claim
about pin time, not a runtime guarantee.

**Operator diagnosis is preserved, not deleted.** Removing a token from the user's eyes must not
remove it from an operator's reach, or the next lane re-adds the leak to get its diagnosis back —
so both `data-rejection-code` and `data-block-type` are pinned by test.

---

## Evidence

**RED-first at pristine, named signatures, collected counts asserted BY NAME:**

| Spec | RED at pristine | GREEN after |
|---|---|---|
| `wireTokenLeak.spec.tsx` | 6 failed / 6 passed (12) | **12 passed (12)** |
| `revealOlumiSurface.spec.ts` | 4 failed / 3 passed (7) | **9 passed (9)** |
| `PersistentInputStrip.spec.tsx` | — (4 new cases) | **12 passed** (8 pre-existing + 4) |
| `GuidanceStrip.spec.tsx` | — (1 new case) | **23 passed** (22 pre-existing + 1) |

**Mutants — 10/10 BITTEN, 0 void, 0 control failures.** Throwaway worktree OUTSIDE the repo root,
detached at the PR head; **sentinel-WRITE isolation proven with its own positive control** (an
earlier isolation probe of mine reported OK against a worktree that did not exist — trap 13 caught
on myself, hence the control); inodes differ; restores from a **pristine archive**, never from the
other copy; applied-check scoped to `src/` = exactly 1 per mutant, control after restore = exactly 0.

```
M1  unconditional dock claim (the original defect)      BITTEN  2 failed | 7 passed (9)
M2  dock path focuses the floating channel              BITTEN  1 failed | 8 passed (9)
M3  docked channel returns true without focusing        BITTEN  1 failed | 8 passed (9)
M4  strip registers in every mode                       BITTEN  3 failed | 9 passed (12)
M5  rejection-code leak restored                        BITTEN  2 failed | 10 passed (12)
M6  humaniser admits entity ids                         BITTEN  1 failed | 11 passed (12)
M7  proposal pill renders the raw target                BITTEN  1 failed | 11 passed (12)
M8  v5 blockType pill restored                          BITTEN  1 failed | 11 passed (12)
M9  actionLabel loses its default arm                   BITTEN  1 failed | 22 passed (23)
M10 proposal pill renders the raw operation             BITTEN  1 failed | 11 passed (12)
```

**Gate, in the required check's step order** (`Staging Gate` → `TypeScript + Lint` job: lint → typecheck → tests):
- **Lint** — 0 errors on every touched path (pre-existing warnings only).
- **Typecheck** — `pnpm run typecheck`: **563 files / 2306 errors, byte-identical to a pristine
  control worktree at `dd089a50`**.
- **Tests** — the four affected specs are green and mutation-proven above; the wider
  `src/canvas/conversation` sweep was still running when this lane stopped (per the 18 Aug amendment,
  lanes do not poll). **CI is the authority on the full suite.**

Not re-blessed: any visual-regression reference (N-30).

### DIFF
--- a/src/canvas/components/PersistentInputStrip.tsx
+++ b/src/canvas/components/PersistentInputStrip.tsx
@@ -1,9 +1,10 @@
-import { memo, useCallback } from 'react'
+import { memo, useCallback, useEffect, useRef } from 'react'
 import { ChevronUp } from 'lucide-react'
 import { typo } from '../../styles/typography'
 import { useFloatingPanelState } from '../hooks/useFloatingPanelState'
 import { useStageAwarePlaceholder } from '../hooks/useStageAwarePlaceholder'
-import { AIInputBar } from './AIInputBar'
+import { registerDockedOlumiFocus } from '../conversation/dockedOlumiFocus'
+import { AIInputBar, type AIInputBarHandle } from './AIInputBar'
 
 interface PersistentInputStripProps {
   /** True when the docked Olumi tab is the active dock tab. When true and
@@ -56,6 +57,30 @@ export const PersistentInputStrip = memo(function PersistentInputStrip({
   // nowhere else to type.
   const floatingIsMinimised = useFloatingPanelState((s) => s.isMinimised)
   const placeholder = useStageAwarePlaceholder()
+  const inputBarRef = useRef<AIInputBarHandle | null>(null)
+
+  /**
+   * THE DOCKED OLUMI COMPOSER'S FOCUS CHANNEL.
+   *
+   * `revealOlumiSurface()` — the primitive every Ask-Olumi / chip / coaching
+   * action funnels through — could previously only focus the FLOATING panel,
+   * because that was the only surface with a registered channel. On the
+   * dominant deployed path the dock hosts Olumi and the floating panel has
+   * deliberately deregistered, so a revealed action left the composer
+   * unfocused and the user had to click into the box before typing.
+   *
+   * Registered ONLY in composer mode — the same condition the render below
+   * uses, so the channel exists exactly when there is a textarea to focus.
+   * In status mode the floating panel owns focus; in redirect mode there is no
+   * input at all, and registering there would hand `focusDockedOlumi()` a null
+   * ref that silently succeeds (a focus channel that reports success while
+   * focusing nothing is worse than none).
+   */
+  const isComposerMode = isOlumiTabActive && (!floatingIsOpen || floatingIsMinimised)
+  useEffect(() => {
+    if (!isComposerMode) return
+    return registerDockedOlumiFocus(() => inputBarRef.current?.focus())
+  }, [isComposerMode])
 
   const handleFocus = useCallback(() => {
     onFocusFloating?.()
@@ -117,6 +142,7 @@ export const PersistentInputStrip = memo(function PersistentInputStrip({
   return (
     <div className="border-t border-panel-border bg-panel" data-testid="persistent-strip-composer">
       <AIInputBar
+        ref={inputBarRef}
         variant="strip"
         onCogClick={onCogClick}
         onChevronClick={onOpenFloating}
--- a/src/canvas/components/__tests__/PersistentInputStrip.spec.tsx
+++ b/src/canvas/components/__tests__/PersistentInputStrip.spec.tsx
@@ -15,6 +15,7 @@ import { render, screen, fireEvent, act } from '@testing-library/react'
 import type { ReactNode } from 'react'
 import { useFloatingPanelState } from '../../hooks/useFloatingPanelState'
 import { PersistentInputStrip } from '../PersistentInputStrip'
+import { focusDockedOlumi } from '../../conversation/dockedOlumiFocus'
 import {
   ConversationProvider,
   useConversationContext,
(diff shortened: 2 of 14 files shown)


Talchain:
## ⚠ Correction + follow-up commit `bc73df59`

**Correcting the "Tests" line in the PR body.** It said the wider sweep "was still running when this lane stopped". The sweep then finished and **it was not clean** — and an earlier partial read of a *different*, still-running sweep had shown "0 failures", which I reported as such. That was an unfinished measurement described as if it were informative. The finished numbers:

| Sweep | Result |
|---|---|
| `src/canvas/conversation` + `src/canvas/components` + `src/v5` + `tests/ci-guards` | **2 failed / 8,033 passed / 50 skipped (8,085)** — 2 files of 592 |
| `src/canvas/conversation` alone (independent run, separate worktree) | **2 failed / 2,372 passed / 22 skipped (2,396)** |

Both runs failed on **exactly the same two specs and nothing else** — and both were specs that **asserted the defects this branch removes**:

- `GraphPatchBlock.spec.tsx` › *"shows rejection info inline when rejected"* required `/INVALID_NODE/` **in the rendered text** — it pinned the rejection CODE reaching the user, i.e. the `VALIDATION_FAILED: …` log-language defect.
- `ProposalBlock.spec.tsx` › *"renders description, changes, and consequences"* required `getByText('add')` — it pinned the raw wire operation token in the pill.

Neither is a **historic record** (no dated capture, no leak corpus — trap 14b checked before editing), so re-pointing them is correct rather than falsifying evidence. Each is re-pointed in **both directions** so the change cannot decay into the assertion simply being deleted:

- the humanised form must be **present** AND the raw token **absent**;
- the rejection code must be **absent from the text** AND **present on `data-rejection-code`**.

### Proof the repaired assertions are load-bearing, not swallowed

Re-ran the two relevant mutants against the **repaired** specs at the new head `bc73df59` (trap 13b — closing a hole must not swallow the discrimination while doing it):

```
control (pristine at bc73df59)                        63 passed (63)
M5b  rejection-code leak restored   vs repaired spec  BITTEN  1 failed | 62 passed (63)
M10b raw operation token restored   vs repaired spec  BITTEN  1 failed | 62 passed (63)
```

⚠ The first attempt at this returned `VOID — no test files` for both, because zsh does not word-split an unquoted `$VAR` and the whole spec list went to vitest as one argument. **The harness reported VOID rather than SURVIVED**, which is the only reason it was not read as two clean passes.

### State after `bc73df59`
- four B3 specs: **56/56**
- `ProposalBlock` + `GraphPatchBlock`: **63/63**
- typecheck: **563 files / 2306 errors** — unchanged, still identical to the pristine control
- production code **unchanged** by this commit, so the original 10/10 mutant ledger stands

Remote head verified both ways: `git ls-remote` and `gh api` both `bc73df59456dfa358800700955c1dff2bc5fdcab`.

PR #776, opened 2026-08-18
fix(guest): stop presenting Olumi's own saved example as the visitor's recovered work (W-1)

Talchain:
## The defect, reproduced before it was fixed

**W-1.** Measured live on the **deployed staging build `6524caed`** (2026-08-18), in a real browser, against a genuinely clean origin.

> **Method note, because W-1's own method note was not sufficient.** Clearing `localStorage` and reloading in the same JS tick still risks the SPA rewriting the autosave during unload. I cleared storage while sitting on **`/version.json`** — same origin, no SPA — then navigated to `/`. Nothing is running that can re-seed. That page also confirmed the served bundle is `6524caed`, the same commit this branch is based on.

A guest opens the **"Customer Data Platform Selection"** starter card on the first screen and reloads. What comes back:

| surface | what it says | truth |
|---|---|---|
| toast | *"Recovered unsaved changes from your last session."* | there was no session; the model is Olumi's, shipped in the app |
| pre-analysis panel | *"Olumi estimated 6 values from your brief."* | the visitor has written no brief |
| canvas | **nothing** — the "Saved example" disclosure is gone | it *is* a saved example |

Captured from the live page: `autosaveNodes: 19`, **`nodesCarryingStarterId: 0`**, `bannerPresent: false`, `redraftCta: false`.

## Root cause — one ordering

`applyDraftResult` persists the autosave from its own body (`if (!opts.skipAutosave)`). `applyStarter` stamped `starterId`/`starterTitle` **after** it returned. So the copy that reached `localStorage` — the copy the boot arbiter restores from — was the **unstamped** graph. The stamp existed in memory and died at the page boundary, and with it **both** honesty mechanisms `starterId` carries: the canvas disclosure (`StarterProvenanceBanner`) and the run gate (`computeCeeCannotSeeModel`). The boot arbiter then saw a plain autosave and said the only sentence it knew.

`applyStarter.spec.ts` stayed green through all of it: it asserts the **in-memory** stamp and never looks at what was written. That is the seam one past the guard.

## ⭐ The reported premise is corrected

**A genuinely first-time guest was never shown this.** Measured on the same build with storage empty: **empty canvas, no toast, no autosave written, `localStorage = ["sandbox.help.open","sandbox.mode"]`.** W-1's *"the app minted it at boot"* does not hold — its "fresh" state was seeded, which is exactly the hazard its own method note warns about.

That does not shrink the finding. The reachable path is **one click on the first screen a customer link lands on**, and everything downstream of that click was as reported.

## What changed

**1 · `starters/loadStarter.ts` — the identity survives the write.** Ingest with `skipAutosave`, stamp, then persist the stamped graph once. If that write throws, **no** record is left rather than an unstamped one — the safe direction.

**2 · `persist/recoveryNotice.ts` (new) — the flag carries *what* was restored.** The decision and the sentence used to sit ~200 lines apart in `ReactFlowGraph.tsx`, joined by a `sessionStorage` value of `'true'`; a boolean cannot carry what was restored, so the copy had nowhere to come from but a literal at the far end. Both halves now live in one module. `armRecoveryNotice()` **takes no argument** — it reads the graph the arbiter has just hydrated, so there is no array for a later edit to get wrong in a place no unit test can reach.

**3 · `pre-analysis-v3` — the panel asks whose brief it was.** Via `resolveStarterId`, the **same** predicate the disclosure and the run gate use. Not a second answer to one question: that is how this panel came to contradict the banner directly above it.

### The choice this makes, stated plainly
The bundled example is **deliberately** shown to newcomers — `StarterDecisions` says so in its own copy ("Or open a saved example — a real decision Olumi has modelled") and `StarterProvenanceBanner`'s header calls the disclosure a *"NON-NEGOTIABLE HONESTY REQUIREMENT"*. So it is kept, and made **honestly labelled again**: the disclosure returns on reload because the stamp now survives, and the toast says `"Reopened the saved example."` — true in every sub-case including a visitor who edited it for an hour. *"Nothing of yours was recovered"* would become a fresh false claim the moment they had.

### Recovery is not deleted
`RECOVERY_NOTICE_COPY.unsaved_work` is byte-identical to the sentence that shipped. The defect was never that sentence; it was that sentence being said about something else. Every case here has its **opposite-direction twin**, and mutants M4/M8 exist to prove the twins bite.

## Evidence

**RED-first at pristine** (behavioural, not module-missing):
- `starterIdentitySurvivesPersistence` — `expected 19, received 0` stamped nodes in the persisted autosave. Same number the browser measured.
- `estimatesAttribution` — 3 collected, **2 failed / 1 passed**; the one that passed is the twin ("a real user's brief still says so").

**Mutants — 9/9 bitten.** Throwaway worktree **outside** the repo root, isolation proven by **sentinel WRITE** (sentinel absent from source, SHAs diverged, inodes differ, source `git status` empty), restores from a **pristine archive** (never `git checkout --`, which stages), applied-check scoped to `src/` **= 1** on every mutant, **control = 0**, and a named 7-spec / **214-test** collected baseline asserted non-zero before any verdict.

| # | mutation | bitten by |
|---|---|---|
| M1 | drop `skipAutosave` (unstamped write lands first) | "never lets an UNSTAMPED copy reach storage" |
| M2 | drop the stamped write | persistence spec, 2 tests |
| M3 | classify everything as `unsaved_work` (**the pre-fix behaviour**) | saved-example assertions — twin stays **green** |
| M4 | classify everything as `saved_example` (**opposite direction**) | genuine-work assertions — saved-example test stays **green** |
| M5 | give the example the unsaved-work sentence | notice spec + single-writer guard |
| M6 | classify from `nodes[0]` only | "finds the stamp wherever it sits" |
| M7 | restore "from your brief" for an example (**the W-1 fabrication**) | attribution spec — twin stays **green** |
| M8 | call a real user's brief an example (**opposite direction**) | the twin — saved-example test stays **green** |
| M9 | hook stops asking the graph | mounted-panel spec |

**M3/M4 and M7/M8 fail on *different* assertions** — the discriminating pair, not one biting mutant.

**M1 and M9 both survived the first battery, and both were real findings, not noise.** M1 survived because the spec asserted the *final* persisted state, which a superseding write satisfies — so the assertion moved to **every** write, since between the two writes the stored record is an unstamped starter and a crash or a second tab in that window gets W-1 back. M9 survived because *nothing exercised the hook* — a constant `false` re-opened the whole fabrication under 212/212 green. That is the lesson `reactFlowGraph.restoreFreshnessOnBoot.spec.ts` already records, so M9 is now killed by a spec that **mounts the real panel over a real store and reads the sentence off the screen**, in both directions, with the starter stamp as the only difference between the two arms.

**Gate steps run locally, in the required check's order** (`Staging Gate` = `tsc` → `typecheck-selftest` → `vitest` → `vitest-summary` → `build`):
- `pnpm run lint` — **0 errors**; hooks ratchet **229/13, exactly baseline** (no new conditional hook)
- `pnpm run typecheck` — **PASSED**, baseline **2306 unchanged**, no `--update-baseline`
- `ci:guard:zustand` · `ci:guard:starters` · `check:vendor` · `ci:guard:schemas-version` — all **PASS**
- Three pre-existing spec files needed a real edit, not a suppression: `SignalDetectionInput` gained a **required** field, so the two `input()` helpers and the glossary scan had to state it. The glossary scan now scans **both** attributions — a scan that saw one branch would be a guard watching one door.

Per the 18 Aug amendment: **not polling CI, not merging.**

## For the reviewer

- **`workspaceShell/shellContract.ts` untouched** (shared). No file under `src/components/results/**`, the canvas edges, the inspector, or the conversation surface was modified.
- **Same defect class, left alone because sibling lanes own those files** — reporting, not fixing: `src/canvas/ui/inspector-v2/inspectorStrings.ts:184` maps `brief_extraction` → *"Generated from your brief"*, and `src/canvas/nodes/FactorNode.tsx:598` renders *"Olumi estimated this from your brief."* Both are false over a starter for the same reason, and both now have `resolveStarterId` available to fix it the same way.
- **Residual gap, disclosed:** no test pins the `armRecoveryNotice()` **call site** inside `ReactFlowGraph`'s boot effect — the component is not unit-mountable and extracting the arbiter is out of scope. Mitigated by removing the argument (nothing to mis-bind) and by a filesystem-derived guard, with a positive control, asserting that module is the **only** production writer of the key and the only carrier of either sentence.
- **Not re-blessed:** no visual-regression reference touched (N-30).

## P-COMPLIANCE BLOCK
Read `STANDING-BRIEF-PREAMBLE.md` **v4 · 9 rules (P1–P9)**, plus the 18 Aug amendment (lanes do not poll CI).

- **P7 — producer semantics from the PRODUCER.** `starterId`'s meaning read at `loadStarter.ts`'s own header (*"PROVENANCE IS NOT OPTIONAL… Every node carries `data.starterId`"*) and at `resolveStarterId`'s docstring (scan **every** node; the gate and the disclosure must not disagree). The copy's meaning read at `constants.ts` and at `StarterDecisions`'s header. No predicate inferred from an observed corpus.
- **P2 — the MOUNTED consumer.** Three ways: the deployed build driven in a browser (what it renders, quoted verbatim above); `PreAnalysisPanelV3` mounted over a real store and its sentence read off the DOM; `ReactFlowGraph` wired to the module rather than to a literal.
- **P1 — one seam BEYOND the guard.** The named guard is the in-memory stamp (`applyStarter.spec.ts`, green throughout). The seam past it is **persistence and rehydration** — what reaches `localStorage` and what the boot arbiter says about it.
- **P5 — grounded in canonical persisted state.** Every claim keys off `resolveStarterId` over the **restored graph** — the authoritative persisted read, present exactly when the starter graph is. No parallel flag.
- **P6 — no manufactured obligation.** No structure inferred, none created; only what the product is permitted to *say* changed.
- **P8 — every affordance's answer is acceptable.** No new question is asked. The example's existing affordance ("Re-draft this live") is unchanged and reachable again on reload, because the banner that offers it is back.
- **P4 — the mutation harness typechecks the mutated tree.** Clean tree asserted PASSING first (the mandatory control). Mutants written so none can orphan an import — the `noUnusedLocals` false-survivor case — and the two with genuine type risk (M2, M9) batched into one gate run per the 17 Aug cost refinement. **Result recorded in a follow-up comment; it had not returned when this was written, and a verification result is not written before it is measured.**

### DIFF
--- a/src/canvas/ReactFlowGraph.tsx
+++ b/src/canvas/ReactFlowGraph.tsx
@@ -29,6 +29,7 @@ import {
   shouldReleaseTextFocusOnCanvasPointerDown,
 } from './useKeyboardShortcuts'
 import { loadState, saveState } from './persist'
+import { armRecoveryNotice, consumeRecoveryNotice } from './persist/recoveryNotice'
 import * as scenarios from './store/scenarios'
 import type { Scenario } from './store/scenarios'
 import { isUUID } from '../services/turn-request-builder'
@@ -1644,11 +1645,18 @@ const ReactFlowGraphInner = memo(function ReactFlowGraphInner({ blueprintEventBu
 
       // Show recovery toast if we loaded from autosave
       // Note: Toast will be shown after component mounts via a separate effect
+      //
+      // ⚠ WHAT was restored decides WHAT we are allowed to say about it (W-1).
+      // This used to write a bare `'true'`, and the effect below turned every
+      // restore into "Recovered unsaved changes from your last session." —
+      // including a restore of Olumi's own bundled starter example, which is
+      // neither the visitor's work nor unsaved. `armRecoveryNotice` classifies
+      // by the graph's own starter stamp (the same predicate the canvas
+      // disclosure and the run gate use) and records the kind, not a boolean.
       if (recoveredFromAutosave) {
-        // Store flag for toast effect to pick up
-        try {
-          sessionStorage.setItem('olumi-recovered-from-autosave', 'true')
-        } catch {}
+        // No argument by design — it reads the graph this effect has just
+        // hydrated, so there is no array for a later edit to get wrong.
+        armRecoveryNotice()
       }
 
       // Restore ceeAnalysisReady with source-aligned fallback and validation
@@ -1771,16 +1779,17 @@ const ReactFlowGraphInner = memo(function ReactFlowGraphInner({ blueprintEventBu
       currentGraphHash: uiGraphHashSeedless(st.nodes, st.edges),
     })
 
-    try {
-      const recovered = sessionStorage.getItem('olumi-recovered-from-autosave')
-      if (recovered === 'true') {
-        sessionStorage.removeItem('olumi-recovered-from-autosave')
-        // Slight delay to ensure component is fully mounted
-        setTimeout(() => {
-          showToast('Recovered unsaved changes from your last session.', 'info')
-        }, 500)
-      }
-    } catch {}
+    // The sentence comes from `recoveryNotice`, which decided it from what was
+    // actually restored — never from a literal here. See its header for the
+    // measured defect (W-1: a bundled saved example announced as the user's
+    // own unsaved work).
+    const recoveryMessage = consumeRecoveryNotice()
+    if (recoveryMessage) {
+      // Slight delay to ensure component is fully mounted
+      setTimeout(() => {
+        showToast(recoveryMessage, 'info')
+      }, 500)
+    }
 
     // Uninstall the guidance persistence context when this canvas unmounts, so
     // a later mount cannot write under a stale decision identity.
--- a/src/canvas/components/pre-analysis-v3/__tests__/savedExampleAttribution.spec.tsx
+++ b/src/canvas/components/pre-analysis-v3/__tests__/savedExampleAttribution.spec.tsx
@@ -1,0 +1,145 @@
+/**
+ * THE MOUNTED PANEL MUST NOT ATTRIBUTE A BUNDLED EXAMPLE TO THE USER'S BRIEF.
+ *
+ * The pure-copy pin lives in `signals/__tests__/estimatesAttribution.spec.ts`.
+ * This file exists because that pin is not evidence that the PRODUCT asks the
+ * question: a mutation making `usePreAnalysisModel` compute `isSavedExample`
+ * as a constant `false` — i.e. the whole W-1 fabrication, restored — survived
+ * the copy spec, the registry spec and the guard, 212/212 green. The same
+ * lesson `reactFlowGraph.restoreFreshnessOnBoot.spec.ts` records: a perfect
+ * unit kit is not evidence that anything calls the unit.
+ *
+ * So this renders the real panel over a real store and reads the sentence off
+ * the screen, in both directions.
+ */
+
+import { describe, it, expect, beforeEach, beforeAll, vi } from 'vitest'
+import '@testing-library/jest-dom/vitest'
+import { render, screen, fireEvent } from '@testing-library/react'
+import type { Node } from '@xyflow/react'
+import { PreAnalysisPanelV3 } from '../PreAnalysisPanelV3'
+import { ToastProvider } from '../../../ToastContext'
+import { useCanvasStore } from '../../../store'
+import { useReadinessStore } from '../../../stores/readinessStore'
+import { useGuidanceStore } from '../../../stores/guidanceStore'
+import { useSignalSessionStore } from '../signals/signalSessionStore'
+import type { PreAnalysisSensitivity } from '../../../../adapters/cee/types'
+
+const SENSITIVITY: PreAnalysisSensitivity = {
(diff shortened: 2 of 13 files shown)


Talchain:
**P4 — mutation harness typecheck, now measured (it had not returned when the description was written).**

M2 and M9 applied together (different files, the two carrying genuine `noUnusedLocals` orphan risk), batched into one gate run per the 17 Aug cost refinement:

```
applied files: 2   staged: []
── phase 1: coverage  3514 / 3554 tracked TypeScript files loaded (40 declared out of scope)
── phase 2: ratchet   within baseline — 563 file(s) / 2306 error(s) (baseline: 2306)
✅ Typecheck gate PASSED (coverage + ratchet).
```

The mutated tree typechecks clean at the same baseline as the clean tree, so **neither mutant was type-invalid** — their RED verdicts are behavioural, not a transform failure wearing a survivor's clothes. The mandatory control (clean tree passing the same check first) was asserted before any mutant verdict was trusted.

PR #783, opened 2026-08-18
fix(N-23): host guidance invalidation on the live posture (wire-free)

Talchain:
**N-23 was real.** `clearGuidanceItems()` had exactly one production caller, in a hook whose only host is `DraftChat` — mounted only when `aiPanelV2` is OFF (`flags.ts:358` `defaultValue: true`). So coaching invalidation on graph edit has been **dark on the deployed posture**.

**Fix:** a new wire-free `useGuidanceInvalidationOnEdit`, reusing the *same* diff helpers so there is no second authority on "the graph changed", mounted via `GuidanceInvalidationHost` beside the two existing drain hosts. Exactly one host runs per posture — no double-clear.

**Inverse failure deliberately avoided:** re-hosting `useGraphEditEvents` would have switched on `direct_graph_edit` wire emission for every user — a wire change smuggled in as a UX fix. The new hook takes no transport and imports none, enforced at the source with contrast controls. `useGraphEditEvents` correctly stays in `KNOWN_DARK_DRAINS`; the wire half is still dark by design.

---

## Review disposition — the 7 fixes from the DO-NOT-MERGE review

⚠ **The review itself is not on this PR** (no reviews, no inline comments, no issue comments beyond the StackBlitz bot). It was recovered from the body of commit `6f2078cf`, which enumerates it. Recording that here so the next reader does not conclude from a clean comments tab that there was nothing to answer.

| # | Finding | Disposition |
|---|---|---|
| 1 | **P0-A / P0-B** — four PRODUCER writers had no `beginExternalGraphMutation` window, so mounting the hook destroyed coaching on the two most important journeys, and (because `clearGuidanceItems` also clears the blob) **the loss survived a reload** | **FIXED at the WRITER, not the call site** — `mirrorAnalysisReady.ts:77`, `applyDraftResult.ts:234`/`:596`/`:684`, `mergeAppliedGraph.ts:702`, `mergeServerGraph.ts:448`. Two of the four were found by a **surviving mutant**, not by inspection; `f72641cc` then measured which guard actually closes P0-B and labelled the redundant outer one in place |
| 2 | `useGraphEditEvents.ts` and `guidanceStore.ts` both asserted a suppression true only of the **guarded window** | **FIXED** — original wording quoted, then refuted, in place (2 sites + 1 site verified) |
| 3 | `stripComments` applied to the negative assertions and **not** to the load-bearing positive one; M1b left the central claim green | **FIXED**, M1b pinned permanently (`…reachability.production.spec.tsx:295-305`) |
| 4 | `drainHostReachability.derived.spec.ts` `imports()`/`callsHook()` were comment-blind — the control written to catch a stale import **passed on a commented-out call** | **FIXED**; drain set re-derived 5 → 4, `useSessionResumeEvent` reclassified `HOSTED_NOWHERE:264` |
| 5 | M6 (`<MaybeConversationProvider>` → `<>`) and M8 (flag inverted) left the spec 22/22 green | **FIXED** — render path and the literal flag guard now pinned (`:323`, `:328`) |
| 6 | The "SINGLE AUTHORITY" claim was not earned — the diff is shared, the ~30-line invalidation decision was copied | **FIXED** — divergence guard added; M2c proved the divergence was invisible without it |
| 7 | Stale line refs in `coachingCurrency.ts` | **FIXED** — replaced with derive-don't-trust |

All seven verified present **at the bytes** on this head, not inherited from the commit prose.

---

## ⚠⚠ Review round 2 returned BLOCK — a live regression, now fixed (`309f35fe`)

The independent review verified the §C latent-defect find (MUT-S/MUT-I/MUT-P all reproduced, plus a new **MUT-ORDER** — counter raised in a separate later `set()` — which also reds §C alone; that pin holds against removal **and** refactor). But it found that this PR **introduced a live regression reachable by 100% of deployed users**, of the predicate-breadth class.

**The defect.** `serialiseNode` stringifies the WHOLE `data` object, so `diffSnapshots` returned non-null for ANY `data` change and the new hook answered with a blanket `clearGuidanceItems()` — which also wipes the persisted blob, so the loss survived a reload. Four gestures wiping screen **and** disk: rename the goal (`store.updateNodeLabel`), inspector label rename, inspector **description** edit, inspector **category** edit. Newly reachable *because of this PR*: at base the only caller of `clearGuidanceItems()` was hosted where `aiPanelV2` is OFF, and it defaults TRUE — mounting the new host in the flag-ON branch took it from unreachable to universal.

**The codebase already owned the question.** `hasAnalyticalNodeChange` / `ANALYTICAL_NODE_DATA_FIELDS` exists to decide exactly this, and its registry **explicitly excludes** `label`, `body`, `description`, position and colour as cosmetic. The `diffSnapshots` path was a **fourth authority** answering the same question differently. The tell was two surfaces disagreeing on one gesture: strip and node markers destroyed while the transcript coaching card beside them still read `'current'`, because `coachingCurrency` asks the canonical owner.

### The three corrections

1. **Bound the predicate by CONVERGING, not by adding a fifth rule.** `hasAnalyticalGraphChange` is added to `domain/analyticalChange.ts` **itself**, deriving from the same registry; the hook defers to it. One deliberate difference from its per-update sibling, and it is not a new rule: the sibling takes an *update partial* so it can only ask `field in newData`; here both sides are complete objects, so the comparison is symmetric.
   **Plus the opposite-direction twins the corpus was missing** (§D): §B had exactly **one** "must not clear" twin — a drag. A contrast-controlled sweep found `position-only` in 8 places and cosmetic/label/description in **zero**. §D drives the **real store actions** with the exact payload shapes `useInspectorMutations` builds, has a control proving each gesture actually moved the graph, and a **contrast** proving an analytical edit through the same action still clears.
2. **The writer manifest.** Bounding the predicate **exempted three** of the ~10 named writers outright (`loadStarter` stamps `starterId` only; `saveScenario`'s two cleansing writes strip `_baseline_snapshot`, which the registry marks `ephemeral`) — recorded as exempt **with the field that makes them safe**, so the exemption dies if they start writing analytical data. The rest are guarded: `applyStructuralDeleteRevert`, `applyClarifierGraph`, `clearClarifierPreview`, `RecoveryBanner` (recovering unsaved work was destroying the coaching that belonged to it), the streamed-preview withdrawal, `applyV5State`'s graph_patch application, and both `optimisticFactorEdit` writes. Every counter is raised in the **same `set()`** as the write — MUT-ORDER proves a later one is too late.
   ⭐ `applyV5State` is guarded **at its call site**, and the reason is derived not assumed: it has **exactly one** production caller. The estate's rule is guard-the-leaf *because* a second caller usually exists; here there is none, and it takes an injected store, so a leaf guard means widening that interface and every double. Re-derive the caller count before moving it.
3. **The "SINGLE AUTHORITY" claim in the source is withdrawn**, quoted then refuted — the spec already said it was unearned. The FIX-6 divergence guard that asserted the two chains "ask the same questions" is **superseded**: it was pinning the very coupling that caused this. The chains answer **different questions** (trap 21), and the new invariant is that the invalidation **defers to the canonical owner and re-implements no field list**, with a contrast control proving the probe can tell the chains apart.

Also removed the last harness shape this file's own note bans (a host mounted and never unmounted before `reloadTheTab()`), and inverted the reachability spec's RELABEL case with its original wording quoted and refuted in place.

### ⚠ Known divergence, pinned as an EXACT set

A rename now survives the edit but still **not a reload**, because `generateGraphHash` puts `data.label` in the hash while `analyticalChange` calls it cosmetic. **Both are pre-existing** — the adoption gate and its hash were already on `staging`; this PR only mounts the host. Changing the hash also moves the autosave dirty-gate, a separate decision with its own blast radius. So the gap is **recorded in the suite** as `{label: dropped, description: kept, category: kept}`, and REDs if it grows **or** shrinks.

### Mutants — round 2, each guard bitten by a mutant that bites only it

| mutant | reds |
|---|---|
| **MUT-COSMETIC** restore the whole-`data` predicate | **§D only** (4 gestures + the divergence pin) |
| **MUT-I** remove `clearGuidanceItems()` | §B ×2, §C control, §D contrast — **§A fully green** |
| **MUT-S** remove the `hydrateGraphSlice` suppression | **§C ⭐⭐ only** |
| **MUT-ORDER** raise the counter in a later `set()` | **§C ⭐⭐ only** |
| **MUT-CLARIFIER** unguard `applyClarifierGraph` | **§E clarifier only** |
| **MUT-REVERT-GUARD** unguard `applyStructuralDeleteRevert` | **§E revert only** |
| **MUT-P** remove `persistCurrent` | §A ×4 and everything else that reads the blob |

Leading and trailing controls both 37/37 with applied-check 0; each mutant applied-check exactly 1; isolation proven by sentinel write with distinct inodes; restores from a pristine archive.

**Deferred, as the review directed:** the OFF×OFF flag cell, `evictStaleItems`' dead graph limb, `useModelActionApply.ts`'s apparent zero callers, and `diffSnapshots` ignoring edge `type` / being key-order sensitive.

---

## ⭐⭐ Domain 3 (Context / Memory) — the exit condition, now executable

Context/Memory was **JOURNEY-WITNESSED PARTIAL**: conversation + graph restore witnessed; **coaching persistence neither confirmed nor refuted**, because the witness run never ran an analysis, so `guidance.items.v1` was never written. `coachingSurvivesReload.acceptance.spec.tsx` (19 tests) is the executable half of that settlement — one journey rerun now has something to agree with.

**Two harms, pinned separately, because a fix for either one alone produces the other.** "Persist everything forever" closes §A and opens §B; "clear on every tick" closes §B and opens §A.

- **§A** the coaching a **real analysis** minted survives a **real reload** — minted by the real chain (`parseV5Response` → `extractPhase3FromV5Response` → `toStoreGuidanceItem` → `setGuidanceItems`, exactly `useConversation.ts:5074-5079`) from **three independent live analysis captures**; the reload round-trips graph + answer through the real autosave projection and `restoreAnalysisFromAutosave`, and the read-side graph hash is computed over the **restored** nodes, never the objects still in memory. Bound by **item id** and by **deep-equality against the producer's own values** — never against literals this spec invented.
- **§B** a **local structural edit** invalidates it, on screen **and on disk** — including the discriminating case the adoption gate cannot fake: edit the model and edit it **back**, so the hash returns to the stamped value and the gate would wave the stale items straight through. Only the invalidation keeps them out.
- **§C** ⭐⭐ **the interaction, which would have killed this domain and was unpinned.**

### ⭐ Measured at the captures — this moves which gate is load-bearing

A real analysis turn pins **every** guidance item with `valid_while.graph_hash` (`graph_hash_at_generation`) and **no `analysis_hash`**. All three captures agree (11 / 13 / 11 items). So on the live journey the **analysis-hash limb of `rehydrateGuidance` is never exercised**, and whether the user's coaching survives a refresh rests **entirely** on the UI-side graph-hash comparison. The existing `analysis_hash` corpus in `guidanceStore.rehydration.spec.ts` is a real pin on a limb the wire does not currently reach. Derive again before relying — it is a fact about what CEE emits.

### ⭐⭐ §C — the defect that was one store action away

`GuidanceInvalidationHost` is a **child**, so it subscribes **before** `ReactFlowGraph`'s boot effect writes the whole restored graph into an empty store — the largest structural change the hook will ever see — a few lines before `rehydrateGuidance` reads the blob. Unsuppressed, that clears the blob and **every user loses all coaching on every reload**, silently: §A is green (mounts no host) and §B is green (never touches the disk).

It is safe **only** because `hydrateGraphSlice` raises `_externalMutationActive` in the **same `set()`** as the node write (`store.ts:5886`) — a property of a store action three files from either mechanism, which nothing told either of them was load-bearing. Now something does, with a **control** writing the same bytes through a bare `setState` so the case is evidence about the *suppression* and not about a hook that never fires.

### Mutants — different mutants, different reds, measured

| mutant | §A | §B ⭐⭐ (edit-and-edit-back) | §C ⭐⭐ |
|---|---|---|---|
| **MUT-P** remove `persistCurrent` from `setGuidanceItems` | **RED** ×4 | GREEN | RED |
| **MUT-I** remove `clearGuidanceItems()` from the hook | **GREEN (all 14)** | **RED** | GREEN |
| **MUT-S** remove the `hydrateGraphSlice` suppression | GREEN | GREEN | **RED (only this)** |

Two clean discriminating pairs: §A is red under MUT-P and green under MUT-I (a pure persistence pin); §B ⭐⭐ is red under MUT-I and green under MUT-P (a pure invalidation pin). MUT-S produces exactly **one** red, bound to the property it names. Throwaway worktree outside the repo root, **isolation proven by a sentinel write** (distinct inodes), restores from a pristine archive, applied-check scoped to `src/` = exactly 1 per mutant, leading and trailing controls both 0 / 19-of-19.

⚠ **Two harness defects found by execution, both kept as notes in the file.** Simulating the boot with a bare `setState` instead of `hydrateGraphSlice` manufactures a total coaching loss the product does not have; leaving a host subscribed across the page teardown makes emptying the store look like the user deleting every node. Both produce a **false red that reads exactly like a real one**.

### Measured, not fixed — out of this lane's path

- `VITE_FEATURE_THREAD_PERSIST` / `VITE_FEATURE_THREAD_HYDRATE` are inert, and **none of their gates is in the guidance-persistence path** — zero overlap with the files this PR touches. Blast radius if removed: `flags.ts:237-244,488-489,556-557`, `lib/flagEnv.ts:104-105`, `services/threadService.ts:33,70`, `hooks/useThreadPersistence.ts:221,370,398`, `useConversation.ts:71,2837,2867`, plus specs. `useConversation.ts` is a high-collision file; doing it here would be "while we're here" work. **Smallest enabling change: one PR that deletes the two flag entries and inlines each gate to its always-on branch.**
- `conversation_turns` — the table still has **zero readers**, and there is a trap for the next lane that greps it: `recent_conversation_turns` in the debug bundle (`useDebugData.ts:4589`) is a **same-name different-thing** — `selectRecentConversationTurns(tracedPayloads)` reads in-memory network traces, not the table. Writers are `useThreadPersistence.ts:247` and `:312`, neither in this PR's path.

**Evidence:** lane specs 19/19 collected **by name**; neighbouring specs 166 passed across 8 files (arg count asserted = 8, no shrunk collect); typecheck gate **PASSED** (557 files / 2263 errors vs baseline 2272 — drift shrank, none added); eslint clean on the new file.

**CI on `309f35fe`:** `Staging Gate` **SUCCESS**; all four Full Test Suite shards, `TypeScript + Lint`, `Typecheck Gate Self-Test`, `Production Build`, `Security Audit`, `CEE Contract Validation` all success. `Visual Regression (advisory)` fails — the standing base red, not required and not this branch's.

**Unmeasured:** the live journey rerun. **DO NOT MERGE** — held for review.

### DIFF
--- a/src/canvas/ReactFlowGraph.tsx
+++ b/src/canvas/ReactFlowGraph.tsx
@@ -112,6 +112,7 @@ import { isAiPanelV2Enabled } from '../flags'
 import { ConversationProvider } from './conversation/ConversationContext'
 import { PanelApplyDrainHost } from './conversation/PanelApplyDrainHost'
 import { StructuralDeleteDrainHost } from './conversation/StructuralDeleteDrainHost'
+import { GuidanceInvalidationHost } from './conversation/GuidanceInvalidationHost'
 import { FloatingOlumiPanel } from './components/FloatingOlumiPanel'
 import { FirstUseComposer } from './components/FirstUseComposer'
 import { StarterProvenanceBanner } from './components/StarterProvenanceBanner'
@@ -2562,6 +2563,14 @@ export function MaybeConversationProvider({ children }: { children: import('reac
             the drain there alone shipped the capability dark. Same two-host
             shape as PanelApplyDrainHost above, for the same reason. */}
         <StructuralDeleteDrainHost />
+        {/* N-23 — guidance invalidation's flag-ON host. `clearGuidanceItems()`
+            had one production caller, inside `useGraphEditEvents`, whose only
+            host is DraftChat (mounted only when aiPanelV2 is OFF). With the flag
+            ON — every deployed context — stale coaching survived a local
+            structural edit. This host carries the coaching half ONLY: it takes
+            no transport, so it cannot switch on `direct_graph_edit` emission as
+            a side effect. */}
+        <GuidanceInvalidationHost />
         {children}
       </ConversationProvider>
     )
--- a/src/canvas/components/RecoveryBanner.tsx
+++ b/src/canvas/components/RecoveryBanner.tsx
@@ -58,7 +58,18 @@ export function RecoveryBanner() {
       }
     }
 
+    // ⚠ NOT A USER EDIT — RECOVERING THE USER'S OWN UNSAVED WORK. Without this
+    // window `useGuidanceInvalidationOnEdit` reads the restore as the user
+    // rebuilding their model from nothing and destroys the coaching that
+    // belonged to exactly this recovered graph, on screen and on disk. Raised in
+    // the SAME `set()` as the write; a later one arrives too late (MUT-ORDER).
+    // Read-then-write rather than the updater form: this is one synchronous
+    // click handler with no concurrent writer, and the object literal keeps the
+    // precise typing the updater form widened away (the typecheck ratchet caught
+    // that, which the named local gate alone would not have).
+    const suppressed = useCanvasStore.getState()._externalMutationActive + 1
     useCanvasStore.setState({
+      _externalMutationActive: suppressed,
       nodes: autosaveData.nodes,
       edges: autosaveData.edges,
       currentScenarioId: autosaveData.scenarioId || null,
@@ -69,6 +80,7 @@ export function RecoveryBanner() {
       ceeAnalysisReady: autosaveData.ceeAnalysisReady ?? null,
       selectedGoalNode: goalNodeId,
     })
+    useCanvasStore.setState({ _externalMutationActive: Math.max(0, suppressed - 1) })
 
     // Clear autosave after recovery and mark as dismissed
     clearAutosave()
--- a/src/canvas/conversation/GuidanceInvalidationHost.tsx
+++ b/src/canvas/conversation/GuidanceInvalidationHost.tsx
@@ -1,0 +1,39 @@
+/**
+ * Headless aiPanelV2 host for guidance invalidation on a local structural edit.
+ *
+ * ⚠⚠ THIS FILE EXISTS BECAUSE THE CAPABILITY WAS DARK FOR EVERY REAL USER, AND
+ * THE WHOLE SUITE STAYED GREEN. `clearGuidanceItems()` had exactly ONE
+ * production caller — `useGraphEditEvents.ts:293` — and that hook's only host is
+ * `DraftChat`, which `ReactFlowGraph.tsx:2484` mounts ONLY when `aiPanelV2` is
+ * OFF. The flag is ON in every deployed context (`netlify.toml:57` `"true"`), so
+ * a user could restructure their model while coaching minted against the
+ * PREVIOUS model stayed on screen — the guidance strip, the on-canvas node
+ * coaching markers and every inspector coaching section — until the next
+ * assistant turn happened to replace the whole list.
+ *
+ * This is the same two-host shape the estate already uses for
+ * `PanelApplyDrainHost` and `StructuralDeleteDrainHost`, and it is a SEPARATE
+ * component from both so that no file's name becomes a lie about what it hosts.
+ *
+ * ⭐ WHAT THIS HOST DELIBERATELY DOES NOT DO, and why the lane stopped here
+ * before: it does NOT re-host `useGraphEditEvents`. Doing that would fix the
+ * coaching defect and, as a side effect, switch on `direct_graph_edit` wire
+ * emission for every user — CEE would begin receiving a system event no flag-ON
+ * user currently sends. That is a wire-behaviour change wearing a UX fix's
+ * clothes, and it is a separate decision with a separate blast radius. So this
+ * host consumes `useGuidanceInvalidationOnEdit`, which takes no transport and
+ * imports none. The absence of the wire is enforced by source inspection in
+ * `guidanceInvalidationReachability.production.spec.tsx`, not by this comment.
+ *
+ * It needs no `ConversationContext`: it reads the canvas store and clears the
+ * guidance store, nothing else. It is mounted inside `MaybeConversationProvider`
+ * anyway, because that is where the flag-ON host set lives and where the
+ * derived reachability walk looks for it.
+ */
+
+import { useGuidanceInvalidationOnEdit } from './useGraphEditEvents'
+
+export function GuidanceInvalidationHost(): null {
+  useGuidanceInvalidationOnEdit()
+  return null
+}
--- a/src/canvas/conversation/GuidanceInvalidationHost.tsx
+++ b/src/canvas/conversation/GuidanceInvalidationHost.tsx
@@ -5,8 +5,13 @@
  * THE WHOLE SUITE STAYED GREEN. `clearGuidanceItems()` had exactly ONE
  * production caller — `useGraphEditEvents.ts:293` — and that hook's only host is
  * `DraftChat`, which `ReactFlowGraph.tsx:2484` mounts ONLY when `aiPanelV2` is
- * OFF. The flag is ON in every deployed context (`netlify.toml:57` `"true"`), so
- * a user could restructure their model while coaching minted against the
+ * OFF. The flag is ON for every fresh user by `flags.ts:358` `defaultValue: true` — and it is the DEFAULT, not the
+ * staging entry, that carries this claim. `netlify.toml:57` sits under
+ * `[context.staging.environment]`, which applies ONLY to staging builds
+ * (production inherits from `[build.environment]` alone), so citing it proves
+ * STAGING and not "every deployed context" — an overclaim this comment made
+ * until review caught it.
+ * So a user could restructure their model while coaching minted against the
  * PREVIOUS model stayed on screen — the guidance strip, the on-canvas node
  * coaching markers and every inspector coaching section — until the next
  * assistant turn happened to replace the whole list.
--- a/src/canvas/conversation/__tests__/coachingSurvivesReload.acceptance.spec.tsx
+++ b/src/canvas/conversation/__tests__/coachingSurvivesReload.acceptance.spec.tsx
@@ -1,0 +1,558 @@
+/**
+ * ⭐⭐ DOMAIN 3 (CONTEXT / MEMORY) ACCEPTANCE — COACHING PERSISTENCE, END TO END.
+ *
+ * WHY THIS FILE EXISTS. The Context/Memory journey witness was recorded
+ * JOURNEY-WITNESSED PARTIAL: conversation + graph restore were witnessed, and
+ * coaching persistence was NEITHER CONFIRMED NOR REFUTED — because the witness
+ * run never ran an ANALYSIS, so `guidance.items.v1` was never written. An
+ * unwitnessed capability is not a working one, and the prior lane was right to
+ * refuse to call it dead. This file is the executable half of that settlement:
+ * it drives the REAL producers of both halves so a single journey rerun has
+ * something to agree with.
+ *
+ * ⚠ TWO DIFFERENT HARMS, AND THEY MUST NOT SHARE A PREDICATE.
+ *   §A  COACHING VANISHES ON REFRESH. The user runs an analysis, gets eleven
+ *       pieces of coaching, reloads, and the strip, the on-canvas markers and
+ *       every inspector coaching section are empty.
+ *   §B  STALE COACHING SURVIVES A LOCAL STRUCTURAL EDIT (N-23). The user
+ *       changes the model and the advice about the PREVIOUS model stands.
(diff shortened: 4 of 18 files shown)


Talchain:
## HOLD LIFTED — merging. Body corrections recorded here rather than rewritten.

The body still opens `DO NOT MERGE — held for review` and states 19 tests for a spec that now collects **37**. I am correcting that here rather than editing the body, because a PR body is a dated record of what was believed at the time and this estate treats such records as append-only.

**User-visible outcome:** when a user changes their model in a way that affects the analysis, the coaching minted against the old model is cleared from the strip, node markers and inspector — instead of standing there confidently wrong. Cosmetic edits (rename, description, category, drag) no longer destroy it.

### Why the hold no longer applies — four questions settled

**1. Deployed posture, settled at the deployed artefact.** `aiPanelV2` is **ON** for a fresh staging user, and it is a **single flag, not a conjunction**. The bundle chunk carries `VITE_FEATURE_AI_PANEL_V2:"true"` **and** `{envKey:"VITE_FEATURE_AI_PANEL_V2", storageKey:"feature.aiPanelV2", defaultValue:!0}`. Resolution is localStorage → env → default; a fresh user has no localStorage key. 84 chunks / 5.67 MB crawled to closure, both positive controls firing and three fabricated controls at zero **in the same run**. `netlify.toml:57` agrees but sits under `[context.staging.environment]` — the PR's own comment correcting an earlier "every deployed context" overclaim was the right call.

**2. The defect is real and dark.** At the deployed tip, `ReactFlowGraph.tsx:2508` is `{!isAiPanelV2Enabled() && <DraftChat />}`; `useGraphEditEvents` has exactly one production call site (`DraftChat.tsx:153`); `clearGuidanceItems()` has exactly one production caller. **Flag ON ⇒ it never runs.**

**3. It mounts on the surface the flags actually render.** `<GuidanceInvalidationHost />` sits inside `MaybeConversationProvider`'s `if (isAiPanelV2Enabled())` branch, beside the two existing drain hosts. Exactly one host per posture; the OFF branch returns `<>{children}</>`. **No double-clear.**

**4. 18 files is not a split.** Stripping comments, the 14 source files carry **174 added code lines**; the rest is ~1,879 lines across 4 specs. The producer-writer guards (18 `begin/endExternalGraphMutation` calls) are a **precondition** for the mount — landing the host without them re-ships the P0 that destroys coaching on accept and draft-apply. They must go together.

### The corpus bites — verified independently
Worktree isolation proven by sentinel write (distinct inodes, source sha unchanged), restores from a pristine archive, applied-check scoped to `src/`:
- **C0** 76/76 at applied 0
- **MUT-I** (remove `clearGuidanceItems()` from the new hook) → **13 failed, 3 of 4 files red**
- **MUT-COSMETIC** (restore the whole-`data` predicate — the round-2 live regression) → **9 failed**
- **C1** 76/76 at applied 0, source clone `src/` dirty 0 throughout

### The invalidated state is honest, and not a dead end
`GuidanceStrip` returns `null` when empty — but the transcript's V5 coaching/evidence/review/exercise blocks survive and render *"Your model has changed since this was written — it may no longer apply."* (`coachingCurrency.ts:180`), because they read the CEE-side currency verdict rather than `guidanceItems`. Composer and thread remain, so the user keeps a route back.

**But there is no signal at the moment of clearing** — the strip simply vanishes. A one-line notice would close that. Rowed, not blocking.

### Two things recorded for whoever touches this next
- **Rollback now changes semantics, not just the surface.** The OFF-branch hook still clears via `diffSnapshots` (whole-`data`), so a rename destroys coaching there, while the ON branch uses `hasAnalyticalGraphChange`. An emergency `aiPanelV2=false` rollback would reintroduce the cosmetic-clear regression on the fallback path. Pre-existing, but newly asymmetric.
- **The blanket clear is over-broad but defensible.** `clearItemsByTargetIds` exists and is used elsewhere. An analytical change propagates through the graph, so a blanket clear is arguable — but it is a choice the PR does not explicitly defend.

### Rung
**TESTED** (76/76 across 4 specs, asserted by name). **Not MOUNTED** — the deployed bundle is pre-#783. The body's own *"Unmeasured: the live journey rerun"* stands: **the deploy-verify must produce a fresh-session journey witness before anyone calls this live.**

Gate at head `e2725961`: 14 checks, `running=0`, required `Staging Gate` **success**. The sole red is `Visual Regression (advisory)` — non-discriminating estate-wide (noise floor 5.5266% against a 0.05% tolerance; completeness guard reports zero screenshots), read as neither signal nor absolution.

Separately flagged for follow-up: the deployed `useAnalysisTrust` chunk passes a **hardcoded** `aiPanelV2On:!0` rather than reading the flag. Not in this PR's path, but it sits next to the flag this PR depends on.

PR #784, opened 2026-08-18
fix(D5): gate the chat run chip on analysisHeldOn

Talchain:
**Domain 5 · AnalysisState — the defining defect.**

`SuggestedChips` gated only on `ceeAnalysisReady.status === 'ready'` — CEE's verdict about *its own* graph. **Every other run affordance** (node chips, command palette, actions menu, inspector inline rerun, define-success modal) already routes through `executeCanonicalRun`, which applies `analysisHeldOn`. **The chat chip was the last bypass**, and `relabel-rerun` let it bypass even the readiness gate.

The RED run reproduced the disagreement as a test: `analysisHeldOn === 'starter'` and `canRunAnalysis().allowed === false` both passed **while the chip rendered**. That is P8 — offering an action the product cannot accept.

**Two authorities, deliberately kept distinct** — they answer genuinely different questions:
- *Cross-turn analysis state* → CEE `composeAnalysisStateV1`; UI single reader `analysisStateSelector`.
- *Whether this canvas's model is analysable at all* → `analysisHeldOn(nodes)`, **client-owned**. CEE cannot answer it — it cannot know the canvas holds a client-injected graph it never received. `analysis_state` cannot subsume it.

**Inverse failure guarded:** over-suppression — silencing legitimate runs, or silencing the chips that get the user *out* of the held state. Covered by opposite-direction twins and a run-path-OFF case. The deeper mirror (minting a third derived value) is avoided by adding a **reader** of the existing authority, not a new derivation.

---

## ⚠ CORRECTED AFTER REBASE — the "demonstrated-equivalent" mutant is no longer equivalent

Rebased onto staging `30db8ac3` and the documented mutants re-run at the rebased head. **One recorded result is now false and is withdrawn.** The original text is kept below rather than deleted, because a silently-overwritten label is the defect this section exists to correct.

> ~~**Author's own correction, kept in the record:** the first commit message claimed the hold *must* sit before the readiness gate. **Mutant M7 refuted it** — the same filter applied downstream survived 12/12. The narrower load-bearing claim: the hold must not be a *clause of* the readiness predicate (M3 red on 2 cases, due to its early `return true`s). M7 is recorded as a demonstrated-equivalent mutant.~~

**M7 now BITES.** Measured at the rebased head, applying the same filter downstream over `polished` reds **1 test**:

- `does not mis-report a HELD chip as removed-for-unreadiness in the dev log`

The reason is not that the code changed — it is that **the spec grew from 12 to 16 tests during review**, and one of the later tests pins the DEV-log mirror, which reads `admissible`. So the upstream placement genuinely is load-bearing for that assertion, exactly as the code comment beside it claims. **The PR is better pinned than this body said it was**; `M7 = demonstrated-equivalent` is withdrawn.

**M3 likewise reds 3, not 2** — same cause, the third being the dev-log test:

- `hides the chip on a held graph even via the relabel-rerun readiness-gate BYPASS`
- `hides a prompt-only "Rerun analysis" affordance on a held graph`
- `does not mis-report a HELD chip as removed-for-unreadiness in the dev log`

*Why this correction is worth making:* an honest "equivalent" label that has quietly become false teaches every later reader to stop looking at precisely the mutant that turned out to discriminate. A stale label is worse than no label.

**Mutant results at the rebased head** (throwaway worktree outside the repo root, isolation proven by *writing* a sentinel **and asserting it landed**, distinct inodes, source hash unchanged; restores from a pristine tar archive, never `git checkout --`; semantic applied-checks, not file counts; leading and trailing controls both GREEN 56/56 with `dirty=0`):

| mutant | reds |
|---|---|
| M-CORE remove the held-model filter | RED ×6 |
| M3 hold as a *clause of* the readiness predicate | RED ×3 |
| M7 same filter applied downstream | **RED ×1** (was recorded equivalent) |
| M-UNSUP revert the unsupplied-sentinel fallback | RED ×2 |
| M-CONVERGE re-declare the local sentinel | RED ×2 |
| M-GUARD-VACUITY break the convergence detector | RED ×2 (both positive controls) |

---

## Convergence — the unsupplied sentinel now has one owner

`analysisStateSelector.ts` declared its **own** `= 'unknown'` sentinel beside a docstring arguing it was therefore *"named ONCE, at the single seam that reads it"*.

**That claim was false when it was written**, not merely made false by this branch. `adapters/cee/types.ts:401` already exported `ANALYSIS_READY_STATUS_UNSUPPLIED` with the identical value, and already had a consumer. The premise reasoned about the wrong artefact: there is indeed no export in the CEE service or the contract package, but this repo's own cee-adapter has owned the constant all along. So the local copy was never the only name — it was a second authority under a comment asserting there was one.

**Canonical owner: `ANALYSIS_READY_STATUS_UNSUPPLIED` (`adapters/cee/types.ts`)** — the sentinel crosses the CEE boundary, so it is named at the boundary and consumed here. The local declaration is removed and its justification comment rewritten to state what is now true. **No behavioural change: both were `'unknown'`.**

**Why this was safe to converge rather than adjudicate — and it is the load-bearing point.** The two lanes reached the **same semantic rule independently**: the sentinel falls back to the last known verdict and is never read as "not ready". `selectAnalysisReadinessAuthority` returns a null authority (so `canRunAnalysis` falls back to the side-car verdict); `composeAnalysisState` falls back to `ceeAnalysisReadyStatus`. Same rule, two seams, arrived at separately. There was no disagreement to settle — only a duplicate name to remove.

**Pinned, and the pin is proven able to fail.** A convergence guard asserts this module declares no sentinel of its own and consumes the canonical one. It scans **comment-blanked** source, reusing the canonical scanner (`helpers/derivedCallSites.ts`) rather than a second copy of one — necessary because the module legitimately *quotes* CEE's declaration inside a comment as producer evidence, and over raw text the guard would red against correct code. It carries a discriminating pair: a fixture with a **real** declaration must be SEEN, one with only a **commented** declaration must be IGNORED.

- **M-CONVERGE** (re-introduce the local const) → **RED ×2**: `declares no sentinel of its own`, `the canonical owner is the string the comparison actually needs`.
- **M-GUARD-VACUITY** (break the detector regex) → **RED ×2**, both positive controls. The guard is not agreeing with itself.

**⚠ What the guard deliberately does NOT claim.** A **third** declaration of the same `'unknown'` string exists at `lib/coherence/crossSurfaceCoherence.ts:156` (exported, read by the coherence gate). It is **out of scope here** — it is not in this PR's diff, and pulling it in would create a file overlap with #788 and break the pairwise disjointness that currently lets these land in any order. The guard is therefore scoped to **this module, by name**, and asserts nothing repo-wide. A guard that claimed something false about the estate would be weakened the first time it fired honestly, and a weakened alarm teaches people to stop looking.

---

**Evidence at the rebased head.** Spec files re-run with counts asserted **by name** (a spec collecting `(0 test)` is invisible in every aggregate), `VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` exported:

- `SuggestedChips.heldModelGate.spec.tsx` — **16**
- `analysisStateSelector.unsuppliedReadiness.spec.ts` — **15** (was 8; +7 convergence guard)
- `blockedReasonStaleWiring.derived.spec.ts` — **5**
- `runGateCallSites.derived.spec.ts` — **20**

**56 passed / 0 failed / 4 files.** Both *derived* specs (which scan source) still pass against staging's 206-file delta. `Staging Gate` **SUCCESS**; the only red is `Visual Regression (advisory)`, which fails identically on staging's own untouched head `30db8ac3` and is absent from the gate's `needs`.

> Superseded pre-rebase evidence line, kept for the record: ~~RED-first 5 failed / 7 passed / **12 collected** → 12/12 … Mutants **6/7 bitten, 1 demonstrated-equivalent** … Regression `src/canvas/conversation/__tests__` 167 files / 2271 passed / 0 failed; coherence gate + selector 142/142.~~ The collected count is now 16, and the equivalence claim is withdrawn above.

**MOUNTED acceptance (not yet witnessed — rung is TESTED):** on staging, V5 canonical path, load a starter — the Analyse control says *"Analysis is held on a saved example. Re-draft it live to run one."* **and** chat offers no Run/Rerun chip. Click *Re-draft this live* → the chip returns.

Pairs with CEE #1004 (rebased `877affe2`, head `259e6c75`).

### DIFF
--- a/src/adapters/cee/types.ts
+++ b/src/adapters/cee/types.ts
@@ -393,10 +393,21 @@ export const ANALYSIS_READY_STATUSES = [
 export type AnalysisReadyStatus = (typeof ANALYSIS_READY_STATUSES)[number]
 
 /**
- * NOT a producer value — a sentinel the UI mints itself. `applyV5State.ts:262`
- * coerces a missing/non-string `analysis_ready.status` to `'unknown'` so the
- * lenient V5 normaliser never drops the payload. It means "CEE said nothing
- * about readiness on this turn", which is not a verdict of any kind.
+ * "CEE said nothing about readiness on this turn" — an ABSENCE, not a verdict
+ * of any kind, and specifically NOT a synonym for `blocked`.
+ *
+ * ⚠ CORRECTED. This said "NOT a producer value — a sentinel the UI mints
+ * itself". That was true when written and is now FALSE at both ends: CEE
+ * declares the identical string as `READINESS_STATUS_UNSUPPLIED`
+ * (`compose/analysis-state-v1.ts`) and emits it on the 12 of 22 turn exits that
+ * supply no readiness payload, `clarify_v2` included. The UI still mints it too
+ * (`applyV5State.ts:262` coerces a missing/non-string `analysis_ready.status`),
+ * so it now arrives by BOTH routes and means the same thing on each.
+ *
+ * ⚠ THE OBLIGATION THAT FOLLOWS, because a consumer already got this wrong:
+ * treat it as "fall back to the last known verdict", NEVER as "not ready".
+ * `analysisStateSelector` does exactly that; reading it as a negative removed
+ * the Run CTA from every pre-analysis clarification turn.
  */
 export const ANALYSIS_READY_STATUS_UNSUPPLIED = 'unknown'
 
--- a/src/canvas/conversation/__tests__/SuggestedChips.heldModelGate.spec.tsx
+++ b/src/canvas/conversation/__tests__/SuggestedChips.heldModelGate.spec.tsx
@@ -58,6 +58,14 @@ import { describe, expect, it, beforeEach, afterEach, vi } from 'vitest'
 import { render, screen, cleanup } from '@testing-library/react'
 
 import { SuggestedChips } from '../zones/SuggestedChips'
+
+// Mocked so the dev-log counter assertion (case 9) can observe emissions.
+// `importOriginal` is spread so every OTHER export stays real — a bare factory
+// REPLACES the module and silently removes anything it forgets (trap 12).
+vi.mock('../../../v5/debugLog', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../../../v5/debugLog')>()),
+  logV5StateEvent: vi.fn(),
+}))
 import { useCanvasStore } from '../../store'
 import type { ActionChip } from '../types'
 
@@ -289,9 +297,11 @@ describe('SuggestedChips — the held-model gate (PoC domain 5)', () => {
   //    all, this goes red for the right reason.
   // ───────────────────────────────────────────────────────────────────────────
 
-  it('agrees with canRunAnalysis: whenever the run gate refuses for the hold, the chip is absent', async () => {
+  it('agrees with canRunAnalysis: the gate refuses FOR THE HOLD, and the chip is absent', async () => {
     const { canRunAnalysis } = await import('../../utils/canRunAnalysis')
-    const { analysisHeldOn } = await import('../../utils/analysisHeldOnInjectedModel')
+    const { analysisHeldOn, ANALYSIS_HELD_NOTICE } = await import(
+      '../../utils/analysisHeldOnInjectedModel'
+    )
 
     setReady('ready')
     setNodes(STARTER_NODES)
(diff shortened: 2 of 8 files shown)


Talchain:
## Title corrected before merge — "the last bypass" was falsified

Retitled from *"close the **last** run-affordance bypass of `analysisHeldOn`"* to *"gate the chat run chip on `analysisHeldOn`"*.

An independent adversarial review enumerated every path that can start an analysis and found **two bypasses survive this PR**. Merging under the original title would have recorded a false *closed* verdict — and a false "closed" is worse than an open row, because it teaches the next session to stop looking.

**The code is unchanged and merges as reviewed.** It is a strict improvement. Only the claim was too broad.

### What survives, rowed separately

**B1 — the chip CLICK path (this PR discloses it itself).** The fix gates chip **visibility** at `SuggestedChips.tsx:253`; `ConversationPanel.handleChipClick` (`:187`, `await sendChip(chip)`) consults no gate. The PR's own comment at `SuggestedChips.tsx:176-178` says so. Visibility filtering alone is one refresh away from a bypass.

**B2 — `GuidanceStrip` "Discuss", not previously disclosed, verified at the bytes.** A button labelled **"Discuss"** starts a hidden `run_analysis` turn on a held graph:

```
GuidanceStrip.tsx:274   onSendMessage(prompt, { hidden: true, debugSource: 'guidance_discuss' })
useConversation.ts:6437  forwards hidden, turnType: undefined
useConversation.ts:2262  if (hidden) return 'run_analysis'      ← falls past every earlier clause
useConversation.ts:4540  isRunAnalysisTurn → resultsAnalysing() + claims activeRunTurnIdRef
```

Zero `canRunAnalysis`, zero `analysisHeldOn`. Mounted live via `ChatComposer.tsx:260` → `ConversationPanel.tsx:839`. **Pre-existing on staging, not introduced here** — but exactly the class the old title claimed closed. The two clauses above `:2262` (`analyse_now`, `right_panel_action`) are unreachable from any live caller, so `if (hidden)` is the only live arm: it is doing work nobody designed it to do.

### Verified under attack

- Trial-merges clean into staging `25486c2e`; 21 commits since base `30db8ac3`, **none touching its production files**. Not superseded.
- Specs pass at the exact claimed counts, asserted by name: 16 / 15 / 20 / 5 = **56/56**.
- Mutants with isolation proven by sentinel write, leading and trailing controls green: M-CORE (remove filter) → RED ×6; **M-INVERT** (opposite-direction twin) → RED ×10 including the run-path-OFF case, so **no over-suppression — the corpus watches both doors**; M-DETECTOR-NARROW → RED ×1, not vacuous. M-CORE and M-INVERT fail on **different assertion sets**, so the pair discriminates.
- It **consumes** `analysisHeldOn(s.nodes)` rather than recomputing — no client-side freshness derivation.
- The PR's own self-correction withdrawing its "demonstrated-equivalent" label on M7 is **accurate**.

### Two corrected premises, recorded so they are not re-inherited

1. The `OutputsDock` `runV2Analysis()` ungated-fallback concern is **stale**. ROADMAP 2.1229 retired the direct browser→PLoT `/v2/run` seam; it now returns `{ status: 'unavailable' }` and runs nothing (`OutputsDock.tsx:1379-1393`), already on staging.
2. `aiPanelTranche1.spec.tsx:84` fails **identically on pristine `25486c2e`** with no PR applied — not a regression here, but staging cannot pass its own spec and the spec file is unchanged since `30db8ac3`, so a production change broke it. Rowed separately.

Rung: **TESTED**, not witnessed — matching the PR's own honest claim.

PR #791, opened 2026-08-19
feat(coaching+stage): typed coaching intents actually leave the UI, and the stage vocabulary converges

Talchain:
**Domain 10 — Strategic Reasoning. Branch + review only; do not merge.**
UI half of **CEE #1042**. Builds on the banked `feat/typed-coaching-intent-arm` and **closes the transport gap that branch left open**.

## ⚠⚠ The banked branch did not actually close the motivating defect

It added `SendChipMeta.intent` and the `ConversationPanel` forward, so the **type layer** was complete — and **`ActionChip.tsx:70` still called `sendChip(label, message)` with no third argument**. That component renders CEE's widening card chip. So `action_intent` reached the DOM as `data-action-intent` and **stopped there**: exactly the hop ROADMAP 2.1288 names (`draft-option-widening-blocks.ts:720` authors it, `route-v2.ts:2819` requires it, nothing carried it between them).

A field that exists and is never populated is, from the type's point of view, **indistinguishable from one that works**. That is why the test for it is on a **rendered click**, not on the interface.

## What a user gets

Four **mounted** sparks — `pressure_test_frame`, `define_success`, `widen_options`, `reflect_bias` — carry `action_type: null` because no honest handler exists for a conversation. Their typed intent now reaches `chip.intent` on the wire, so CEE's arm can route them instead of re-inferring intent from prose. The card-hosted path (the widening card) works too.

The gate still **fails closed** and is still two-signal: `outside_view`, `pre_mortem`, `elicit_risks`, `estimate_help` are **declared and withheld**, because CEE does not route them. Listing an intent is the claim *"the deployed service routes this"* — so `CEE_ACCEPTED_INTENTS` grows only in lockstep with CEE's arm.

## Stage convergence: two more spellings removed

Beyond the three known vocabularies, the UI carried **two further declarations**, both hand-maintained mirrors:

- **`applyV5State.isStage`** hand-spelled the **canonical** union (`v === 'frame' || 'analyse' || 'decide' || 'review'`) — a fourth declaration of a set the shared schema owns and explicitly says to derive from. It reads **green while being wrong**: a member added in a re-vendor would be silently rejected as *"no stage signal"*. Now `Stage.safeParse`.
- **`guidanceEvents.profile_stage`** declared the UI union **minus `optimise`** — a fifth spelling, and a wrong one. Every writer reads `store.currentStage` (a full `ScenarioStage`), so **five call sites cast** to satisfy it and an `optimise` value shipped as an off-type string the declaration says cannot occur. *A type narrower than its only source is not a constraint; it is a lie with casts holding it up.* Derived from `ScenarioStage`; all five casts deleted.

Canonical is `Stage`/`StageType`. `ScenarioStage` survives behind **one** named edge, `src/v5/stageMapper.ts`. The guard pins one declaration per vocabulary, one edge, no fourth — and, the part a sweep cannot give you, that **the edge is total in both directions**, derived by iterating the contract so a re-vendor REDs here rather than silently writing `null` into a store column whose CHECK constraint would reject it.

## Verification

**Mutants — 5/5 bite** (applied-check ≥1 file each, restore clean each time, trailing control green). Isolation proven by **writing a sentinel** into the throwaway worktree and asserting the source was unchanged, not by locating the path.

| Mutant | Expected RED | Result |
|---|---|---|
| **U1 revert the `CEE_ACCEPTED_INTENTS` widening** (the brief's named mutant) | intent gate + seam | 3 tests RED across 2 files |
| U2 `ActionChip` stops forwarding the intent (the gap, restored) | chip transport | 3 tests RED |
| U3 `isStage` hand-spelled again | convergence | RED — **see below** |
| U4 a spark's declared intent dropped | seam contrast | RED |
| U5 the edge stops being total | mapper + convergence | 5 tests RED |

### ⚠ U3 SURVIVED the first round, and the fix is the most useful thing here

Restoring `isStage`'s hand-spelled union left the suite **entirely green**. The guard bound "a declaration" to a **list shape** (members separated by `|` or `,`) and was blind to the same re-declaration written as a **`||` comparison chain** — which is the form the real defect actually took. A survivor is a claim either way, and this one was **not equivalent**.

The predicate now catches both shapes, with the threshold at **three** members and the reasoning stated in the test: a two-way check (`stage === 'analyse' || stage === 'decide'`) is legitimate consumption and must not be flagged, so a reader can argue with the line rather than discover it. Both readings are pinned by controls. **The gap was found by mutation, not by reading.**

### Controls, each one differing from the others

- at least one spark **sent** and at least one **withheld** — so the per-case assertions cannot all be satisfied by a gate that accepts (or rejects) everything;
- a card with **no** `action_intent` sends **no** meta — so the transport assertion is provably about the producer's value, not a UI invention;
- the sweep must read **>500 files** and positively identify the one legitimate declaring file — an empty offender list is otherwise indistinguishable from a mis-rooted sweep;
- `HelpType = … | 'evaluate' | …` must **not** be flagged. A **homograph is not a vocabulary**, and sweeping on `evaluate` would have produced a false positive on a file this ruling has no business touching.

Cases are derived by **reading the mounted spark registry**, never a hand-typed intent list — a test spelling `'challenge_frame'` itself would pass while the spark a user clicks declared nothing, which is exactly how four mounted affordances sat untyped.

Two things measured rather than assumed: the registry **aliases** `pressure_test_frame` and `widen_options` across `ACTIONS_MENU` and `SPARK_PROMPTS` (first contrast assertion reported one spark twice — deduped by id); and the existing `intentGate` spec used `.has()` spot checks, which **cannot see an entry that should not be there** — replaced with `toEqual` on the whole set.

Local: typecheck gate **PASSED**, drift **shrank** 2272 → 2263 (9 fixed, none added — attributable to the removed casts and the derived predicate); eslint **0 errors** on every touched path; 7 touched specs **183/183**, collected counts verified **by name**.

## Findings, not fixed here (reported rather than crossed into)

- **`ChatThread.tsx:207` renders only the LAST assistant message's `actionChips`**, so any new assistant message wholesale replaces the previous turn's chip row. That is the UI-side mechanism by which a refusal turn takes away `Run analysis`. Out of scope here; the Run-affordance assertion lives at CEE (#1042), where the executor is the only party that can put the affordance in the payload.
- **`setStage` is dead.** `scenarioService.setStage` → `useScenario.setStage` → the `set_stage_and_log` RPC has **zero product call sites** (contrast control: seven sibling mutators return 1–3 each). So `scenarios.stage` is written exactly once, as the literal `'frame'` at insert — **the hub badge reads "Frame" for every scenario, forever**, and the Journey timeline collapses to a single "Frame" group. The RPC also holds a **second** copy of the five-member guard list. Deleting it is a capability question, not a convergence one.
- **`scenarioStageToV5Stage` is exported with no product callers** outside its own module (only `deriveV5Stage` uses it). Making it private would move `stageMapper.test.ts`; left alone.

### DIFF
--- a/src/canvas/components/ModelTabBody.tsx
+++ b/src/canvas/components/ModelTabBody.tsx
@@ -151,7 +151,7 @@ export const ModelTabBody = memo(function ModelTabBody({
       item_type: 'trust',
       surface: 'model_tab',
       scenario_id: state.currentScenarioId ?? undefined,
-      profile_stage: (state.currentStage ?? undefined) as 'frame' | 'ideate' | 'evaluate' | 'decide' | undefined,
+      profile_stage: state.currentStage ?? undefined,
     })
   }, [])
 
--- a/src/canvas/components/pre-analysis-v3/constants.ts
+++ b/src/canvas/components/pre-analysis-v3/constants.ts
@@ -469,37 +469,52 @@ export const ACTIONS_MENU = [
     label: 'Pressure-test the frame',
     prompt: 'Is this the right question to be asking, and does it fit my wider goals?',
     // Frame-challenge coaching. explain_from_structure EXPLAINS the model;
-    // this spark challenges it — no honest vocabulary entry.
+    // this spark challenges it — no honest ACTION vocabulary entry.
     action_type: null,
+    // ⭐ ROUTED. CEE's typed coaching arm serves this intent, so the click
+    // arrives TYPED instead of as anonymous prose CEE has to re-infer.
+    intent: 'challenge_frame',
   },
   {
     id: 'widen_options',
     label: 'Widen the options',
     prompt: 'Suggest materially different options that work through a different mechanism from the ones I already have.',
-    // Option-elicitation coaching; the boundary enum has no add_option.
+    // Option-elicitation coaching; the ACTION enum has no add_option.
     action_type: null,
+    // ⭐ ROUTED, and the only one of the four with a published protocol behind
+    // it: CEE cites DSK-P-004 (Opportunity cost prompting) on frame-stage
+    // turns, where the bundle's own stage_applicability admits it.
+    intent: 'elicit_options',
   },
   {
     id: 'outside_view',
     label: 'Take the outside view',
     prompt: 'Take the outside view on this decision: what do similar decisions and base rates suggest?',
-    // Base-rate coaching — no vocabulary entry.
+    // Base-rate coaching — no ACTION vocabulary entry.
     action_type: null,
+    // Declared but NOT yet routed by CEE, so the send gate withholds it. The
+    // declaration is the point: when CEE's arm grows this intent, the send
+    // lights up with no change here.
+    intent: 'outside_view',
   },
   {
     id: 'pre_mortem',
     label: 'Run a pre-mortem',
     prompt: 'Run a pre-mortem with me: imagine this choice failed a year from now. What went wrong?',
-    // Failure-imagination coaching — no vocabulary entry.
+    // Failure-imagination coaching — no ACTION vocabulary entry.
     action_type: null,
+    // Declared, not yet routed — withheld by the send gate.
+    intent: 'pre_mortem',
   },
   {
     id: 'risks_upside',
     label: 'Find risks and upside',
     prompt: 'What risks and best-case upsides are missing from my model?',
     // Gap-elicitation coaching (what is MISSING) — explain_* describes what
-    // is present; no honest entry.
+    // is present; no honest ACTION entry.
     action_type: null,
+    // Declared, not yet routed — withheld by the send gate.
+    intent: 'elicit_risks',
   },
   {
     id: 'calibrate_estimates',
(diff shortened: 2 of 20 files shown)


Talchain:
## CI polled to conclusion — required check GREEN, first run

`Staging Gate` **success** at `63c463d3`.

⚠ It reported `total=0` checks for the first five polls. **Zero checks is silence, never success** — polling continued until a check-run actually existed and completed, rather than reading an empty result as a pass.

| | |
|---|---|
| **Staging Gate** (required) | ✅ success |
| TypeScript + Lint | ✅ |
| Full Test Suite — shards 1/4, 2/4, 3/4, 4/4 + Summary | ✅ all |
| Production Build · Install & Cache · Typecheck Gate Self-Test | ✅ |
| CEE Contract Validation · Flag Deployment Drift · Security Audit (pnpm) | ✅ |
| Visual Regression (advisory) | ❌ — **standing base red, derived not inherited**: identical `failure` on `staging` head `aa916511`, which is this branch's exact base |

`mergeable: true`, `mergeable_state: unstable` (the advisory Visual Regression only). No `DIRTY`. Base `staging` has not moved since this branch was cut.

Note the **Typecheck Gate Self-Test** is green alongside the gate itself — so the gate that reported the drift SHRINKING (2272 → 2263) has independently proven it still bites.

PR #792, opened 2026-08-19
fix(analysis): the Analyse control stops refusing a model the producer just called ready — and the chat stops swallowing the click

Talchain:
## The blocker, at the wire

Frozen quartet (UI `aa916511` · CEE `a61fe7f` · PLoT `fb63b03` · ISL `28fe0c95`), fresh guest, governed brief `04-conflicting-constraints`, 16-node model drafted successfully. From `useCanvasStore.getState()`:

```
analysisStateV1.readiness = { status: "ready", blockers: [] }   <- CANONICAL: ready, ZERO blockers
analysisStateV1.run_state = { kind: "never_run" }
ceeAnalysisReady.status   = "ready"   (all three options ready)
analysisRefusalNotice     = null
```

…and `[data-testid="pre-analysis-v3-analyse"]` was **`disabled: true`**, titled:

> *"Olumi needs something more from this model before the next analysis. Ask in the chat and it will explain what is missing."*

**That sentence is false**, and the escape route it named was worse: the chat's "Run analysis" produced no analysis, no refusal, no error. A fresh user with a ready model had a button that lied and a chat affordance that did nothing.

## Root cause: two readiness notions under one name (trap 21)

| | |
|---|---|
| **COMPETITOR** | `readinessStore.readiness.can_run_analysis` — a **side-car** assessment fetched from `/bff/cee/graph-readiness`. The only thing `canRunAnalysis` ever asked. It can object while naming nothing: its empty-canvas verdict carries five fields and **no structured cause**, so a retained one lands on `BLOCKED_REASON_COPY.unspecified` **by construction** — the exact sentence on the build. |
| **CANONICAL** | `analysis_state.readiness`, stated by the producer on every turn. The contract is explicit that an empty `blockers` list *"is a POSITIVE claim: the producer assessed readiness and found nothing blocking"*. It reached the store, was exposed by `analysisStateSelector` as `readinessStatus`/`readinessBlockers` — and had **zero product consumers**. Computed, delivered, unread. |

Established by elimination at the bytes: every earlier rung of `canRunAnalysis` returns a *different* sentence (`nodeCount === 0`, `analysisHeldOn`, `draftStreamPhase`, validation blockers), and inside `composeReadinessBlockedReason` only rung 4 can emit `unspecified` — reached when the verdict refuses and carries no `goal_node_valid`, no `options_total`, no not-ready options. `ceeAnalysisReady` had all options `ready`, so rung 1 was empty. Rung 4 is the only survivor.

## The convergence — name the owner, supersede the rival, add no parallel rule

`readinessObjectsToRun` stays the **one** definition of a readiness objection and gains the canonical authority as its first argument:

- when the producer has spoken, the side-car is **not consulted** — not weighted, not OR-ed, not a tiebreak;
- when the producer is silent, the side-car answers exactly as before, so every pre-0.46 fixture is byte-identical;
- the precedence lives **inside the predicate**, so the render gate, the dispatch barrier and the panel cannot drift apart;
- the blocked **sentence** is composed from whichever authority decided, never from the other one.

The predicate is `blockers.length > 0`, **not `status`**. Gating on a status whose cause cannot be named would re-create the refusal-without-a-reason this deletes — and a run the producer then declines still refuses *with a stated reason*, which beats a control that lies about why it is dead.

## Second, independent defect: the silent refusal

`ConversationPanel.handleRunAnalysis` opened with a bare `if (!runGateResult.allowed || isAnalysisRunning) return`. **Even a correctly closed gate said nothing.** It now surfaces the same sentence the composer button already shows on hover, through the same toast the canonical runner uses for the same refusal in `OutputsDock`.

It deliberately does **not** write `analysisRefusalNotice`. That slice is sourced only from a typed CEE refusal on `response.analysis_ready`; fabricating one client-side would claim the producer refused a turn that was never sent — the guarantee-theatre class, not a cure for it.

## Third surface, caught by the mounted spec

`usePreAnalysisModel` derived its own `canRun` off the side-car. The new mounted test found an **ENABLED** "Analyse first pass" sitting directly beneath the line **"Not ready for analysis yet"** — the same defect class, created by fixing it one layer up. It now asks the same predicate instead of restating it. Its stale "keyed on `readiness == null` ALONE" comment is amended rather than left to mislead.

## Evidence

**RED-first at pristine `aa916511`**, in a worktree outside the repo root, isolation proved by a **write test** (sentinel written to the worktree, asserted absent from the source clone) and distinct inodes: **22 collected, 14 failed / 8 passed**. The 8 passing are the legacy contrast controls — they prove the specs bind to the real gate and are not vacuous.

**After the fix: 25/25 green** — 17 gate + 3 mounted `OutputsDock` + 5 mounted `ConversationPanel`. Collected counts asserted **by name**, per file, not inferred from a suite total.

**Five mutants, each with a distinct red set**, applied-checked against a pristine archive (restores from the archive, never `git checkout --`), with a trailing control restoring 25/25:

| Mutant | Reds |
|---|---|
| M1 delete the supersession | the **ENABLED** tests only |
| M2 always permit (`blockers.length > 0` → `false`) | the **DISABLED/naming** tests only |
| M3 bind to `status` not `blockers` | **exactly one** test |
| M4 compose the reason from the superseded authority | the naming tests only; the gate tests still pass |
| M5 restore the silent `return` | **exactly** the three silence tests |

M1 and M2 are the acceptance pair: different mutants, different reds, neither red set containing the other's.

**Scoped suites** (no full-suite run — seven lanes were live): 2533 passed across `utils` / `pre-analysis-v3` / `state` / `stores`; 4047 passed across `components` / `conversation`.

**Required-check steps run in order at this tip**: `lint` (0 errors), `typecheck` (gate PASSED; drift **shrank** 2272 → 2263, `--update-baseline` **declined**), `ci:guard:zustand`, `ci:guard:starters`, `check:vendor`, `ci:guard:schemas-version`, `ci:guard:ds:enforce`, `build:ci` (bundle 46.84 KB gzip, within budget).

## Two findings for the record — not fixed here, deliberately

1. **A third readiness predicate exists and is untouched.** `SuggestedChips` gates a `run_analysis` chip's **visibility** on `useAnalysisStatus() === 'ready'`, i.e. `ceeAnalysisReady.status`. On the captured state that read `ready`, so the chip **rendered** — this predicate is not part of the defect. It answers a different question ("may we offer this chip?"), and converging it is a separate reviewable change. Flagged rather than widened.
2. **A cross-lane hypothesis, refuted at this tip.** `usable_for_*` gates **nothing** in the run path. Complete non-test manifest in `src/`: `analysisStateSelector.ts` (exposes them) and `lib/coherence/crossSurfaceCoherence.ts` (the diagnostic). `SuggestedChips.tsx` contains **zero** occurrences — contrast control in the same file: `READINESS_GATED_ACTIONS` returns 3, so the sweep sees the file. Consistent with the brief's own read that the post-run flags were correctly `false`.

**Scope note on the chat click.** Two production `canRunAnalysis` call sites exist at this tip (`OutputsDock`, `ConversationPanel`); both now read the canonical authority. The only chat path that can consume a click and emit *nothing* — no turn, no toast, no error — is the bare `return` fixed here, reached via `useGuidanceStore._runAnalysis`. A CEE-*suggested* `run_analysis` chip routes to `sendChip` and always sends a turn, so it cannot be silent client-side. **If a live capture shows the silent click was that path, the remaining silence is CEE-side and outside this fix** — worth one re-capture on the deployed build.

### DIFF
--- a/src/canvas/components/OutputsDock.tsx
+++ b/src/canvas/components/OutputsDock.tsx
@@ -156,6 +156,10 @@ import { verboseDebug } from '../../utils/verboseLog'
 import { AnalysisFooter } from '../shared/AnalysisFooter'
 import { derivePostFooterStatus, derivePostFooterMeta } from './utils/postAnalysisFooter'
 import { useGraphReadiness } from '../hooks/useGraphReadiness'
+import {
+  selectAnalysisReadinessAuthority,
+  useAnalysisReadinessAuthority,
+} from '../state/analysisStateSelector'
 // ROADMAP 2.635 (I-4) — read at DISPATCH time, not via a render-scope selector:
 // the whole point of the licence barrier is to see the store as it is when the
 // run actually goes out, not as it was when the gate was computed.
@@ -1081,6 +1085,12 @@ function OutputsDockBody({ sendMessage }: OutputsDockBodyProps) {
   // ROADMAP 2.635 — `stale` feeds the gate's COPY (I-3) and `verdictAtMs`
   // identifies WHICH verdict licensed a run (I-4).
   const { readiness, stale: readinessStale, verdictAtMs: readinessVerdictAtMs } = useGraphReadiness()
+  // ⭐ The CANONICAL readiness authority — `analysis_state.readiness`, stated by
+  // the producer on the turn. Read beside the side-car verdict deliberately: the
+  // two used to be one name with two meanings, and seeing them on adjacent lines
+  // is the cheapest possible reminder of which one decides. `canRunAnalysis`
+  // applies the precedence; this surface never does.
+  const analysisReadiness = useAnalysisReadinessAuthority()
 
   // C1 review: the orphan-banner footer suppression is GONE. It rested on a
   // premise that is false at this ref — it claimed the footer would carry
@@ -1144,6 +1154,7 @@ function OutputsDockBody({ sendMessage }: OutputsDockBodyProps) {
   const runGateResult = canRunAnalysisUtil({
     graphHealth: graphHealth ?? null,
     readiness,
+    analysisReadiness,
     hasBlockers: hasValidationBlockers,
     nodeCount: nodes.length,
     isRunning,
@@ -1220,12 +1231,22 @@ function OutputsDockBody({ sendMessage }: OutputsDockBodyProps) {
     // `readinessObjectsToRun`, the gate's own rung predicate, so there is still
     // exactly one definition of what a readiness objection is (I-5).
     const licenceAtDispatch = useReadinessStore.getState()
+    // Re-read at DISPATCH time, not closed over from render: the flush above is
+    // a real window and a fresh turn can land inside it. Reading the render-time
+    // value here would answer with a verdict the store may have already replaced.
+    const analysisReadinessAtDispatch = selectAnalysisReadinessAuthority(
+      useCanvasStore.getState().analysisStateV1,
+    )
     if (
       verdictLicenceSuperseded(licensedByVerdict, {
         verdictAtMs: licenceAtDispatch.verdictAtMs,
         stale: licenceAtDispatch.stale,
       }) &&
-      readinessObjectsToRun(licenceAtDispatch.readiness)
+      // ⭐ Same precedence at the barrier as at the gate. Without the second
+      // argument this asks the SIDE-CAR about a run the PRODUCER licensed, so a
+      // gate that correctly opened would be re-refused between the click and
+      // the wire — the silent-death class, re-created by omission.
+      readinessObjectsToRun(licenceAtDispatch.readiness, analysisReadinessAtDispatch)
     ) {
       console.warn('[OutputsDock] run licence superseded before dispatch', {
         licensedByVerdictAtMs: licensedByVerdict.verdictAtMs,
--- a/src/canvas/components/__tests__/OutputsDock.canonicalReadinessWiring.spec.tsx
+++ b/src/canvas/components/__tests__/OutputsDock.canonicalReadinessWiring.spec.tsx
@@ -1,0 +1,314 @@
+/**
+ * THE SENDABLE BLOCKER, REPRODUCED ON THE MOUNTED SURFACE AND KILLED THERE.
+ *
+ * ── THE CAPTURE THIS FILE IS BUILT FROM ────────────────────────────────────
+ * Frozen quartet, 19 Aug 2026, fresh guest, governed brief
+ * `04-conflicting-constraints`, 16-node model drafted successfully.
+ * `useCanvasStore.getState()` read:
+ *
+ *   analysisStateV1.readiness = { status: 'ready', blockers: [] }
+ *   analysisStateV1.run_state = { kind: 'never_run' }
+ *   ceeAnalysisReady.status   = 'ready'    (every option 'ready')
+ *
+ * …and `[data-testid="pre-analysis-v3-analyse"]` was DISABLED, titled
+ * *"Olumi needs something more from this model before the next analysis. Ask in
+ * the chat and it will explain what is missing."* The producer had just said
+ * nothing was missing. The product asserted an untruth about the user's own
+ * model, and the chat route it pointed at was itself silent.
+ *
+ * ── WHY A MOUNTED SPEC, WHEN THE GATE IS ALREADY PINNED PURE ───────────────
+ * `canonicalReadinessAuthority.spec.ts` proves `canRunAnalysis` answers
+ * correctly. It proves NOTHING about whether the dock asks it correctly — and
+ * this repo has the receipt: the sibling `OutputsDock.blockedReasonWiring.spec`
+ * exists because reverting the dock's wiring hunk alone left 268/268 tests green
+ * while the headline behaviour vanished. A pure-function fix nothing mounts is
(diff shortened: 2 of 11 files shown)


Talchain:
## Round 2 — review findings addressed at `b043023b`

Both permissive routes closed. **`Staging Gate` = success** on this head (all 4 full-suite shards, TypeScript + Lint, Typecheck Gate Self-Test, Production Build). The only outstanding check is `Visual Regression (advisory)` — not required, and proven to fail identically on the base staging tip `aa916511`.

### F1 — the `unknown` sentinel (applied as specified)

Confirmed at the producer, not inherited: `READINESS_STATUS_UNSUPPLIED = 'unknown'` (`orchestrator-v5/compose/analysis-state-v1.ts:106`), substituted at `:314` whenever `canonical.status` is not a non-empty string, with the full `analysis_state` still emitted around it and `mapWireBlockers` returning `[]` for an absent list (`:180`). `selectAnalysisReadinessAuthority` now returns `null` for it.

**One correction to the premise, which strengthens it.** The vocabulary is **five** stated values (`ANALYSIS_READY_STATUSES`, `canonical-analysis-state.ts:137-142`), not a six-member allowlist — `'unknown'` is not in that set. It is the sentinel substituted when `coerceReadyStatus` returns `null`, minted in a **different file**. Folding it in as a sixth value is precisely the conflation the guard undoes. The enum *is* closed: `coerceReadyStatus` returns `null` for anything unrecognised.

### F2 — `status !== 'ready'` is REFUTED; a narrower derived clause ships instead

Refutation check 1 fires. The payload-status chain at `cee/transforms/analysis-ready.ts:958-969`:

```
needs_user_input   <- dedupedBlockers.length > 0
needs_user_mapping <- unreachableControllableBlockers.length > 0
needs_user_mapping <- hasIncompleteOptions
needs_encoding     <- hasEncodingNeeded
ready              <- otherwise
```

- `needs_user_mapping` fires on an unconnected controllable factor — which that same file's payload step calls **"informational, alongside existing blockers"**. A model with fully-encoded options and one unconnected factor is analysable.
- `needs_encoding` is exactly the **UI-SEM-091 scaffold state**, whose shipped point (CEE #612) is that *"the graph is runnable even though `can_run_analysis` is false"* — the run triggers the draft.

Refusing on those re-creates the original blocker in a new spelling **and deletes a shipped capability**. Check 3 corroborates: CEE's integrity guard flags only `status_ready_with_actionable_blockers` (`canonical-analysis-state.ts:494`) and deliberately excludes advisory blockers that *"ride along on otherwise ready payloads by design"*. There is **no reverse guard**, because non-ready-yet-runnable is normal there.

**But your witness concern is real and separable.** `blocked` is provably different from the other three: it is **absent** from that priority chain (0 occurrences; contrast control `needs_encoding` = 2 in the same range) and is written only by hard-block and refusal paths (`analysis-ready-helper.ts:1117`, `:1123`, `buildAnalysisRefusalReadiness:1440`). So the predicate is now:

```ts
return analysisReadiness.status === ANALYSIS_READINESS_BLOCKED
    || analysisReadiness.blockers.length > 0
```

A refused turn no longer hands back an enabled button; the scaffold and informational-factor cases keep running.

### AC1 re-derived

The old case used `{ status: 'unknown_code' }` — a fixture the producer cannot emit, justified by the contract's `describe()` prose rather than the emission (trap 13c). Replaced with a **case per stated status, bound by literal**: `ready` / `needs_user_mapping` / `needs_encoding` / `needs_user_input` with `blockers: []` **permit**; `blocked` with `blockers: []` **refuses**. The empty-list refusal sentence is pinned as honest — no quoted label, no count, no identifier.

### Mutants — four, distinct red sets, trailing control 32/32

| Mutant | Reds |
|---|---|
| **M-a** delete the `unknown` early return | the two sentinel tests **only** |
| **M-b** sentinel escape also fires for `blocked` | the DISCRIMINATION test only — **sentinel tests stay GREEN** |
| **M-c** widen to `!== 'ready'` | the runnable-status tests RED |
| **M-d** drop the `blocked` clause | the refusal-turn tests RED |

M-a/M-b are the discriminating pair: bound to `unknown` **by identity**, not a widened escape. M-c proves the predicate does not over-refuse.

⚠ **The first M-b was botched and scored a false survivor** — the replacement appended to `return null` instead of the condition, so it changed the file (applied-check passed) without changing the rule. Caught by reading the mutated line back before believing the result; re-run against the condition. An unapplied mutation is indistinguishable from an equivalent one without that read-back.

### Removal trigger added

The side-car fallback is **not** a pre-0.46 shim — `analysisStateV1` is turn-scoped and cleared on silence, parse failure, import, reset and scenario switch, so the branch fires routinely mid-journey. The code now names the deletion condition: the store retaining the verdict (with its own staleness mark) or CEE re-supplying it on every turn. It is a retention-policy condition, not a schema version, and no date makes it true.

### Correction to the round-1 figures

The `4047 passed` scoped figure in the description ran inside the sibling lane's full-suite contention window. Re-derived: **4685 passed / 0 failed across 363 files**. An all-green run cannot be a contention artefact — that failure mode manufactures false reds only. CI's 4-shard full suite on this head is the authority in any case, and it is green.

**Not touched, still rowed to you:** the CEE-side repair of `needs_user_mapping`/`needs_encoding` emission and the dropped unrecognised `blocker_type`.

PR #802, opened 2026-08-19
DO NOT MERGE ALONE — fix(readiness): send scenario_id so the panel and the run assess the same model

Talchain:
## ⛔ DO NOT MERGE ALONE

**This change on its own makes the readiness panel agree by reporting the value MISSING too — which is CORRECT, but reads as a regression to a tester.** Where the persisted scenario graph lags the canvas, the panel will begin reporting what the analysis turn has been seeing all along. Agreement is the point; the lag is closed by the coupled CEE `options[]`-sync change, not by a projection that flatters.

**It therefore lands WITH the coupled CEE PR, not before.**

**Coupled CEE PR: Talchain/olumi-assistants-service#1049** — `fix/options-interventions-sync-at-persist` (required check `Lint, TypeCheck, Unit Tests` = success; `Typecheck Drift (ratchet)` = success).

---

## The defect

Two readiness authorities disagree because the UI has never sent `scenario_id` to CEE's graph-readiness endpoint.

Derived at CEE `cbc3ea3d44e0f117872b2fc7a0faa2efafcecc9f`, `src/routes/assist.v1.graph-readiness.ts`:

```ts
const GraphReadinessInput = z.object({
  graph: Graph,
  analysis_ready: AnalysisReadyPayload.optional(),
  scenario_id: z.string().min(1).optional(),          // :118
});
...
let assessedGraph: unknown = input.graph;             // :312
let assessedFrom = "request_graph";
if (input.scenario_id) {
  const persisted = await loadPersistedScenarioStateStrict(input.scenario_id);
  if (persisted.graph != null) { assessedGraph = persisted.graph; assessedFrom = "persisted"; }
}
const admission = assessRouteAdmission(assessedGraph);
```

CEE runs **one predicate** (`assessRouteAdmission` → `assessCanonicalAnalysisReadiness`, the same whole-model authority the turn path uses) over **one of two inputs**, and the **caller** picks which. The run path always reads the persisted scenario. This projection never sent the id, so the panel has always been answered over `request_graph` — the browser's local copy. Same predicate, two inputs. That is how the panel can read "3/3 ready" about a model the analysis turn never sees.

## The change

`src/canvas/stores/readinessStore.ts` (verified line numbers, at `70adaa39`):

| What | Pristine lines | Now |
|---|---|---|
| `WATCHED_ROOTS` | 236–241 | gains `'currentScenarioId'` |
| `ReadinessPayloadInputs` | 244–249 | gains `currentScenarioId: string \| null \| undefined` |
| `buildReadinessPayload` | 274–377 (sole production call site of the projection) | emits `payload.scenario_id` |

`currentScenarioId` is already on the object passed in — `src/canvas/store.ts:430` (`currentScenarioId: string | null  // Active scenario ID`), so no new plumbing. The brief's pointer said `store.ts:421`; the field is at **`src/canvas/store.ts:430`**, and the file is `src/canvas/store.ts`, not `src/canvas/stores/store.ts`.

## State-class: the guest case is a correctness constraint, not a nicety

A guest, or a canvas drafted but not yet saved, **has no scenario id**. For those callers `request_graph` is the correct input, and CEE says so in its own comment ("the fallback is not a failure path"). So:

- **the key is omitted entirely** when there is no id — `null`, `undefined`, or `''`;
- **an empty string is never sent.** CEE types the field `z.string().min(1)`, so `""` is a validation failure, not a benign no-op — it would 400 every guest readiness request.

Both halves are tested separately, and the mutant that removes the guard (M2) REDs the guest tests, so those absence assertions are not vacuous.

## The `WATCHED_ROOTS` completeness guard fired, as designed

`readinessStore.invalidation.spec.ts` Proxy-wraps the state handed to `buildReadinessPayload`, records every top-level property read, and asserts that set is a subset of `WATCHED_ROOTS`. Adding the read without the watcher REDs it — measured, not assumed:

```
FAIL  ... > the watched-root list is derived from the payload builder, not from memory
      > reads nothing from canvas state that is not a watched root
AssertionError: expected [ 'currentScenarioId' ] to deeply equal []
```

Saving an unsaved canvas changes **nothing** about the model and **everything** about the question — CEE switches from assessing the request graph to assessing the persisted scenario — so it must be re-asked. That behaviour is pinned by a test, not just by the list.

## Evidence

**RED-first at pristine** (source untouched, new spec only), 9/9 collected by name, **5 failed | 4 passed**:

| Test | Failure at pristine |
|---|---|
| positive control: the helper builds a real payload and CAN emit scenario_id | `expected false to be true` |
| sends the scenario_id the store holds, by identity, when the canvas IS saved | `expected undefined to be 'scn_alpha_7c31'` |
| names currentScenarioId as a watched root, so saving re-asks the question | `expected [ 'nodes', 'edges', …(2) ] to include 'currentScenarioId'` |
| posts scenario_id for a SAVED canvas, bound to that canvas's id | `expected undefined to be 'scn_wire_1d4b'` |
| asks again when an unsaved canvas is SAVED, because the assessed model changes | `expected "spy" to be called 2 times, but got 1 times` |

The four that passed at pristine are the guest/absence guards — correctly green, because at pristine nothing ever sent the field. They are made non-vacuous by M2/M3 below.

**Mutation check** — throwaway worktree at an absolute path outside the repo root, mutating only committed state, restores HEAD-relative (`git checkout HEAD -- <path>`, never index-relative), clean-tree assertion before and after every mutant, applied-check scoped to `src/` (exactly 1 changed path per mutant, exactly 0 for the controls), per-spec collect counts asserted **by name** (`scenarioIdConvergence=9/9`, `invalidation=19/19`) on every run. Isolation was proven **by writing** a sentinel into the worktree and asserting the source's SHA-256 was unchanged, plus an APFS inode comparison — not by locating the path.

| Mutant | Result |
|---|---|
| **Control** (pristine) | exit 0, 28 passed |
| **M1** revert the emission | 4 RED — every presence assertion; guest assertions correctly stay green |
| **M2** drop the state-class guard (send unconditionally) | 4 RED — guest, empty-string, wire-guest, save-reasks |
| **M3** send `''` instead of omitting | 5 RED — guest, undefined, empty-string, wire-guest, save-reasks |
| **M4** hardcode one id (**discriminating pair**, trap 19) | 3 RED — the `scn_alpha_7c31` half stays **GREEN**, the `scn_beta_4e08` half REDs with `expected 'scn_alpha_7c31' to be 'scn_beta_4e08'`. That RED/GREEN pair is what proves the assertion binds by IDENTITY rather than by a value predicate any string could satisfy |
| **M5** drop the watched root, keep the emission | 4 RED — the Proxy completeness guard, the hardcoded-list guard, the watched-root assertion, and the behavioural save-reasks test |
| **Trailing control** | exit 0, 28 passed — identical to the first control |

**Gates run locally at this tip** (derived from `package.json`, not habit):
- `pnpm run typecheck` → **PASSED**. Coverage 3562/3602 tracked TS files loaded (40 declared out of scope); ratchet 557 files / 2263 errors against a 2263 baseline — no new diagnostics.
- `pnpm run lint` → **0 errors** (1176 pre-existing warnings), plus `lint:hooks-ratchet` exactly at its 229-violation baseline.
- Every spec in the repo that references `readinessStore` / `WATCHED_ROOTS` / `useReadinessStore` — **33 files, 335 tests, all passing** (arg count asserted at 33 before the run; `VITE_SUPABASE_URL`/`VITE_SUPABASE_ANON_KEY` exported throughout).

The named local gate is a fast proxy, not the authority — **`Staging Gate` is.**

## Blast radius

`buildReadinessPayload` has exactly four call sites, all inside `readinessStore.ts` (`:453`, `:490`, `:879`, `:992`) — the projection, the empty-canvas arm, the post-fetch identity re-check, and the change detector. All four take the canvas store state whole, so all four pick the field up. `POST /bff/cee/graph-readiness` is the only route that receives this payload, and `readinessStore.ts:530` is its only caller in `src/`.

### DIFF
--- a/src/canvas/stores/__tests__/readinessStore.invalidation.spec.ts
+++ b/src/canvas/stores/__tests__/readinessStore.invalidation.spec.ts
@@ -649,7 +649,7 @@ describe('readinessStore — invalidation on mutation (ROADMAP 2.332)', () => {
 
     it('names every root the subscription watches', () => {
       expect([...WATCHED_ROOTS].sort()).toEqual(
-        ['ceeAnalysisReady', 'currentBriefText', 'edges', 'nodes'].sort(),
+        ['ceeAnalysisReady', 'currentBriefText', 'currentScenarioId', 'edges', 'nodes'].sort(),
       )
     })
   })
--- a/src/canvas/stores/__tests__/readinessStore.optionInterventions.spec.ts
+++ b/src/canvas/stores/__tests__/readinessStore.optionInterventions.spec.ts
@@ -47,7 +47,7 @@ function factorNode(id: string, label: string, data: Record<string, unknown> = {
 
 function wireNodes(nodes: Node[], edges: Edge[] = []): WireNode[] {
   const payload = JSON.parse(
-    buildReadinessPayload({ nodes, edges, ceeAnalysisReady: null, currentBriefText: null }),
+    buildReadinessPayload({ nodes, edges, ceeAnalysisReady: null, currentBriefText: null, currentScenarioId: null }),
   )
   return payload.graph.nodes as WireNode[]
 }
@@ -159,7 +159,7 @@ describe('buildReadinessPayload — option interventions reach the wire', () => 
     ]
 
     const payload = JSON.parse(
-      buildReadinessPayload({ nodes, edges, ceeAnalysisReady: null, currentBriefText: null }),
+      buildReadinessPayload({ nodes, edges, ceeAnalysisReady: null, currentBriefText: null, currentScenarioId: null }),
     )
 
     // Factor projection intact, field for field.
--- a/src/canvas/stores/__tests__/readinessStore.scenarioIdConvergence.spec.ts
+++ b/src/canvas/stores/__tests__/readinessStore.scenarioIdConvergence.spec.ts
@@ -1,0 +1,245 @@
+/**
+ * readinessStore — the readiness panel and the run path must assess the SAME
+ * model (ROADMAP: readiness/run convergence).
+ *
+ * THE MECHANISM, derived at CEE `cbc3ea3d` (`src/routes/assist.v1.graph-readiness.ts`)
+ * and at this UI tip, not inherited:
+ *
+ *   const GraphReadinessInput = z.object({
+ *     graph: Graph,
+ *     analysis_ready: AnalysisReadyPayload.optional(),
+ *     scenario_id: z.string().min(1).optional(),
+ *   });
+ *   ...
+ *   let assessedGraph: unknown = input.graph;
+ *   let assessedFrom = "request_graph";
+ *   if (input.scenario_id) {
+ *     const persisted = await loadPersistedScenarioStateStrict(input.scenario_id);
+ *     if (persisted.graph != null) { assessedGraph = persisted.graph; assessedFrom = "persisted"; }
+ *   }
+ *
+ * So CEE runs ONE predicate (`assessRouteAdmission` →
+ * `assessCanonicalAnalysisReadiness`, the same whole-model authority the TURN
+ * path uses) over one of TWO inputs, and the caller picks which by supplying
+ * `scenario_id` or not. The run path always reads the PERSISTED scenario. This
+ * projection has never sent `scenario_id`, so the panel has always been
+ * answered over `request_graph` — the browser's LOCAL copy. Same predicate,
+ * two inputs, and the divergence is exactly how the panel can read "3/3 ready"
+ * about a model the analysis turn never sees.
+ *
+ * ⚠ STATE-CLASS, and it is load-bearing rather than a nicety (CLAUDE.md
+ * fixture state-class rule). A guest, or a canvas drafted but not yet saved,
+ * has NO scenario id. For those callers `request_graph` is the CORRECT input —
+ * CEE's own comment says so ("the fallback is not a failure path"). Two things
+ * therefore have to hold at once, and they are tested separately below:
+ *   · a SAVED canvas sends the id, so both authorities read the persisted model;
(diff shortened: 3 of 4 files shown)


Talchain:
**Coupled CEE PR: Talchain/olumi-assistants-service#1049** — `fix/options-interventions-sync-at-persist`.

These two land together and neither merges alone.

**Why they are coupled.** CEE #1049 fixes the *content* defect: a user's intervention write lands on the option NODE, but the persist projection was **append-only**, so the stale top-level `options[]` entry survived — and `projectSemanticAnalysisReadyFromGraph` prefers `options[]` over the node (`topLevelById.get(id)` wins in **both** branches). The value was never lost; it was never propagated.

This PR fixes the *input* defect: without `scenario_id`, `assist.v1.graph-readiness.ts` sets `assessed_from = "request_graph"` **unconditionally** and assesses the browser's local copy. One predicate (`assessRouteAdmission`), two inputs.

**Order matters.** This PR alone makes the panel agree by reporting the value MISSING too — correct, but it reads as a regression until #1049 makes the persisted graph actually carry the value.

Two things worth confirming against the CEE contract, both derived at `cbc3ea3d`:
- `scenario_id: z.string().min(1).optional()` — the field **is** accepted, and `.min(1)` means an **empty string is rejected**. Omitting it for guest/unsaved canvases (rather than sending `""`) is required, not just tidier.
- The route's own comment confirms the fallback is a legitimate pre-save path, not a failure path: guest sessions correctly stay on `request_graph`.

CEE #1049 status: required check `Lint, TypeCheck, Unit Tests` = **success**; `Typecheck Drift (ratchet)` = **success**. `Security Audit` is red there, but it is red identically on merged PRs #1046/#1047/#1048 — a standing red on the PR trigger, contrast-controlled, not introduced by either of these changes.

PR #804, opened 2026-08-19
feat(conversation): render the model-building notices the user never sees

Talchain:
## What this closes

When Olumi drafts a model from a brief it drops things it could not represent, and it **records why** — `model_building_notices`. That field had **ZERO renderers repo-wide**: 56 generated, 0 delivered. A model that silently drops half a brief is indistinguishable, to the user, from one that captured it. Coupled producer PR: **CEE #1051** (wire-reachable, not mounted). This lane makes it mounted.

## Does a renderer already exist? — NO

Swept the fresh clone at `ad79b344` with `rg -a` **and a contrast control in the same run**:

| symbol | files |
|---|---|
| `model_building_notices` (target) | **3** — `src/v5/responseParser.ts` (quarantine list) + 2 parser specs |
| `analysis_state` (contrast control) | **20+** across product + specs |

The contrast fires, so the instrument sees. All three target hits are **parser-level tolerance only** — the field was validated and then dropped on the floor. No renderer, no store slice, no consumer.

## The field's real shape at the vendored 0.48.0

Read at the tarball bytes, not inferred:

```
{ total_count: number,               // .positive()
  groups: Array<{ kind, count }>,    // .min(1).max(6), strict
  details_redacted: true }           // z.literal(true)
```

`.strict()` at every level, a closed 6-member `kind` enum, and a `superRefine` enforcing **two** cross-field rules: kinds are unique, and `total_count` equals the group sum.

**`details_redacted` is a literal `true` — the wire carries aggregate counts per kind and NOTHING else.** No free text, no per-item detail, no next step. So every user-visible string is UI copy keyed to a closed enum, which is why the copy lives in exactly one place.

## What I built

Follows the established `_grounded_selection` sidecar pattern **exactly**: wire binding + copy authority → optional field on `ConversationMessage` → extraction on the live V5 path → **one ungated mount** in `MessageBubble`.

- `canvas/conversation/modelBuildingNotices.ts` — the canonical owner: extractor (parses through the **published schema itself**, so the producer's two cross-field rules are enforced rather than restated) and the sole kind→string map.
- `canvas/conversation/ModelBuildingNoticesNotice.tsx` — one count-led line, breakdown + next-route pointer behind a toggle.
- Mount in `MessageBubble.tsx`, assistant-only, conditioned on the payload's own presence.

### Why NOT inside `ModelReceiptBlock`

`ModelReceiptBlock` is the other post-draft card and looked like the natural home. It is gated **three times over**: the `preAnalysisEnriched` flag (**`defaultValue` false, and its envKey `VITE_FEATURE_PRE_ANALYSIS_ENRICHED` is ABSENT from `netlify.toml`**), a non-empty `coachingSummary`, and a normalised `ceeAnalysisReady`. Hanging an honest disclosure off three preconditions it does not need is exactly how a correct thing ships dark — the failure this lane exists to end. A test asserts the mount guard contains **no flag reference at all**.

## Product constraints

- **No raw wire vocabulary.** Kind codes never reach the DOM as text; an unnameable kind renders nothing rather than falling back to its code. Pinned across all six enum members plus a general snake_case ban.
- **Progressive disclosure.** Collapsed to one line by default.
- **No invented content.** `details_redacted` means no examples exist to show, so none are synthesised.
- **No minted zero — and no minted reassurance.** The contract *cannot encode zero* (`total_count` positive, `groups` min 1), so an absent field means **no attestation was supplied**, NOT "nothing was dropped". Absence renders nothing and claims nothing **in either direction**; a completeness reassurance built on a silent field would be the more dangerous fabrication.
- **`total_count` is the PRODUCER's.** Never derived from rendered groups — the two can legitimately differ, and they are pinned as separate quantities.
- **A useful next route, not a dead end.** One pointer line, **true by construction**: it names a conversational action available wherever the bubble renders, not a control this notice does not draw.

## RED-first (recorded at pristine, implementation stashed)

- `ModelBuildingNoticesNotice.spec.tsx` / `modelBuildingNotices.wire.spec.ts` — **collect failure**: `Failed to resolve import "../modelBuildingNotices"`.
- `modelBuildingNotices.mountSite.spec.ts` — 4 failed / 1 passed (the passing one is the scanner's own positive control): `expected -1 to be greater than -1` (no mount), `expected [] to deeply equal [ …(3) ]` (no importers).

## Mutants

Throwaway worktree at an absolute path outside the repo root, committed state only, applied-check scoped to `src/` asserting exactly 1, control asserting exactly 0, restore from a pristine **archive**. **Isolation proven by WRITING a sentinel** and asserting the source clone unchanged — source sha identical before/after, inodes differ (`681717507` vs `681894595`).

| # | mutation | verdict |
|---|---|---|
| M0 | control, no mutation | GREEN, applied-check 0 |
| M1 | delete the mount in `MessageBubble` | RED (8) |
| M2 | drop `!isUser` from the mount guard | RED (2) |
| M3 | mint a ZERO on absence | RED (2) |
| M4 | fall back to the raw kind code | RED (3) |
| M5 | always-plural summary copy | RED (2) |
| M6 | render unnameable kinds anyway | **survived → hole closed → RED (3)** |
| M7 | pre-expand the disclosure | RED (3) |
| M8 | drop a kind from the copy map | RED (5) |
| M9 | bypass the schema | RED (3) |
| M10 | headline count = row count, not producer total | RED (1) |
| M11 | cosmetic spacing (discriminating partner) | **GREEN, as required** |

**M6 survived and was a real finding, not an equivalent mutant.** The test *named* for the unnameable-kind drop carried a fixture of only nameable kinds, so it could never observe one — trap 19 inside my own suite. The branch is unreachable through the extractor by construction (closed enum), so it is now exercised at the shaper, the only entry point that reaches it. A **rot-check** confirms the new test pins its own precondition: making its fixture nameable REDs on a *different* assertion.

## Gates

- Typecheck gate: **PASSED** (3564/3604 files; ratchet 2263, baseline 2263 — unchanged).
- ESLint on all changed files: **0 errors** (12 pre-existing warnings in files I touched, none from these lines).
- Specs, collected **by name**: `ModelBuildingNoticesNotice.spec.tsx` **10**, `modelBuildingNotices.wire.spec.ts` **16**, `modelBuildingNotices.mountSite.spec.ts` **5** — **31 passed**.

## Rung, honestly

**TESTED.** Not deployed, not journey-witnessed. jsdom proves the notice is in the DOM under the real `MessageBubble` mount — it does **not** prove visibility or layout, and no such claim is made here. Deliberately not deployed: a SENDABLE integration batch is in flight.

### DIFF
--- a/src/canvas/conversation/MessageBubble.tsx
+++ b/src/canvas/conversation/MessageBubble.tsx
@@ -21,6 +21,7 @@ import { typography } from '../../styles/typography'
 import { safeRichText } from '../utils/safeRichText'
 import { AnswerBody, resolveAnswerBodyText } from './AnswerBody'
 import { GroundedOnNotice } from './GroundedOnNotice'
+import { ModelBuildingNoticesNotice } from './ModelBuildingNoticesNotice'
 import { InlineBlocks } from './InlineBlocks'
 import { AlertCircle, ChevronDown, ChevronUp, ListPlus, AlignLeft, RefreshCw } from 'lucide-react'
 import { FeedbackRow } from './FeedbackRow'
@@ -490,6 +491,17 @@ export const MessageBubble = memo(function MessageBubble({
         * one. */}
       {!isUser && message.groundedSelection && (
         <GroundedOnNotice groundedSelection={message.groundedSelection} />
+      )}
+      {/* What Olumi had to leave out of the model it drafted on THIS turn.
+        * NO FLAG (CEE emits `model_building_notices` unconditionally, and it is
+        * a declared 0.48.0 contract field), and the condition is the PAYLOAD'S
+        * OWN PRESENCE: a draft that dropped nothing carries no key, so
+        * `message.modelBuildingNotices` is undefined and nothing renders.
+        * Rendering a zero here would report an omission that never happened.
+        * Assistant-only: an omission is a fact about a MODEL OLUMI BUILT, and
+        * the user's own bubble never carries one. */}
+      {!isUser && message.modelBuildingNotices && (
+        <ModelBuildingNoticesNotice notices={message.modelBuildingNotices} />
       )}
       {message.insights && message.insights.length > 0 && isDeterministicCeeEnabled() && (
         <InsightsStrip insights={message.insights} onSendMessage={onArtefactMessage} />
--- a/src/canvas/conversation/MessageBubble.tsx
+++ b/src/canvas/conversation/MessageBubble.tsx
@@ -495,9 +495,11 @@ export const MessageBubble = memo(function MessageBubble({
       {/* What Olumi had to leave out of the model it drafted on THIS turn.
         * NO FLAG (CEE emits `model_building_notices` unconditionally, and it is
         * a declared 0.48.0 contract field), and the condition is the PAYLOAD'S
-        * OWN PRESENCE: a draft that dropped nothing carries no key, so
-        * `message.modelBuildingNotices` is undefined and nothing renders.
-        * Rendering a zero here would report an omission that never happened.
+        * OWN PRESENCE: with no attestation `message.modelBuildingNotices` is
+        * undefined and nothing renders. The contract cannot encode zero, so
+        * absence is SILENCE, not a clean bill of health — rendering a zero here
+        * would report an omission that never happened, and rendering
+        * reassurance would claim a completeness the field cannot evidence.
         * Assistant-only: an omission is a fact about a MODEL OLUMI BUILT, and
         * the user's own bubble never carries one. */}
       {!isUser && message.modelBuildingNotices && (
--- a/src/canvas/conversation/ModelBuildingNoticesNotice.tsx
+++ b/src/canvas/conversation/ModelBuildingNoticesNotice.tsx
@@ -1,0 +1,115 @@
+/**
+ * ModelBuildingNoticesNotice — the reader for CEE's `model_building_notices`.
+ *
+ * Until this component the field had ZERO renderers repo-wide: 56 notices
+ * generated, 0 delivered. Olumi was recording what it had to leave out of a
+ * user's model and telling nobody — which is indistinguishable, to the user,
+ * from a model that captured their brief completely.
+ *
+ * ── WHY THIS SURFACE ───────────────────────────────────────────────────────
+ * The same argument `GroundedOnNotice` makes, for the same reason: a
+ * model-building notice is a fact about ONE TURN's draft. It rides that turn's
+ * bubble so scrolling back shows each draft with its OWN omissions. A canvas or
+ * dock treatment is global and single-slot, so the latest draft's omissions
+ * would sit beside every earlier one — a mis-attribution the user cannot
+ * detect.
+ *
+ * It is deliberately NOT hosted inside `ModelReceiptBlock`, the other
+ * post-draft card. That card is gated THREE times over — the
+ * `preAnalysisEnriched` flag (envKey `VITE_FEATURE_PRE_ANALYSIS_ENRICHED`,
+ * `defaultValue` false and ABSENT from `netlify.toml`), a non-empty
+ * `coachingSummary`, and a normalised `ceeAnalysisReady`. Hanging an honest
+ * disclosure off three preconditions it does not need is how a correct thing
+ * ships dark, which is the failure this component exists to end. This one is
+ * UNGATED and its only condition is the payload's own presence.
+ *
+ * ── PROGRESSIVE DISCLOSURE ─────────────────────────────────────────────────
+ * Collapsed by default to ONE count-led line. A first-draft wall of notices is
+ * its own failure — the user has just been handed a model and the primary act
+ * is to read it, not to audit it. Depth is one click away and never pre-opened.
+ *
+ * ── WHAT IT NEVER CLAIMS ───────────────────────────────────────────────────
+ * · Never a kind code. `describeModelBuildingNoticeKind` is the only path to a
+ *   rendered string and returns null rather than falling back to the code.
+ * · Never that the listed rows account for the headline count. A kind this UI
+ *   cannot name is dropped from the rows while `total_count` stays the
+ *   producer's, so the two quantities can legitimately differ; the copy is
+ *   written so neither is stated in terms of the other.
+ * · Never a specific example. `details_redacted: true` means the producer sent
+ *   counts and nothing else — naming an item would be invention.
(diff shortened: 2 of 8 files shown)


Talchain:
## Post-push corrections (two), and the final gate state

**1. `Staging Gate` caught a real defect on the first push — reported rather than glossed.**

`Full Test Suite (shard 3/4)` failed on `tests/ci-guards/shell-conformance.spec.ts`:

```
NEW raw typography in the dock closure. DS v5 §2.4: panel scope uses only
panelHeader/panelBody/panelMeta — raw sizes, raw WEIGHTS, arbitrary sizes and
inline fontSize/fontWeight are all banned.
  src/canvas/conversation/ModelBuildingNoticesNotice.tsx: 0 -> 1
```

Mounting this component in `MessageBubble` put it **inside the dock closure**, which my local runs (the three lane specs + the whole `canvas/conversation` suite, 173 files / 2372 tests green) never covered — the guard lives in `tests/`, not `src/`. Fixed:

- `font-medium` → emphasis by **colour** (the row is `text-text-light`, so the darker body token carries the count without a raw weight);
- `mt-1.5` (6px) and `space-y-0.5` (2px) → `mt-2` / `space-y-1`, both off the 4px DS scale (`SHELL_SPACING_SCALE_PX`).

Verified by running **the guard that caught it**: `shell-conformance` 49/49 locally.

**2. Absence means less than my first comments claimed.** The contract **cannot encode zero** (`total_count` is `.positive()`, `groups` is `.min(1)`), so an absent field means *no notice attestation was supplied*, **not** "this draft dropped nothing". Behaviour was already correct; the stated rationale was too strong. Corrected, and it opened a second fabrication worth banning explicitly: a **completeness reassurance** built on a silent field is a positive claim with no evidence behind it, and is the more dangerous of the two. Now pinned alongside the minted-zero ban, plus a test that a zero-shaped payload is rejected by the schema rather than rendered.

## Final gate state at `25303159`

| check | result |
|---|---|
| **Staging Gate** (the sole required check) | **success** |
| Full Test Suite shards 1–4 + Summary | success |
| TypeScript + Lint · Typecheck Gate Self-Test | success |
| Production Build · Security Audit · CEE Contract Validation · Flag Deployment Drift | success |
| Visual Regression (advisory) | failure — red on the base branch and every open PR (3.78% measured noise floor against a 0.05% tolerance); **not evidence either way** |

`mergeable=MERGEABLE`. `mergeStateStatus=UNSTABLE` is the advisory red alone, not the required check.

**No file collision** with #802 / #803 / #791: this PR touches none of `canvas/stores/readinessStore.ts`, `FloatingOlumiPanel.placementCommit.spec.tsx`, `e2e/geometry/decisionNodeHittest.measure.ts`, `canvas/stores/guidanceStore.ts`, or `v5/buildPayload.ts`.

**Rung: TESTED.** Not deployed, not journey-witnessed — deliberately, with a SENDABLE batch in flight. jsdom proves presence in the DOM under the real mount; it does not prove visibility, and no such claim is made.

Talchain:
## Copy defect fixed at `5d3424a3` — `Staging Gate` green

### The fix

Took the one-line version, not the prose one. The count is now a **trailing quantity**, so it never has to agree with a category label that has fixed grammatical number:

```
[ONE]   COLLAPSED: Olumi left 1 thing from your brief out of this model
[ONE]   ROW      : Alternatives that were merged into a single option (1)

[MIXED] COLLAPSED: Olumi left 5 things from your brief out of this model
[MIXED] ROW      : Details you mentioned that aren't linked to anything else yet (1)
[MIXED] ROW      : Relationships you described that the model doesn't use (4)
```

Dumped from the **real `MessageBubble`** at this head, not inferred. Reads correctly at every count, no new data, compile-time exhaustiveness preserved.

### The guard gap that let it survive — closed

M5 pointed at `modelBuildingNoticesSummary` and bit cleanly, which made the kit *look* like it covered number agreement while every row went unguarded. Three row-level tests added, written against the **shape** rather than a phrasing, so a future copy edit that reintroduces a leading count REDs however the descriptions are worded:

| # | mutation | verdict |
|---|---|---|
| **M12** | **reintroduce the exact shipped defect (leading numeral)** | **RED (3)** |
| M13 | drop the count entirely — dodges agreement by losing information, and would pass a naive "no leading numeral" check | RED (3) |
| M14 | every row shows `rows[0]`'s count (identity binding, trap 19) | RED (1) |

Fresh worktree, isolation re-proven **by writing a sentinel** (source sha unchanged, inodes `682397150` vs `682673569`), applied-check exactly 1, control exactly 0, trailing tree clean.

### Limit 1 recorded in the spec's own header, as instructed

The mountSite "ungated" test now reads:

> ⚠⚠ **WHAT THE "UNGATED" TEST BELOW DOES NOT PROVE.** It reads the ONE LINE ABOVE the mount [...] It is **STRUCTURALLY INCAPABLE** of seeing an ANCESTOR gate: a flag on MessageBubble itself, on ChatMessage, on ChatThread, on ConversationPanel, or on any host above them would leave this test fully green while the notice never reached a user. **This test is not product-level reachability evidence, and must never be reported as such.**

with your hand-traced chain recorded beside it as a **dated manual derivation** this file does not re-derive — so a future component-tree change has to be re-walked by hand, because nothing here will fail.

The test's name changed too: `the mount is UNGATED…` → `the mount guard is LOCALLY flag-free (see header: says nothing about ancestors)`. The old name was the overclaim in miniature.

### Limit 2 noted, not touched

`MessageBubble.tsx:281` returning `null` for a block-less stock acknowledgement would swallow the notice. Pre-existing, narrow, out of this lane's scope — flagged, not fixed.

### Banked

`model_building_notices` is already in `QUARANTINABLE_ADDITIVE_KEYS` (`responseParser.ts:266`, pre-existing), so a malformed payload is lifted out and the turn survives with the notice silent — the UI fails softer than CEE, where the same field failing its contract fails egress for the whole turn.

### Gate at `5d3424a3`

`Staging Gate` **success**; all four Full Test Suite shards + Summary, TypeScript + Lint, Typecheck Gate Self-Test, Production Build, Security Audit, CEE Contract Validation, Flag Deployment Drift all success. Locally: `shell-conformance` 49/49, typecheck ratchet 2263 = baseline, lane specs **34** (`ModelBuildingNoticesNotice` 13 · `wire` 16 · `mountSite` 5).

`Visual Regression (advisory)` red — also red on staging tip `2b6ec553`, absent from the Gate's `needs`, and the sole cause of the UNSTABLE read. Not evidence.

**Not merged** — banked behind the SENDABLE freeze.

PR #807, opened 2026-08-20
fix(chips): map analysis_readiness so CEE's readiness recovery chip can render

Talchain:
**Paired with CEE #1059. Land both or neither — CEE #1059 alone is strictly worse than the bug it fixes.**

## Why

CEE's `analysis_not_ready` recovery chip is now typed `analysis_readiness` — the one `action_type` `route-v2.ts:2789` routes to the readiness arm without an LLM round-trip, so it **cannot** be demoted by the analysis-election gate.

Witnessed on the deployed build (fresh guest, 2026-08-20T00:00Z, UI `2b6ec553` / CEE `19a60fd`): as an **untyped** chip it reached CEE as ordinary chat, the LLM router elected `run_analysis`, and `analysis-election-gate.ts` demoted it — so the product's own **"Review the model"** CTA answered *"I have not run the analysis, because I did not read that as a request to run one."* The product refused its own affordance.

## ⚠ Without this, the CEE half makes things worse

`analysis_readiness` had **no `ACTION_TO_TURN_TYPE` mapping**, so it was absent from the derived `V5_ENABLED_ACTIONS`, and `SuggestedChips.tsx:194` filtered the chip out **entirely**. A chip that misroutes at least gives the user something to click; a chip that never renders gives them nothing. This is the "do not build half a capability" case, caught by tracing the chip hop by hop before landing CEE #1059.

## The fix is the mapping, not a filter bypass

`SuggestedChips`'s filter is **derived and correct** — do not patch it. Patching it would leave the click routing by accidental fall-through, because `dispatchAction` (`useConversation.ts:6524-6526`) reads the same map and its keyword-heuristic branch is `else if (!opts.action_type)`, which a typed chip skips. One line in `ACTION_TO_TURN_TYPE` self-repairs the whole chain by derivation:

`DISPATCHABLE_ACTION_TYPES` → `isHonestlyDispatchableAction` → `V5_ENABLED_ACTIONS` → chip renders → `dispatchAction` routes **deliberately** → `buildV5Payload` emits `source: 'chip_click'` + `chip.action_type: 'analysis_readiness'`.

`'conversation'` is the honest turn type: readiness coaching **is** a conversational answer, and it is what the UI's own `prepare_first_analysis` spark (`pre-analysis-v3/constants.ts:525-533`, the same prompt string) already gets.

## The wire was already ready — this PR does not touch it

- `analysis_readiness` is published in `ActionType` (vendored `@talchain/schemas` 0.48.0, `dist/boundary/enums.js:96-108`)
- it is in `CEE_ACCEPTED_ACTION_TYPES` (`buildPayload.ts:948`, CEE #578, deploy-confirmed 2026-07-20)
- a bound `action_type` promotes `source` `'chip'` → `'chip_click'` (`buildPayload.ts:159-160`)
- ingest already forwards it (`suggestedActionChips.ts:78-84`)

**Only the client's own dispatch vocabulary was short.**

## Evidence

- **RED-first at pristine**: 5 collected by name, 2 failed on named signatures (`DISPATCHABLE_ACTION_TYPES.has` and the turn-type mapping). **The wire assertion PASSED at pristine** — that is the positive control proving the gap was vocabulary, not plumbing.
- **Opposite-direction assertion**: `explain_from_structure` must STILL be refused, so the admission is a real discrimination and not a filter that now says yes to everything.
- `explainChips.vocabulary.spec.ts` **anticipated exactly this** (its own comment: *"Add the mapping and the chip lights up on its own; nobody has to remember to edit a list"*). Its excluded set narrows to `{explain_from_structure}` **by derivation**.
- `dispatchAction.spec.ts`'s hand-kept key mirror is updated **deliberately**, at source.
- **Typecheck ratchet clean at baseline** (2263 errors, +0). Two type errors in my own test were fixed at source, never by `--update-baseline`.
- `src/canvas/conversation` + `src/v5`: **281 files / 3928 passed, 0 failed.**

⚠ Suites were run with **real-shaped** `VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY`. Placeholder values silently shrank collection from **3949 → 3135** tests (trap 2b) while still reporting zero failures; every number above is from the healthy run.

## Not done

Not merged, and deployed staging was not driven (frozen for the Milestone A witness). No test yet covers a **producer-supplied** `analysis_readiness` chip driven through `SuggestedChips` to a real click — `sparkIntentContract.spec.ts` covers the UI-owned sparks, which bypass `SuggestedChips` entirely. Adding a producer-chip case to `proposeConfirmChipRender.spec.tsx` would close that blind spot.

### DIFF
--- a/src/canvas/conversation/__tests__/analysisReadinessChipRenders.spec.tsx
+++ b/src/canvas/conversation/__tests__/analysisReadinessChipRenders.spec.tsx
@@ -1,0 +1,106 @@
+/**
+ * ⭐ A CEE-EMITTED `analysis_readiness` CHIP MUST REACH THE USER.
+ *
+ * WITNESSED (deployed, fresh guest, 2026-08-20): a fresh user was refused an
+ * analysis and every offered route ended in a refusal. CEE's recovery chip
+ * carried no `action_type`, so it reached CEE as ordinary chat, the LLM router
+ * elected `run_analysis`, and the election gate DEMOTED it: "I have not run the
+ * analysis, because I did not read that as a request to run one." The product
+ * refused its own CTA.
+ *
+ * The CEE half of the fix types that chip `analysis_readiness` — the one
+ * `action_type` route-v2 routes to the readiness arm without an LLM round-trip,
+ * so it cannot be demoted.
+ *
+ * ⚠ BUT THE UI WOULD HAVE DELETED IT, AND THAT IS STRICTLY WORSE THAN THE
+ * MISROUTE. `analysis_readiness` has no `ACTION_TO_TURN_TYPE` mapping, so it is
+ * absent from the DERIVED `V5_ENABLED_ACTIONS`, and `SuggestedChips` filters out
+ * any chip whose bound `action_type` is not V5-known. A chip that misroutes at
+ * least gives the user something to click; a chip that never renders gives them
+ * nothing. Typing the chip in CEE without this mapping would have shipped half a
+ * capability.
+ *
+ * The wire was ALREADY ready — `analysis_readiness` is published in
+ * `ActionType`, present in `CEE_ACCEPTED_ACTION_TYPES` (`buildPayload.ts:948`),
+ * and a bound action_type promotes `source` to `chip_click`
+ * (`buildPayload.ts:159-160`). Only the client's own dispatch vocabulary was
+ * short — which is why the fix is a mapping, never a filter bypass:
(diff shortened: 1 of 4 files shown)


Talchain:
Paired with CEE https://github.com/Talchain/olumi-assistants-service/pull/1059 — **land both or neither.** This PR only makes the chip renderable; CEE #1059 is what emits it (typed) and what carries the six named readiness questions the refusal used to discard.

PR #809, opened 2026-08-20
feat(chips): render Run on the turn that offers it — gate on CEE's may_run

Talchain:
## The defect

The readiness answer loop's payoff turn says *"that's enough to run — two options have effects set, I'll leave the others out and say so"*. **At exactly that turn the Run chip was filtered out.**

The gate read `analysis_ready.status === 'ready'` (`SuggestedChips.tsx:196`). But `status` answers the **stricter** question *"is this model ready as it stands?"*. The payoff turn is `needs_user_input` **with an admitting run path**, because the run proceeds by excluding the options the user left open.

Measured in CEE on the real `live-4day-week` capture — **one status value carries both verdicts**:

| unconfigured options | `status` | `willProceed` |
|---|---|---|
| 1 | `needs_user_input` | **true** |
| 2 | `needs_user_input` | false |
| 3 | `needs_user_input` | false |

So no reading of `status` can recover admissibility. CEE now publishes its own verdict as `analysis_ready.may_run`.

## The change

```
admitted = status === 'ready' || may_run === true
```

A **strict widening**, deliberately a disjunction:

- The old term is untouched, so **everything that renders today still renders**.
- When CEE sends no `may_run`, behaviour is byte-for-byte today's — so **the two services may deploy in either order**.
- `=== true` is strict, so a malformed value can never widen the gate by accident.

`may_run` is `resolveRunAdmission(...).willProceed` — literally the boolean CEE's own run path throws `AnalysisNotReadyError` on — so **a chip shown on it cannot lead to a refusal**.

The DEV filter log mirrors the same predicate: a mirror reporting a chip as "removed for unreadiness" while the gate had admitted it would tell a different story from the code beside it.

## No contract change

`may_run` rides inside `analysis_ready`, which is `.passthrough()` in the pinned `@talchain/schemas` 0.48.0 — **no re-vendor, no schemas release**. Proven by executing the vendored schema; the same field at *top level* is rejected `unrecognized_keys`, which is why it rides inside.

## Evidence

**RED-first** (`SuggestedChips.mayRunGate.spec.tsx`, 7 tests): 1 failing at pristine — the payoff turn — and **6 passing**. The opposite-direction and fallback guards already held, which is what shows the gate was *widened*, not thrown open.

**Mutants** (throwaway worktree outside the repo root; isolation proven by writing a sentinel and asserting the source was byte-identical; distinct inodes; restore from a pristine archive, never `git checkout -- <path>`; applied-check scoped to `src/`, control exactly 0; trailing control green):

| mutant | expect | result |
|---|---|---|
| drop the `may_run` term | RED | RED — bit **only** the payoff test |
| drop the `status` term | RED | RED — bit both fallback / no-narrowing tests |
| `=== true` → `!== false` | RED | RED ×3 incl. the strict-true guard |
| always admit | RED | RED ×4 incl. opposite-direction |
| **unrelated fn (discriminator)** | **GREEN** | **GREEN 7/7** |

The last row is the discriminating pair: mutating a *different* function in the *same file* leaves the spec green, so the spec is bound to `admitsRunAffordance` by identity rather than merely sensitive to any edit.

**Gates:** `pnpm typecheck` PASSED (coverage + ratchet, 3570/3610 files). `pnpm lint` 0 errors; rules-of-hooks ratchet exactly matches baseline (229 known / 13 files) despite the added hook. All four `SuggestedChips` specs green together — **71 tests, 0 failures** — so the existing readiness-gate and aiPanelV2-polish specs are unregressed. My own spec's collected count asserted **by name**: 7.

## Land order

**CEE first, then this.** Both halves are safe alone: CEE's field is unread until this lands, and this falls back to today's behaviour until CEE ships. Cross-linked to the CEE PR below.

⚠ **Do not merge** — cross-repo, and a staging witness is running.

### DIFF
--- a/src/adapters/cee/types.ts
+++ b/src/adapters/cee/types.ts
@@ -443,6 +443,23 @@ export interface CEEAnalysisReady {
    */
   blocked_reason?: string
   /**
+   * CEE's OWN admission answer for this turn — will the analysis proceed if
+   * asked, right now? (cee `src/orchestrator/types.ts`,
+   * `GraphPatchBlockData.analysis_ready.may_run`, `resolveRunAdmission(...)
+   * .willProceed` — the boolean CEE's own run path throws on.)
+   *
+   * ⚠ NOT A RESTATEMENT OF `status`. `status` is the stricter *"is this model
+   * ready as it stands?"*; `may_run` is ALSO true when the run will proceed by
+   * excluding options the user left open. One `status` value
+   * (`needs_user_input`) carries BOTH verdicts, so `status` cannot be read to
+   * recover this — which is the whole reason the field exists.
+   *
+   * ABSENT means a pre-`may_run` CEE, never "no". Consumers must fall back to
+   * their existing behaviour when it is missing, so the two services can deploy
+   * in either order.
+   */
+  may_run?: boolean
+  /**
    * User-facing questions explaining issues when status is not 'ready'.
    * E.g., "The factor 'Price' doesn't have a path to the goal. Is this correct?"
    */
--- a/src/canvas/conversation/__tests__/SuggestedChips.mayRunGate.spec.tsx
+++ b/src/canvas/conversation/__tests__/SuggestedChips.mayRunGate.spec.tsx
@@ -1,0 +1,125 @@
+/**
+ * `analysis_ready.may_run` — the Run chip must render on the turn that offers it.
+ *
+ * ## THE DEFECT
+ *
+ * The readiness answer loop's payoff turn says *"that's enough to run — two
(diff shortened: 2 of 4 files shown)


Talchain:
Cross-repo pair: CEE producer is Talchain/olumi-assistants-service#1064. **Land order: CEE #1064 first, then this.** Both halves are safe alone — #1064's field is unread until this lands; this falls back to today's behaviour until #1064 ships. No `olumi-schemas` change is needed: the field rides inside `analysis_ready`, which is `.passthrough()` in the pinned 0.48.0.

PR #819, opened 2026-08-23
fix(results): no surface may state an influence number as a share of the outcome

Talchain:
## The measured failure

21 Aug fresh-guest journey witness: the T1 dominance nudge read

> **Dominant factor: <label> drives 100% of the outcome.**

**beside two other factors on the same screen reading Influence 75% and 68%.** Three shares of one outcome summing to **243%**.

Reproduced at pristine `54cc9cbb`, RED-first, verbatim:

```
AssertionError: expected 'Dominant factor:Cloud migration progr…' not to match /\d+\s*%\s*of the outcome/i
+ Received: "Dominant factor:Cloud migration progressdrives 100% of the outcome."
```

## What I am fixing: the CLAIM. Not the number, not the gate.

**No number this product holds is a share of the outcome**, so no adjustment to the number and no tightening of the gate can make that sentence true.

- `'normalised_elasticity'` is `|elasticity| / max|elasticity|`. The top factor is **1.0 by construction**, per set. Never a share.
- `'influence_score'` is the producer's **absolute causal influence SCORE** — this repo's own canonical wording (`INFLUENCE_EXPLANATION_ABSOLUTE`: *"an absolute causal influence score from the analysis"*). **An absolute SCALE is not a PARTITION.** Measured at the bytes on the repo's own golden staging capture `src/test/fixtures/golden-path-staging-2026-04-05.json`, the seven `influence_score` values read

  `1.0 / 0.8494 / 0.7304 / 0.6694 / 0.6562 / 0.3730 / 0.2238` → **sum 4.5022, i.e. 450% of one outcome.**

So on the very basis the old gate called *"honest only on the producer influence_score basis"*, the shares are impossible. The number survives and is honest; only the partition framing is removed.

## Second, real defect: trap 21 — two authorities, one sentence

The brief's derivation is **confirmed at `54cc9cbb`**:

```ts
const topInfluence = topDriver
  ? (topDriver.displayInfluence ?? topDriver.influenceScore ?? topDriver.normalisedInfluence ?? 0) : 0
const absoluteBasis = topDriver
  ? (topDriver.displayProvenance ? topDriver.displayProvenance === 'influence_score'
     : typeof topDriver.influenceScore === 'number') : false
```

`absoluteBasis` asks *"does an absolute producer score exist?"*; `topInfluence` — the number **printed** — reads `displayInfluence` **first**, which is set-relative by construction. On an unstamped payload (legacy fixture / cached payload) carrying both a set-relative `displayInfluence` of 1.0 and a modest `influenceScore`, **the gate passes on one field and the sentence prints the other.**

## Convergence — owners named, competitors REMOVED not wrapped

| Question | Canonical owner | Competitor removed |
|---|---|---|
| **Which number may back a basis-stamped influence claim, and on which basis** | `driverDisplayModel.resolveDriverClaimBasis` — returns value **and** the basis that produced it, from **one read** | the inline `displayInfluence ?? influenceScore ?? normalisedInfluence` / `displayProvenance ? … : typeof influenceScore` pair in `TriageActionCardsBody` (both deleted) |
| **How an influence number may be described in words** | `influenceScaleCopy.influenceMagnitude{Predicate,Title}` — already *"the ONE home for influence-scale disclosure wording"* | both `of the outcome` literals: `TriageActionCardsBody.tsx:433` and `TriageCard.tsx:483` (both deleted) |

Net: **26 deletions against 567 insertions**, of which 397 are the new spec and ~120 are the derivation written into the two owners' headers. The production logic in `TriageActionCardsBody` **shrank**.

`TriageCard` receives a bare `influence` prop and **never learns its provenance**, so it fails closed: the number, no basis claimed.

## Both directions

- **The honest case still fires.** Gate semantics are deliberately **unchanged**: absolute producer basis AND `>= 0.8` AND clear of the runner-up. A genuinely dominant factor still gets its warning, now reading:
  > *Dominant factor: Cloud migration progress has an influence score of 95%. If your assumptions about this factor are wrong, the result could change.*
- **The set-relative basis still gets no dominance claim at all** (it would fire on every such run).
- **The trap-21 case now fires at the producer number**, not the set-relative 1.0: an unstamped row with `influenceScore 0.93` / `displayInfluence 1` prints **93%**, and one with `influenceScore 0.6` prints **nothing**.

## Evidence

- **RED-first:** 16 collected by name, 15 failing at pristine; signature quoted above.
- **Mutants** (throwaway worktree at an absolute path outside the repo root, mutating committed state only; **isolation proven by WRITING a sentinel** — distinct inodes `694236266` vs `694651240`, source sha unchanged, sentinel absent from source; restored from a **pristine tar archive**, never `git checkout -- <path>`; applied-check scoped to `src/`, leading and trailing controls both **exactly 0**):

  | Mutant | Expect | Result |
  |---|---|---|
  | M1 revert the T1 share literal | RED | RED (2 failed) |
  | M2 revert the TriageCard share literal | RED | RED (1 failed) |
  | M3 reinstate the trap-21 value/basis split | RED | RED (3 failed) |
  | **M4a loosen for ALL objects** (nudge reads a different driver every run) | RED | **RED (3 failed)** |
  | **M4b loosen for a DIFFERENT object only** (runner-up clearance read) | GREEN | **GREEN (16 passed)** |
  | M5 drop the set-relative disclosure | RED | RED (1 failed) |
  | M6 widen the gate to any basis | RED | RED (1 failed) |

  **M4a/M4b is the discriminating pair (trap 19):** the assertions bind to the **top driver by identity**, not to whichever driver happens to satisfy a value predicate.
- **Gate:** `npx eslint` on all six changed files exit 0 · `bash scripts/ci/typecheck-gate.sh` **PASSED** (3596/3636 files, ratchet 2263 = baseline 2263).
- **Neighbours:** all 30 specs referencing `t1-dominant-nudge` / `driverDisplayModel` / `influenceScaleCopy` / `TriageCard` / `TriageActionCardsBody`, run in 3 shards — **1,475 passed, 0 failed** (includes `no-raw-influence-read.spec.ts`, 1,005 tests).
- Copy hygiene extended: the new builders are exercised across every basis, plus a new share-claim ban on the whole module.

## Sweep (with contrast control)

`rg -a "of the outcome" src/` — **target sites carrying a number: exactly two**, both fixed. Contrast control `"the outcome"`: **68 files**, so the sweep can see the family.

## Out of scope — reported, not touched (rule 13)

1. `src/canvas/components/pre-analysis/PreAnalysisPanel.tsx:1513` — *"`{label}` drives most of the outcome…"*. A **qualitative** claim carrying no number, in a file this lane does not own.
2. **`TriageActionCardsBody:169` feeds `influence={gap.voi}`** — value-of-information into a prop named `influence`. The tooltip is now honest about *an influence number*, but the **value** is not an influence at all. That is a different defect (wrong value into a correctly-named prop) and needs its own lane.

### DIFF
--- a/src/components/results/TriageActionCardsBody.tsx
+++ b/src/components/results/TriageActionCardsBody.tsx
@@ -37,7 +37,8 @@ import { openNodeInspector } from '@/canvas/nodes/shared/openNodeInspector'
 import type { ScientificEditorProps } from '@/components/shared/ScientificEditor'
 import { TargetProbabilityBars } from './TargetProbabilityBars'
 import { stripEncodingNotation, cleanFactorLabel } from './utils/cleanFactorLabel'
-import { INFLUENCE_TIE_EPSILON } from './driverDisplayModel'
+import { INFLUENCE_TIE_EPSILON, resolveDriverClaimBasis } from './driverDisplayModel'
+import { influenceMagnitudePredicate } from './influenceScaleCopy'
 // Canonical glossary check shared with the v17 hero row builders. Used in
 // v17 mode (`useV17Copy === true`) to sanitise user-supplied labels before
 // they are interpolated into GENERATED prose, aria-labels, or titles. The
@@ -390,26 +391,22 @@ function T1DominantNudge({
   const rankedDrivers = (drivers.topDrivers?.length ? drivers.topDrivers : drivers.drivers) ?? []
   const topDriver = rankedDrivers[0]
   const runnerUp = rankedDrivers[1]
-  // Lane 2 (policy): gate and phrase the nudge on the SAME display influence
-  // the panel/hero/graph/tornado show (driverDisplayModel via the stamped
-  // displayInfluence) — a raw-score read here claimed dominance the panel's
-  // own bars contradicted under partial producer coverage (Codex R3-B1 class).
-  const topInfluence = topDriver
-    ? (topDriver.displayInfluence ?? topDriver.influenceScore ?? topDriver.normalisedInfluence ?? 0)
-    : 0
-  // Review fold: "drives NN% of the outcome" is an ABSOLUTE causal claim —
-  // honest only on the producer influence_score basis. Under partial
-  // coverage the display value is set-relative (top driver ≡ 1.0 by
-  // construction), so the nudge would fire on EVERY such run claiming
-  // "100% of the outcome", contradicting the V17 dominance gate
-  // (UI-SEM-040, deliberately absolute) on the same screen. No provenance
-  // marker (legacy fixture) falls back to "a finite raw producer score
-  // exists" — the pre-fold absolute semantic.
-  const absoluteBasis = topDriver
-    ? (topDriver.displayProvenance
-        ? topDriver.displayProvenance === 'influence_score'
-        : typeof topDriver.influenceScore === 'number')
-    : false
+  // ⭐ ONE READ FOR THE VALUE AND ITS BASIS (2026-08-23, trap 21). This used
+  // to be TWO reads answering DIFFERENT questions: the value came from
+  // `displayInfluence ?? influenceScore ?? normalisedInfluence` while the gate
+  // separately asked "does an absolute producer score exist?". On an unstamped
+  // payload the gate passed on `influenceScore` while the sentence printed the
+  // SET-RELATIVE `displayInfluence` (top ≡ 1.0 by construction) — the
+  // witnessed "drives 100% of the outcome". `resolveDriverClaimBasis` is now
+  // the canonical owner of that question and returns value + basis together,
+  // so the two cannot disagree. Both former reads are DELETED, not wrapped.
+  const topBasis = resolveDriverClaimBasis(topDriver)
+  const topInfluence = topBasis?.value ?? 0
+  // Unchanged semantics: the dominance warning is licensed only on the
+  // producer's absolute basis. Under partial coverage the display value is
+  // set-relative and the top is 1.0 on EVERY such run, so a relative basis
+  // may not support a dominance claim at all.
+  const absoluteBasis = topBasis?.provenance === 'influence_score'
   // ⚠ DOMINANCE REQUIRES A RUNNER-UP TO BE DOMINANT OVER (2026-08-19). The
   // gate tested only the top's own magnitude, so on a run where three factors
   // tied at 100% it printed "Dominant factor: Cloud migration progress drives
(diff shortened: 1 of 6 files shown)


Talchain:
## Closing — SUPERSEDED. The defect is not live, and staging carries this architecture built better.

**Corrected premise:** triage read `types.ts:647` as a live gate. It is a **JSDoc comment**, not a surface.

**The literal is gone from staging.** `grep -rnaEI '%\s*of the outcome'` over `src/` at `5113f5a0` (3,463 files): **one hit, a deliberate mutant string** in `analysisCopyConsumerCensus.spec.ts:100`. Every `"of the outcome"` hit is a comment or a historic fixture. **Positive control fires:** the same regex at this PR's base `54cc9cbb` returns `TriageActionCardsBody.tsx:433 — const explanation = \`drives ${dominantPct}% of the outcome.\`` — the exact defect. Contrast `"the outcome"` = 146 files; negative control 0.

**Two lanes solved one finding independently.** `885adc63 fix(results): make analysis metric copy basis-safe`, merged in **#822**, landed the *same named function* `resolveDriverClaimBasis` in the *same file*.

**And staging's version is strictly stronger**, diffed side by side:

| | #819 | staging |
|---|---|---|
| stamped value/basis pair | takes both on trust | **cross-attests** `Object.is(displayInfluence, influenceScore)`; contradictory pair → `null`, no claim |
| copy families | 2 | **4**, incl. `value_of_information` — closing **#819's own out-of-scope item 2** |
| corpus | golden fixture | **live capture** `live-influence-score-one-2026-08-23.json` |
| regression guard | one render spec | render spec **+** `analysisCopyConsumerCensus` walking every `.tsx` under `src/`, allow-list asserted by `toEqual` (REDs if it grows **or** shrinks), with an attached self-test mutant |

Staging's spec has a test named *"withholds the mounted claim when the display value contradicts its stamped basis"* — **#819's own trap-21 case**, and it withholds where #819 would have printed the producer number. #819's out-of-scope item 1 (`PreAnalysisPanel` "drives most of the outcome") is also gone, covered by the census regex family.

**Cost:** 5 of 6 files conflict (measured by `git merge-tree`), **19** commits of drift — triage said 16. Only the new spec is conflict-free, and it is weaker than what is already there. Neither rebase nor re-cut is worth it.

### Recorded before closing — the one thing not on staging

#819's arithmetic proof that **neither basis can be a share**. Staging encodes the conclusion, not the proof. **Re-derived and correct:** the golden capture's seven `influence_score` values — `1.0 / 0.8494404883 / 0.7304170905 / 0.6693794507 / 0.6561546287 / 0.3730077993 / 0.2238046796` — **sum to 4.5022041371, i.e. 450% of one outcome.** Worth one comment line in `influenceScaleCopy.ts`.

**User-visible outcome: already delivered.** The nudge reads *"…has an influence score of 95%"* rather than three factors claiming 243% of one outcome.

Scope: source-derived at the staging tip. No deployed bundle reached.

PR #820, opened 2026-08-23
fix(canvas): floor the MOUNT fit at the legibility threshold — the restore arm's 8.56px text

Talchain:
WIP — banking early. RED-first commit only; fix follows.

The 21 Aug fresh-guest witness measured the RESTORE arm at zoom **0.4279** at 1280x800, giving **8.56px** text on 58 elements against a 10px canvas floor.

**Mechanism, derived at the installed bytes (`@xyflow/react` 12.10.2 / `@xyflow/system` 0.0.76):** `<ReactFlow fitView>` QUEUES a fit; `fitViewport` reads its floor as `options?.minZoom ?? minZoom`; with no `fitViewOptions` prop the floor falls through to the canvas instance's `minZoom={0.1}`. Every IMPERATIVE fit already carries `LABEL_LEGIBLE_ZOOM`; the DECLARATIVE mount fit did not.

### DIFF
--- a/src/canvas/ReactFlowGraph.tsx
+++ b/src/canvas/ReactFlowGraph.tsx
@@ -72,7 +72,7 @@ import { executeCanonicalRun } from './analysis/canonicalRunRegistry'
 import { HighlightLayer } from './highlight/HighlightLayer'
 import { computeFitPadding } from './utils/computeFitPadding'
 import { GHOST_OPTION_NODE_ID, excludeNonModelNodes } from './utils/fitTargets'
-import { LABEL_LEGIBLE_ZOOM } from './utils/zoomLegibility'
+import { LABEL_LEGIBLE_ZOOM, MOUNT_FIT_VIEW_OPTIONS } from './utils/zoomLegibility'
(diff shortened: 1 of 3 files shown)


Talchain:
Closing — the cause is being fixed instead of the symptom.

This branch floors the MOUNT fit at the legibility threshold, which leaves the product's own fit dead and makes the fallback merely survivable. The root cause is the latch ordering in `useFitViewOnLayoutVersion.ts`: the identity latch is claimed at `:189` **before** the `requestAnimationFrame` at `:190` that the effect's own cleanup cancels at `:193`. On restore, `nodes` changes between effect and frame by construction (React Flow measures freshly-mounted nodes), the frame is cancelled, the effect re-runs, and `:188` bounces it against the already-claimed identity — so the product's fit never runs and xyflow's mount `fitView` prop takes over, carrying neither `computeFitPadding`'s reservations nor `minZoom`.

One semantic question, one authority. This PR also carries a RED-first commit only, with the fix never following.

Its measurement is fully recorded in the body and is being used as the acceptance target for the cause fix: restore arm at 1280×800, zoom 0.4279 → 8.56px counter-scaled text on 58 elements against a 10px floor.

PR #824, opened 2026-08-24
D7 · The per-factor attribution ISL withheld is now said out loud — the absence, never a number

Talchain:
> # ⚠⚠ ROUND 2 — REVIEW FINDINGS APPLIED. Head is now `aac1faec`.
>
> An independent review returned **four findings**, and **the first two refuted this PR's own
> premise**. All four are re-derived here **at the producer's bytes**, not inherited from the
> review. The original body is preserved below the fold, unedited, with its refuted claims banded —
> it is the record of what was built and why, and rewriting it would falsify that record.
>
> **The guard is KEPT.** What changed is what it reads, what its corpus is made of, which direction
> it fails in, and — most of all — **what this PR is honestly claiming to be**.

---

## F1 · `suppressed_attributions` does **not** name the factors

This PR's module and copy both said it does (`attributionSuppression.ts:40`; `heroCopy.ts` "NO NAMES").
**It does not.** ISL's vocabulary is a closed **four-member set of attribution KINDS**, declared as
module constants at `src/models/response_v2.py:1437-1440` (ISL staging `28fe0c9`):

```python
SUPPRESSED_ATTR_FACTOR_SENSITIVITY   = "factor_sensitivity"
SUPPRESSED_ATTR_STABILITY_THRESHOLDS = "stability_thresholds"
SUPPRESSED_ATTR_CONDITIONAL_WINNERS  = "conditional_winners"
SUPPRESSED_ATTR_P_WIN_SENSITIVITY    = "p_win_sensitivity"
```

The field's own `.describe()` says the same (`:1481-1489`), and factor **node ids** live one field
over in `correlated_factors: List[str]` (`:1459-1462`). PLoT states it verbatim too
(`src/lib/driver-order.ts:351-352`: *"the members are ATTRIBUTION NAMES … not factor ids"*).

**Every positive fixture in the new spec used `['fac_price','fac_demand']`** — values the producer
never emits into this field. The whole corpus sat outside the producer's output domain
(trap 16-inverse: *a fixture you wrote yourself is not evidence about the wire*).

**Rebuilt** from the four literals, using the manifests ISL's three **independent** skip-site gates
actually build. The spec spells the literals out from the producer's bytes and deliberately does
**not** import the module's constant — otherwise a mutation would move both sides together and the
guard would agree with itself (trap 13b). Mutant **M2** proves it: changing the module constant
turns the suite RED.

## F2 · The reader asked the wrong question — and its equivalence was an accident nothing pinned

It returned `'suppressed'` on **any** non-empty manifest, then licensed a sentence specifically about
`p_win_sensitivity`. **All three sibling services discriminate by MEMBER:**

| service | site | predicate |
|---|---|---|
| ISL | `src/utils/response_builder.py:338-347` | `SUPPRESSED_ATTR_FACTOR_SENSITIVITY in …suppressed_attributions` |
| PLoT | `src/lib/driver-order.ts:510-511` | `islSuppressedAttributions?.includes('factor_sensitivity')` |
| CEE | `src/orchestrator-v5/coaching/uncertainty-priority.ts:193` | `suppressed.some((s) => s === 'p_win_sensitivity')` |

CEE asks the **identical** question on the **identical** field. Two services answering one question
two different ways is exactly the defect trap 21 describes.

**Why length ever looked equivalent — and why nothing here was entitled to rely on it.** ISL appends
at each skip site, so membership is a function of which phases were requested *and ran*
(`robustness_analyzer_v2.py`):

- `:2351-2360` — `factor_sensitivity` + `stability_thresholds`, iff `has_uncertainties()` **and** `"sensitivity" in analysis_types`
- `:2373-2375` — `conditional_winners`, iff `has_uncertainties()` **and** `len(request.options) > 1`
- `:2585-2587` — `p_win_sensitivity`, iff **`request.include_voi`** **and** `has_uncertainties()`

Those gates are **independent**. The length reader only matched the member reader because PLoT
**hardcodes `include_voi: true`** (`src/integrations/isl/translator-v3.ts:967`), and
`src/types/engine-v3.ts:707-714` records that an inbound `include_voi` is **ignored**. Nothing in
this PR cited or pinned that.

**Fixed** to `.some(m => m === 'p_win_sensitivity')` — exact identity, no substring, no case folding.
**And the dependence that has been removed is now pinned by a discriminating twin**
(`does NOT depend on PLoT hardcoding include_voi: true`), whose two manifests differ *only* by the
token under test — so it is a discrimination, not two readings that happen to differ. The day PLoT
makes `include_voi` request-gated, that test is what fails loud instead of the product quietly
telling users their win-probability attribution was withheld on runs that never asked for it.

## F3 · ⭐ The honest reachability sentence — the motivating claim was false, and is withdrawn

This PR argued the discriminator *"sat in the browser while the user was told nothing"*. **It never
sat there, and I am withdrawing that claim rather than defending it.**

> **This is a FAIL-CLOSED GUARD for a state the current producer cannot reach — not the repair of a
> defect any user has hit.**

`correlation_model` is emitted by ISL **only** when the request supplied `factor_correlations`:
`_build_correlation_plan` returns `None` with no correlations
(`robustness_analyzer_v2.py:2903-2907`) and `_build_correlation_disclosure` returns `None` with no
plan (`:2930`). PLoT forwards that key **only** from the caller's request body
(`src/routes/v2/run.ts:7162`), request-gated inside the translator (`translator-v3.ts:1032-1034`).
**And nothing upstream populates it.**

**Scope of that claim, stated exactly (trap 20 — a row minted from a sweep must carry the sweep's
scope, never its generalisation):**

> **A STATIC SWEEP of all four repos at their staging tips. NO LIVE WIRE CAPTURE WAS TAKEN.**
> `git grep -a -i -I` over every tracked file in fresh blobless clones, with **contrast controls in
> the same run** (trap 13e — a target-zero is only evidence when a same-family symbol reads non-zero
> beside it):

| repo | tip (date) | tracked | `factor_correlations` | contrast `factor_evppi` | contrast `p_win_sensitivity` |
|---|---|---:|---:|---:|---:|
| **CEE** | `77e2e7d9` (24 Aug) | 4,291 | **0 files / 0 lines** | 24 / 124 | 9 / 42 |
| **UI** | `363d3fac` (24 Aug) | 4,473 | **0 files / 0 lines** | 35 / 107 | 24 / 48 |
| PLoT | `7e5d8a7d` (20 Aug) | 7,227 | 15 / 56 | 30 / 102 | 29 / 82 |
| ISL | `28fe0c95` (9 Aug) | 555 | 13 / 48 | 18 / 167 | 21 / 125 |

The PLoT and ISL hits are the **receiving** side, classified file by file: PLoT's are one OpenAPI
contract, six `src/` sites that are type declarations, the verbatim forward, canonicalisation and a
contract-key list, and eight test/fixture files — **no populator**. ISL's are the request model, its
validator, the analyzer's consumption, `openapi.json`, tests and one acceptance-evidence note —
**no populator**. The only services that could populate the key on a request are **CEE and the UI,
and both read zero.**

**The precise claim is *"no code path in any of the four repos populates `factor_correlations`"* —
NOT *"a capture was taken and none was seen"*.** The chain is complete and **unfed**.

**Why keep the guard.** The chain is live, typed and one caller-supplied key away from firing; when
it does fire, the alternative is the Drivers list silently reading as the complete per-factor
picture. A guard that costs nothing while dark and tells the truth when lit is worth keeping — it
just must not be *sold* as a defect repair. Every prose site that implied otherwise is corrected in
this commit (`types.ts`, `useResultsSectionData.ts`, `HeroEvidenceDisclosure.tsx`).

## F4 · One branch failed **open**

`mapV5AnalysisToReport.ts:1302` dropped a non-array `p_win_sensitivity`, so **present-but-malformed
arrived indistinguishable from absent — and read as suppression.**

The asymmetry with its siblings is the whole point. For `factor_evppi`, dropping a malformed value
costs **silence**. For `p_win_sensitivity`, **absence *is* the contract's suppression signal**
(*"absent from the response, not null"*), so narrowing **manufactures** that signal and the product
tells a user something was withheld when it arrived and was merely unreadable.

**Fixed:** the mapper carries the key **verbatim whenever the producer sent one** (typed `unknown`,
not `unknown[]`), absence still preserved; the single reader treats **any** arrived value — array,
object, string, number, `null` — as *not suppressed*. That is the fail-**closed** direction for this
field.

## ⚠ A limit of the ratified sentence, recorded rather than hidden

Whenever this notice can fire, ISL has **also** suppressed `factor_sensitivity` and
`stability_thresholds` at the same gate (`:2351-2360`) — the members travel together under active
correlation. *"The ranking above is unaffected"* survives **only** because PLoT's **graph-primary**
path re-derives driver order independently (`driver-order.ts:504-520`). On the **ISL-only fallback**
PLoT itself sets `basis: 'none'` for exactly this reason (`:510-514`) — so there the clause is
**vacuous** rather than false, and the deck renders no drivers to host it. Written into `heroCopy.ts`
so that a future basis consuming ISL's per-factor attribution directly makes it false *loudly*.

## RED-first (at the pristine branch head `363d3fac`)

```
13 failed | 31 passed (44 collected)
```

- 6 × `TWIN: …` — a non-empty manifest that does **not** name `p_win_sensitivity` lighting the notice
- 4 × `FAIL-CLOSED: p_win_sensitivity arrived as null / object / string / number`
- `does NOT depend on PLoT hardcoding include_voi: true (the twin that pins it)`
- 2 × mapper: `a malformed p_win_sensitivity is CARRIED` / `a null p_win_sensitivity is CARRIED`

GREEN **44/44** after. Both spec files asserted collected **by name**, never inferred from a total
(trap 2b).

## Mutants — 8, each biting a **distinct named** set

Throwaway worktree **outside** the repo root, detached at the commit SHA, isolation proven by a
**sentinel WRITE** (source SHA-256 unchanged after writing into the worktree) plus distinct inodes;
restores are `git checkout HEAD --` (trap 9h); applied-check scoped to `src/`, **control exactly 0**,
every mutant exactly 1; collected asserted at 44 every run; clean-tree checked before *and* after
each; trailing control clean + GREEN.

| # | Mutation | Result |
|---|---|---|
| M0 | *(control — none)* | applied **0**, **GREEN** |
| M1 | reader reverts to **LENGTH**-based | **RED** — all 6 twins + the `include_voi` pin (7) |
| M2 | constant → `'factor_sensitivity'` | **RED** — 6, incl. 2 positives |
| M3 | constant → `'conditional_winners'` | **RED** — 7, a **different** set from M2 |
| M4 | exact match → **substring** match | **RED** — exactly 1: the `p_win_sensitivity_v2` near-miss twin |
| M5 | exact match → **case-insensitive** | **RED** — exactly 1: the `P_WIN_SENSITIVITY` twin |
| M6 | reader guard → `Array.isArray` only | **RED** — exactly the 4 FAIL-CLOSED shapes |
| M7 | mapper re-narrows to arrays | **RED** — exactly the 2 mapper cases |
| M8 | a **digit** enters the ratified sentence | **RED** — exactly the no-magnitude assertion |
| **M9** | **non-uniformity control** — sibling `decisionVoi.ts` mutated | **GREEN 44/44** |

**M2 and M3 are the discriminating pair:** both flip the positives, but they RED **different**
twins — proving the tests bind to *this* named token, not merely to "some membership test"
(trap 19). **M4 and M5 each RED exactly one test**, the twin written for them. **M9 is there because
a probe that answers the same for every input is reporting on itself** (trap 20).

## Gate

`Staging Gate` **success** at `aac1faec` — polled in the foreground to conclusion, filtering
`status == "completed"` **before** `conclusion` (trap 24b); **0 running, 0 required failures**.
Locally, the required check's **full step sequence** (trap 22e): `pnpm run lint` **0 errors** (hooks
ratchet exactly at baseline) → `pnpm typecheck` **PASSED** with the ratchet **unchanged at 2252**
(no identity churn) → scoped regression `src/components/results` + `src/v5`
**347 files / 5,913 tests passed**.

⚠ `Visual Regression (advisory)` is red, and **its verdict is not evidence about this change in
either direction** — its own harness self-test reports a noise floor of **5.5266%** against a
`maxDiffPixelRatio` of **0.0005**, alongside *"COMPLETENESS GUARD FAILED: ZERO screenshots were
captured"*. It is non-discriminating, not merely noisy.

---

<details>
<summary><b>Original body (round 1) — preserved unedited. ⚠ Its §"What a user now sees" motivating
claim (F3) and its "names the factors" premise (F1) are REFUTED above; its mutant table describes
the round-1 kit, superseded by the 9 above.</b></summary>

## Capability (domain 7 — Science-to-Reasoning)

`enrichment.p_win_sensitivity` (ISL's per-factor percentage-point-of-win deltas) and
`enrichment.correlation_model` were **parsed, stored and read by nothing**. This makes the
producer's *suppression verdict* reach the user — as an honest absence, not a magnitude.

## The prior claim I was sent to build on is REFUTED

The 7 Aug audit said decision-level EVPI "is parsed and stored and has NO rendered consumer."
**False at the current tip.** `voi/decisionVoi.ts` classifies it and
`HeroEvidenceDisclosure.tsx:536` renders it at `data-testid="hero-resolve-next-decision-level"`.
I did not force the brief onto a closed gap; I pivoted to the residual the mapper itself names
("the other three are transported and unread").

**Sweep scope, stated exactly.** UI staging tip `88cb7e3728aff1e18a2d32635b0ac24cab1c348b`, fresh
blobless clone, `rg -a` case-insensitive over `src/` + `tests/`, excluding fixtures / `__tests__` /
`vendor`. Target `p_win_sensitivity` + `correlation_model`: hits **only** the mapper's parse+store
lines (`mapV5AnalysisToReport.ts:1302-1306,1497-1498`). **Contrast control in the same run** —
`factor_evppi`, the sibling that *is* displayed — **30+ consumer hits**. Target zero + contrast
non-zero ⇒ real absence, not a blind probe.

## What a user now sees that they could not before

On a run where ISL suppressed the attribution, the **Drivers** view now carries, directly under the
"Ranked by effect on the analysed outcome" note:

> How much each factor moves an option's chance of coming out ahead wasn't produced for this run —
> this analysis models these factors as moving together, so their separate contributions can't be
> told apart. The ranking above is unaffected.

Before: the drivers list read as the complete per-factor picture, and the reason one per-factor
question went unanswered sat in the browser unread. The contract calls exactly this the
"two-states-one-byte defect"; it was arriving from the consumer side.

## Why an absence and not a number

The producer's own `.describe()` (`@talchain/schemas` 0.48.0, `dist/boundary/enrichment.js:1027-1043`):
*"absence here is a SUPPRESSION VERDICT, not a missing convenience. Transport only — pp display is
barred by PP_TOKEN doctrine."* So the reader returns a **closed two-member string enum** — no digit
from the suppressed payload has a path to the DOM. No factor **names** (shape is typed OPEN; the
neighbouring register already bans id-shaped names) and no **count** (a count is a magnitude about a
suppressed quantity). Fails **closed**: any unreadable shape ⇒ silence.

## Red-first signature

At pristine, with the reader present so the render assertion failed on its own merits:

```
attributionSuppression.disclosure.spec.tsx > Drivers view — the suppression verdict is disclosed
  > renders the withholding notice when the producer attests suppression
  → Unable to find an element by: [data-testid="hero-evidence-attribution-suppressed"]
1 failed | 14 passed (15)
```

## Mutants

Isolated worktree **outside** the repo root, detached at the commit SHA. Isolation proven by a
**write sentinel with its positive control asserted** (distinct inodes; my first probe printed a
false PASS because the write silently failed — caught and redone). Control applied-files = 0;
every mutant applied-files = 1; 15 collected every run (no collection failure scored as a kill);
tree clean throughout; restores are `git checkout HEAD --`.

| Mutant | Result |
|---|---|
| M1 guard forced FALSE (reverts the change) | **RED** — `renders the withholding notice when the producer attests suppression` |
| M2 guard forced TRUE (renders always) | **RED** — `renders NOTHING when the producer attested no suppression` |
| M3 reader fails open on empty `suppressed_attributions` | **RED** — `suppressed_attributions EMPTY — named nothing, so suppressed nothing` |
| M4 reader ignores an arrived `p_win_sensitivity` | **RED** — both arrived-array cases |
| **DECOY** rename a *different* object's testid on the same surface | **GREEN 15/15** — bound by identity, not proximity |

M1 and M2 fail on **different** assertions, which is what makes the both-direction pair evidence
rather than one guard twice.

## ⚠ CI caught a gate defect my local run did not — and the fix was NOT the one the gate suggested

The first push went RED on `Typecheck Gate Self-Test`, which failed scenario 1 — the **negative
control** asserting that *on a clean tree the gate prints no added-diagnostics notice*. Its own
message: *"Treat every 'typecheck clean' claim on this branch as UNPROVEN until it passes."*
`Staging Gate` failed **only as a consequence** (tsc, vitest, vitest-summary, build all success).

**Cause, measured.** Adding a REQUIRED member to `ResultsSectionDataReturn` re-words the
pre-existing baselined TS2739 in six fixtures (its elided-property count moves), and those five
messages render **non-deterministically** — proven by regenerating the identity baseline **twice
over one unchanged tree** and diffing: **10 rows differed across exactly those files**. That is the
union-member ordering hazard the gate documents in its own header.

**So regenerating the baseline was the wrong fix, and I reverted the regeneration I had already
made.** A non-deterministic identity cannot be baselined; banking it would have re-broken on the
next run — a hand-maintained mirror with a fuse. The six fixture edits are reverted to pristine too,
since the source object's own rendering is part of the message.

**The actual fix** keeps the field REQUIRED on `HeroEvidenceModel` — where the silent-surface
argument bites and where the compiler named four constructors — and makes it OPTIONAL on the hook
interface, whose only producer is the hook itself. `buildHeroModel` already coalesces fail-closed.

Result: gate passes **silently**, 556 files / 2252 errors — **identical to pristine**; self-test
**18 passed / 0 failed**.

## Gates

- `pnpm typecheck` (`scripts/ci/typecheck-gate.sh`) **PASSED**. The required field named 10
  constructors; **all fixed at source — no `--update-baseline`.** The 6 residual TS2739s were
  already at 1 error each in the pristine baseline (verified); only their message identity shifted.
- `npx eslint` on changed files: **0 errors** (2 pre-existing `exhaustive-deps` warnings at
  unrelated lines).
- `Typecheck Gate Self-Test`: **18 passed / 0 failed**. `Staging Gate`: **success**.
- CI on `363d3fac`: all 4 test shards, TypeScript + Lint, Production Build, Security Audit, CEE
  Contract Validation, Staging Gate — **success**. The single red is **Visual Regression
  (advisory)**, which is a **pre-existing standing red proven by control**: the pristine staging tip
  `88cb7e37` fails it with **byte-identical counts (13 failed / 21 passed)**, in
  `e2e/visual/states.visual.spec.ts` founder states (fresh draft, blocked/provisional, Model tab,
  conversation tab, graph zoom) — none of which is the Analysis-tab surface I touched.
- Affected suites: **187 files / 3328 tests passed**, including
  `no-evpi-display.contract.test.ts`, both `evpiSurfacesRemoved` honesty specs and the shared
  refuted-EVPI matchers. My own spec collected **15 tests by name** (asserted, not inferred).

## What I do NOT claim

- **Not journey-witnessed, not wire-witnessed.** No live payload in this repo carries
  `correlation_model` — the render path is proven in jsdom only. Whether staging currently returns
  runs under active correlation is **unknown to me** and untested here.
- **Not a visibility claim.** jsdom cannot prove layout; nothing asserts one.
- No claim that the suppressed factors have no effect — the estimator declined to compute the
  attribution and so established nothing of the kind.
- No claim about *which* factors were suppressed, or how many.
- No claim about decision-level EVPI beyond refuting the audit line: I did not change that surface.
- The disclosure self-hides when there is nothing to disclose, so a suppression on a run with **no
  drivers** has no host. Deliberate, not covered.

</details>

### DIFF
--- a/src/__fixtures__/resultsPanelV7.minimal.hook.ts
+++ b/src/__fixtures__/resultsPanelV7.minimal.hook.ts
@@ -87,7 +87,6 @@ export const minimalFixture: ResultsSectionDataReturn = {
   // renders nothing rather than a placeholder. REQUIRED on the hook type
   // precisely so the compiler names every constructor.
   decisionVoi: 'not_computed',
-  attributionSuppression: 'not_attested',
   assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
 }
 
--- a/src/__fixtures__/resultsPanelV7.minimal.hook.ts
+++ b/src/__fixtures__/resultsPanelV7.minimal.hook.ts
@@ -87,6 +87,7 @@ export const minimalFixture: ResultsSectionDataReturn = {
   // renders nothing rather than a placeholder. REQUIRED on the hook type
   // precisely so the compiler names every constructor.
   decisionVoi: 'not_computed',
+  attributionSuppression: 'not_attested',
   assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
 }
 
--- a/src/__fixtures__/resultsPanelV7.normalised.hook.ts
+++ b/src/__fixtures__/resultsPanelV7.normalised.hook.ts
@@ -187,6 +187,7 @@ export const normalisedFixture: ResultsSectionDataReturn = {
   // renders nothing rather than a placeholder. REQUIRED on the hook type
   // precisely so the compiler names every constructor.
   decisionVoi: 'not_computed',
+  attributionSuppression: 'not_attested',
   assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
 }
 
--- a/src/__fixtures__/resultsPanelV7.normalised.hook.ts
+++ b/src/__fixtures__/resultsPanelV7.normalised.hook.ts
@@ -187,7 +187,6 @@ export const normalisedFixture: ResultsSectionDataReturn = {
   // renders nothing rather than a placeholder. REQUIRED on the hook type
   // precisely so the compiler names every constructor.
   decisionVoi: 'not_computed',
-  attributionSuppression: 'not_attested',
   assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
 }
 
--- a/src/__fixtures__/resultsPanelV7.rich.hook.ts
+++ b/src/__fixtures__/resultsPanelV7.rich.hook.ts
@@ -199,6 +199,7 @@ export const richFixture: ResultsSectionDataReturn = {
   // renders nothing rather than a placeholder. REQUIRED on the hook type
   // precisely so the compiler names every constructor.
   decisionVoi: 'not_computed',
+  attributionSuppression: 'not_attested',
   assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
 }
 
--- a/src/__fixtures__/resultsPanelV7.rich.hook.ts
+++ b/src/__fixtures__/resultsPanelV7.rich.hook.ts
@@ -199,7 +199,6 @@ export const richFixture: ResultsSectionDataReturn = {
   // renders nothing rather than a placeholder. REQUIRED on the hook type
   // precisely so the compiler names every constructor.
   decisionVoi: 'not_computed',
-  attributionSuppression: 'not_attested',
   assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
 }
 
--- a/src/components/results/__tests__/buildResultsVM.spec.ts
+++ b/src/components/results/__tests__/buildResultsVM.spec.ts
@@ -133,7 +133,6 @@ function makeData(overrides: {
     // renders nothing rather than a placeholder. REQUIRED on the hook type
     // precisely so the compiler names every constructor.
     decisionVoi: 'not_computed',
-    attributionSuppression: 'not_attested',
     assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
   }
 }
--- a/src/components/results/__tests__/buildResultsVM.spec.ts
+++ b/src/components/results/__tests__/buildResultsVM.spec.ts
@@ -133,6 +133,7 @@ function makeData(overrides: {
     // renders nothing rather than a placeholder. REQUIRED on the hook type
     // precisely so the compiler names every constructor.
     decisionVoi: 'not_computed',
+    attributionSuppression: 'not_attested',
     assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
   }
 }
--- a/src/components/results/__tests__/golden-payload.spec.ts
+++ b/src/components/results/__tests__/golden-payload.spec.ts
@@ -212,6 +212,7 @@ function makeGoldenData(): ResultsSectionDataReturn {
     // renders nothing rather than a placeholder. REQUIRED on the hook type
     // precisely so the compiler names every constructor.
     decisionVoi: 'not_computed',
+    attributionSuppression: 'not_attested',
     assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
   }
 }
--- a/src/components/results/__tests__/golden-payload.spec.ts
+++ b/src/components/results/__tests__/golden-payload.spec.ts
@@ -212,7 +212,6 @@ function makeGoldenData(): ResultsSectionDataReturn {
     // renders nothing rather than a placeholder. REQUIRED on the hook type
     // precisely so the compiler names every constructor.
     decisionVoi: 'not_computed',
-    attributionSuppression: 'not_attested',
     assumedStrength: { selected: null, refusalReason: 'no_robustness_data', assumedFragileCount: 0 },
   }
 }
--- a/src/components/results/analysis-hero/HeroEvidenceDisclosure.tsx
+++ b/src/components/results/analysis-hero/HeroEvidenceDisclosure.tsx
@@ -374,10 +374,17 @@ export function HeroEvidenceDisclosure({
                 CHANCE OF COMING OUT AHEAD (`p_win_sensitivity`) and, under
                 active correlation, SUPPRESSES it — "absent from the response,
                 not null" — naming what it withheld in
-                `correlation_model.suppressed_attributions`. Both halves have
-                been transported since the VOI family landed and read by
-                nothing, which is the contract's own two-states-one-byte defect
-                arriving from the consumer side.
+                `correlation_model.suppressed_attributions` — an ATTRIBUTION-KIND
+                manifest, not a list of factors. Both keys have been CARRIED by
+                the mapper since the VOI family landed and read by nothing.
+
+                ⚠ SCOPE, STATED HONESTLY: this is a FAIL-CLOSED GUARD, not the
+                repair of a defect a user has seen. `correlation_model` is
+                emitted only when the request supplied `factor_correlations`,
+                and no code path in CEE or the UI populates that key — so the
+                chain is complete but UNFED at the current tips. The notice is
+                here so that the day a caller does supply correlations, the
+                withholding is said out loud instead of arriving silently.
 
                 IT SITS DIRECTLY UNDER `driversNote` BY DESIGN. That note is
                 what tells the reader what the ranking IS; this is what tells
--- a/src/components/results/analysis-hero/HeroEvidenceDisclosure.tsx
+++ b/src/components/results/analysis-hero/HeroEvidenceDisclosure.tsx
@@ -366,6 +366,45 @@ export function HeroEvidenceDisclosure({
               <p className={`${typography.panelMeta} text-text-light`}>
                 {HERO_COPY.evidence.driversNote}
               </p>
+              {/*
+                THE WITHHELD PER-FACTOR ANSWER, SAID OUT LOUD.
+
+                The list below is ranked by effect on the ANALYSED OUTCOME.
+                ISL separately computes each factor's effect on an option's
+                CHANCE OF COMING OUT AHEAD (`p_win_sensitivity`) and, under
+                active correlation, SUPPRESSES it — "absent from the response,
+                not null" — naming what it withheld in
+                `correlation_model.suppressed_attributions`. Both halves have
+                been transported since the VOI family landed and read by
+                nothing, which is the contract's own two-states-one-byte defect
+                arriving from the consumer side.
+
+                IT SITS DIRECTLY UNDER `driversNote` BY DESIGN. That note is
+                what tells the reader what the ranking IS; this is what tells
+                them which per-factor question it does NOT answer. Separating
+                them would leave the caveat below a list the reader has already
+                finished interpreting.
+
+                WHAT DECIDES IT lives in `voi/attributionSuppression.ts`; the
+                sentence and its licence live in `heroCopy.ts`. Nothing is
+                decided here: this block renders ONE ratified string on ONE
+                member of a closed two-member enum, and `not_attested` renders
+                nothing at all rather than a placeholder — because an
+                unreadable disclosure is not a withholding claim.
+
+                NO MAGNITUDE, and not by restraint: the producer's own contract
+                text says "pp display is barred by PP_TOKEN doctrine". No value
+                from the suppressed payload is in scope at this site.
+              */}
+              {evidence.attributionSuppression === 'suppressed' && (
+                <p
+                  className={`${typography.panelMeta} flex items-start gap-1.5 text-text-light`}
+                  data-testid="hero-evidence-attribution-suppressed"
+                >
+                  <Info aria-hidden="true" className="mt-0.5 h-3 w-3 flex-none text-info" />
+                  <span>{HERO_COPY.evidence.attributionSuppressed}</span>
+                </p>
+              )}
               {visibleDrivers.map((d) => {
                 const key = `driver-${d.rank}-${d.label}`
                 const grid = 'grid w-full grid-cols-[minmax(0,1fr)_5rem] items-center gap-2'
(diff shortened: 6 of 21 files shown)


Talchain:
## HOLD — not a fix rejection, and not for the reason first given

The disposition that reached me was *"do not merge; the upstream constraint representation/evaluation issue is the owner."* **That premise is wrong and I am recording the correction:** this PR is about ISL's suppressed `factor_correlations`, not about hard constraints. It patches no constraint consumer. Whoever inherits this should not look for a constraint dependency here, because there isn't one.

**The conclusion survives anyway, for a different and better reason — one this lane found and disclosed itself:**

> **This is a FAIL-CLOSED GUARD for a state the current producer cannot reach — not the repair of a defect any user has hit.**

Backed by a four-repo static sweep with contrast controls in the same run: `factor_correlations` reads **0/0** in CEE `77e2e7d9` (4,291 files) and UI (4,473 files), while the controls `factor_evppi` (24 files) and `p_win_sensitivity` (9 files) read non-zero — so the instrument sees. PLoT's 15 and ISL's 13 hits are all on the *receiving* side, classified file by file: **no populator anywhere.**

**"We build more than we plug in" is this programme's chronic failure #1.** A correct, cheap, honest guard for a state no producer can reach is the purest instance of it. That is the whole hold.

### What would change my mind — one of:

1. **A live wire capture** showing `factor_correlations` arriving. The lane was scrupulous that its claim is *"no code path in any of the four repos populates it"* and **not** *"a capture was taken and none was seen"* — those are different claims, and only the second would make this reachable.
2. **A decision to populate it**, at which point this guard should land *with* its producer, not eighteen months ahead of it.

### One defect to fix first either way

The lane surfaced that whenever this notice can fire, ISL has **also** suppressed `factor_sensitivity` and `stability_thresholds` at the same gate. So the copy's reassurance — *"The ranking above is unaffected"* — survives only via PLoT's graph-primary re-derivation, and is **vacuous on the ISL-only fallback where `basis: 'none'`**. Shipping a guard whose reassurance is empty in one of its two reachable arms would be the guarantee-theatre class this estate hunts.

### The genuinely valuable output of this lane

Not the guard — **the finding**: ISL has a suppression gate on `factor_correlations` (`robustness_analyzer_v2.py:2903-2907`, `:2930`) and **nothing in the estate populates the field**. That belongs in the register whether or not this PR lands.

The work stands, the gate is green, and the branch keeps. This is a sequencing call, not a quality one.

PR #826, opened 2026-08-24
feat(explain-diff): mount "Why these changes?" on the applied-edit receipt [PR 2 of 2 — LAND SECOND]

Talchain:
## ⚠ LAND ORDER — THIS IS PR 2 OF 2, AND IT MUST LAND SECOND

| | repo | what it does | order |
|---|---|---|---|
| **PR 1** | CEE [#1082](https://github.com/Talchain/olumi-assistants-service/pull/1082) | registers the handler at `/assist/v1/explain-diff` | **merge + deploy FIRST** |
| **PR 2 — this one** | UI | proxy allowlist + client + host | merge only once #1082 is **serving** |

**Merging this first ships a button that 404s on every assistant edit** — the exact "advertises an action that terminates in refusal" defect this lane exists to remove. CEE #1082 is user-invisible on its own and safe to land alone.

⚠ A freeze window is in force; neither PR is expected to land same-day.

---

## What a user gets

**After the assistant edits their model, the receipt card now offers "Why these changes?", and answers it with CEE's explanation of that specific change.**

## Why it was dark

Not one missing wire — **four independent breaks**, in a component with zero production importers:

1. **No transport.** It POSTed to `/bff/assist/explain-diff`. That seam exists only in `vite.config.ts` for dev; production Netlify has no redirect and no edge function for `/bff/assist/*`.
2. **No reachable path even if it had one.** `/bff/cee/*` rewrites to `/assist/v1/*`, and the handler lived only at `/assist/explain-diff` (legacy). No `<x>` produces it. → fixed by CEE #1082.
3. **Wrong request.** It sent `{patch: string, context}`; the route needs a structured patch and is `.strict()`, so `context` is rejected outright.
4. **Wrong response — and it fabricated.** It read `data.explanation`, a key this route has never returned, and on `undefined` rendered **"No explanation available"** — a false failure report at the exact moment the server answered in full.

### Live probe, with controls in the same run

| path | status | content-type | size |
|---|---|---|---|
| `/bff/assist/explain-diff` (target) | 404 | `text/html` | **3449 B** |
| `/bff/assist/totally-fabricated-xyzzy` (neg control) | 404 | `text/html` | **3449 B** |
| `/bff/nonexistentfamily/foo` (neg control) | 404 | `text/html` | **3449 B** |
| `/bff/cee/graph-readiness` (contrast control) | **400** | **`application/json`** | live CEE JSON |
| `/bff/orchestrate/v5/turn` (contrast control) | **403** | `application/json` | edge fn alive |

The target is **byte-identical to a fabricated route**, while two contrast controls return two different live answers — a discriminating probe, not a blind one.

Confirmed at the deployed bytes: an **83-chunk crawl** of the staging bundle finds `Explain Diff`, `Explain this diff` and `/bff/assist/` in **zero** chunks (positive controls `key-question-card`, `No change` non-zero; fabricated negatives zero). The component was not merely unmounted — it was absent from the shipped build. It is one of **8 of 9** components in `src/components/assistants/` with no production importer.

## Why this host

`V5GraphPatchBlock` — the applied-edit receipt. It is where the user is at the moment the question arises, and **the only surface that knows which change is being asked about**. `useConversation.ts:4057` already calls it "the graph_patch card".

- Rendered for **`applied` only**. A `noop` has no diff to explain and would earn a 400 "patch has no changes to explain".
- Host presence verified at the deployed bytes (`v5-change-receipt`, `v5-change-status`, `v5-change-summary` all in the bundle), **not** from flag prose.
- The `v5_graph_patch` case in `InlineBlocks.tsx` carries **no block-level flag gate**, unlike its V4 neighbours.
- **The mount path is asserted, with a positive control**, so a flag appearing on that case REDs instead of silently darking this again.

## Honesty rules the client now follows

- What is shown is **always the server's text**. Never synthesised.
- `parseExplainDiffResponse` **refuses rather than repairs**: an unrecognised body yields `null`, and `null` means the UI says plainly it could not answer. No placeholder that could be mistaken for an explanation, no partial-credit path.
- A non-JSON body — the SPA-catch-all signature — reads as unavailable, never as an empty explanation.
- The failure notice leaves a **real route**: the chat composer sits directly below the card.
- Every path out of `loading` reaches a terminal state; no spinner that never resolves.

## ⚠ A BLOCKER THE REVIEWER MUST WEIGH BEFORE THE DEPLOY STEP

**Only one of three providers can serve this route**, and I could not derive staging's posture from this seam:

| `LLM_PROVIDER` | outcome |
|---|---|
| `anthropic` | ✅ works |
| `fixtures` | ❌ **500 on every real payload from this UI** |
| `openai` | ❌ 400 `not_supported` — `openai.ts` explicitly does not implement `explainDiff` |

Every V5 receipt operation is an **update**, never an add; `FixturesAdapter.explainDiff` builds rationales from `patch.adds` only, so an updates-only patch yields zero rationales and `.min(1)` 500s. Pinned as an exact test in CEE #1082.

**Posture must be derived from the Render API or a live witness, never YAML.** If staging is not on `anthropic`, hold this PR: the client will degrade honestly (the "couldn't get an explanation" notice), but every user would see that notice every time — a working affordance that never works.

## Verification

**RED-first**, allowlist guard at pristine edge function: `ON-LIST /bff/cee/explain-diff forwards…` → `expected false to be true` (proxy 404s, never forwards). Signature `1 failed | 53 passed (54)`.

**Discriminating mutant pair**, throwaway worktree at an absolute path **outside** the repo root. Isolation proven by **writing a sentinel** and asserting the source sha did not move, plus distinct inodes (`701631827` vs `701638139`), with an **existence precondition** — an earlier attempt in this lane "passed" all three isolation proofs against a worktree that had never been created. Restores are `git checkout HEAD -- <path>`.

| mutant | applied-check | result |
|---|---|---|
| control (pristine) | **0** | 88/88 green |
| **A — delete the `<ExplainDiffButton/>` render from the host** (0 renders remaining, verified) | 1 | **RED — 10 failed / 78 passed** |
| **B — unrelated surface, same card** (freshness copy + status label; **2 hits verified applied**, so not a false survivor) | 1 | **GREEN 88/88** |
| trailing control | 0 | 88/88 — matches pristine |

A alone proves sensitivity to *something*; the pair proves sensitivity to **this host binding**.

**Gate (`Staging Gate`), derived at this tip and run as its step sequence:**

- `pnpm run lint` → **exit 0** (hooks ratchet: 229 known violations across 13 files, exactly matching baseline)
- `pnpm run typecheck` → **exit 0** — `3604 / 3644` tracked files loaded; **`within baseline — 556 file(s) / 2252 error(s)`, zero drift**
- `ci:guard:zustand`, `ci:guard:starters`, `check:vendor` → **all PASS**
- Full unit suite → reported in a comment below once it concludes; **not claimed here, because it had not returned when this body was written.**

**The typecheck ratchet caught a real defect the green tests could not see:** `afterEach(() => vi.restoreAllMocks())` returns `VitestUtils`, not `Awaitable<void>` (TS2322, +1 over baseline). Fixed at the source — **the baseline was NOT regenerated**.

**Own-spec collection asserted by name**, never inferred from a total: `explainDiffRequest.spec.ts` (20) + `V5GraphPatchBlock.explainDiffMount.spec.tsx` (14) = **34**, each listed individually; plus `bffProxyPathAllowlist.spec.ts` 54. Pre-existing host specs re-run clean: `V5GraphPatchBlock.test.tsx` 23, `v5GraphPatchDescription.spec.ts` 30 — the receipt-leak guard still passes with the button inside the card.

## On the deleted tests
The four legacy `Explain Diff` cases were **removed, not ported**. Each passed against a hand-written mock of an imaginary server — a `patch` string, a `context` key the schema rejects, a `data.explanation` key that never existed. That is why this component sat unreachable with a fabricated failure message under a permanently green suite. Porting them would have carried the fiction forward. Real coverage now sits in the two specs above.

## Still owed after merge
A deployed-edge probe of `/bff/cee/explain-diff`, in the same discriminating form as the table above (fabricated-path negative + live-service contrast, one run). **A registration correct in source and unreachable in production is exactly the defect that created this cohort**, so the rung stays below MOUNTED until that probe returns.

### DIFF
--- a/netlify/edge-functions/cee-proxy.ts
+++ b/netlify/edge-functions/cee-proxy.ts
@@ -158,6 +158,17 @@ const ALLOWED_TARGETS: readonly RegExp[] = [
   /^\/assist\/v1\/scenarios\/[^/]+\/versions\/restore$/,
   /^\/assist\/v1\/decision-records\/commit$/,
   /^\/assist\/v1\/decision-records\/[^/]+\/outcome$/,
+  // Explain-diff (CEE #1082). Backs the "Why these changes?" affordance on the
+  // applied-edit receipt card. REQUIRES CEE #1082 to be serving — that PR adds
+  // /assist/v1/explain-diff; the handler previously existed ONLY on the legacy
+  // /assist/* surface, which this rewrite can never reach.
+  //
+  // Its ON-LIST and exact-match cases are in
+  // tests/ci-guards/bffProxyPathAllowlist.spec.ts, in this same PR — a missing
+  // entry here is precisely how this capability was dark: the component shipped
+  // calling /bff/assist/explain-diff, a seam that exists only in vite.config.ts
+  // for dev and has never existed in production.
+  /^\/assist\/v1\/explain-diff$/,
 ]
 
 function isAllowedTarget(pathname: string): boolean {
--- a/src/components/assistants/ExplainDiffButton.tsx
+++ b/src/components/assistants/ExplainDiffButton.tsx
@@ -1,77 +1,153 @@
 /**
- * N4: Explain Diff Button
- * Calls BFF /assist/explain-diff with patch context
- * Renders ≤280-char rationales inline
+ * ExplainDiffButton — "Why these changes?" on the applied-edit receipt.
+ *
+ * The user has just watched the assistant modify their model. This asks CEE why,
+ * and renders the SERVER'S answer.
+ *
+ * ── WHAT THIS COMPONENT USED TO DO, AND WHY THAT MATTERED ───────────────────
+ * The previous version was written against a contract that has never existed on
+ * any deployed surface, and it had never been imported by anything:
+ *
+ *   1. it POSTed to `/bff/assist/explain-diff` — a seam that exists only in
+ *      vite.config.ts for dev. In production a live probe returned the Netlify
+ *      SPA catch-all, byte-identical (3449 B) to a deliberately fabricated path;
+ *   2. it sent `{ patch: string, context: string }` where the route requires a
+ *      structured patch and rejects unknown keys (`.strict()`);
+ *   3. it read `data.explanation`, a key the route has never returned, and on
+ *      `undefined` rendered "No explanation available" — telling the user the
+ *      server had nothing to say at the precise moment it answered in full.
+ *
+ * (3) is the one worth keeping in mind while editing this file. A false failure
+ * report is not a cosmetic bug; it teaches users the feature is broken when it
+ * works, and it is the same class of harm as inventing an explanation.
+ *
+ * ── THE RULES THIS COMPONENT FOLLOWS ────────────────────────────────────────
+ * · What is displayed is ALWAYS the server's text. Never synthesised here.
+ * · If the server cannot answer, say so plainly and leave a route — the chat
+ *   composer is directly below this card, so "ask in the chat" is a real one.
+ * · Never claim an edit was explained when it was not. No placeholder copy that
+ *   could be mistaken for an explanation.
  */
-
 import { useState } from 'react'
-import { Lightbulb, Loader2 } from 'lucide-react'
+import { HelpCircle, Loader2 } from 'lucide-react'
+import { typography } from '../../styles/typography'
+import type { V5GraphPatchBlock } from '../../canvas/conversation/types'
+import {
+  buildExplainDiffRequest,
+  parseExplainDiffResponse,
+  type ExplainDiffRationale,
+} from './explainDiffRequest'
 
-interface ExplainDiffButtonProps {
-  patch: string
-  context?: string
-  onExplanation?: (text: string) => void
+/**
+ * The browser-reachable seam. `/bff/cee/*` is a Netlify EDGE FUNCTION that
+ * rewrites to CEE `/assist/v1/*` and injects the caller-auth key server-side.
+ *
+ * ⚠ Requires TWO things to be live, both shipped alongside this component:
+ *   · CEE registers `/assist/v1/explain-diff` (CEE #1082) — the handler
+ *     previously existed ONLY on the legacy `/assist/*` surface, which this
+ *     rewrite cannot reach;
+ *   · `/^\/assist\/v1\/explain-diff$/` is on `ALLOWED_TARGETS` in
+ *     netlify/edge-functions/cee-proxy.ts, or the proxy 404s before forwarding.
+ */
+const EXPLAIN_DIFF_ENDPOINT = '/bff/cee/explain-diff'
+
+export interface ExplainDiffButtonProps {
+  /** The applied-edit receipt being explained. */
+  block: V5GraphPatchBlock
+  /** Current model size, for context. Omitted rather than guessed if unknown. */
+  graphSummary?: { node_count: number; edge_count: number }
 }
 
-export function ExplainDiffButton({ patch, context, onExplanation }: ExplainDiffButtonProps) {
-  const [loading, setLoading] = useState(false)
-  const [explanation, setExplanation] = useState<string>('')
-  const [error, setError] = useState<string>('')
+type Status = 'idle' | 'loading' | 'answered' | 'unavailable'
+
+export function ExplainDiffButton({ block, graphSummary }: ExplainDiffButtonProps) {
+  const [status, setStatus] = useState<Status>('idle')
+  const [rationales, setRationales] = useState<ExplainDiffRationale[]>([])
 
   const handleExplain = async () => {
-    setLoading(true)
-    setError('')
-
+    setStatus('loading')
     try {
-      const response = await fetch('/bff/assist/explain-diff', {
+      const response = await fetch(EXPLAIN_DIFF_ENDPOINT, {
         method: 'POST',
         headers: { 'Content-Type': 'application/json' },
-        body: JSON.stringify({ patch, context })
+        body: JSON.stringify(buildExplainDiffRequest(block, graphSummary)),
       })
 
       if (!response.ok) {
-        throw new Error(`Failed to explain diff: ${response.statusText}`)
+        setStatus('unavailable')
+        return
       }
 
-      const data = await response.json()
-      const text = data.explanation?.slice(0, 280) || 'No explanation available'
-      setExplanation(text)
-      onExplanation?.(text)
-    } catch (err) {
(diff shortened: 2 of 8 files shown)


Talchain:
## Full test suite — the result promised in the PR body

**CI is the authority here, and it has concluded on the exact head SHA `2e2327cdad79809c5ccca656e4cb9c8437351404`.** All four shards of the full suite passed, alongside the required gate:

| check | conclusion |
|---|---|
| Staging Gate **(required)** | ✅ success |
| Full Test Suite (shard 1/4) | ✅ success |
| Full Test Suite (shard 2/4) | ✅ success |
| Full Test Suite (shard 3/4) | ✅ success |
| Full Test Suite (shard 4/4) | ✅ success |
| Full Test Suite Summary | ✅ success |
| TypeScript + Lint | ✅ success |
| Typecheck Gate Self-Test | ✅ success |
| Production Build | ✅ success |
| Security Audit (pnpm) | ✅ success |
| CEE Contract Validation | ✅ success |
| Flag Deployment Drift (report only) | ✅ success |
| Install & Cache | ✅ success |
| Visual Regression (advisory) | ❌ failure — **pre-existing, see control below** |

`total=14 completed=14 running=0 failed=1`. Polled to conclusion in the foreground; no verdict was formed while anything was running.

### The advisory red, proven rather than assumed
`Visual Regression (advisory)` is **already failing on the UI `staging` tip `f5794541`** — verified by querying that commit's own check-runs, not by repeating a known-reds registry. It is advisory and not in `required_status_checks.contexts`, which is `["Staging Gate"]`.

### Mergeability
`mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE` (the advisory red; required is green). **Not DIRTY**, so no stale-checks hazard. The branch is **1 commit behind `staging`**, and those commits touch **none** of `cee-proxy.ts`, `bffProxyPathAllowlist.spec.ts`, `src/components/assistants/`, `V5GraphPatchBlock.tsx` or `InlineBlocks.tsx` — no overlap with this seam.

### On the local run
A local `vitest run` over the whole suite was still executing when this comment was posted, so **no local full-suite number is claimed**. The CI shards above are the stronger evidence in any case: they run the real gate, in the real environment, against the exact SHA. The scoped local runs that *did* return are reported in the PR body with their counts asserted by name.

### ⚠ Still blocking the deploy step
- **CEE [#1082](https://github.com/Talchain/olumi-assistants-service/pull/1082) must be merged and serving before this lands.** Its required check is green (`Lint, TypeCheck, Unit Tests` ✅).
- **The `LLM_PROVIDER` question in the PR body is unresolved** and needs deriving from the Render API or a live witness. If staging is not on `anthropic`, this affordance degrades honestly but never succeeds — hold rather than ship it.
- The deployed-edge probe of `/bff/cee/explain-diff` is owed after merge. Until it returns, the rung is **DEPLOYED at best, not MOUNTED**.

Talchain:
## Merging — review MERGE. My priority attack was refuted, in the safe direction.

I sent the reviewer at this PR on the premise that it would make raw upstream error text user-visible: every explain-diff failure returns `500 INTERNAL` with `err.message` raw, and the `_not_supported` → 400 branch is unreachable.

**The CEE derivation was right; the conclusion was wrong. #826 never reads the error body.** On `!response.ok` it sets `unavailable` and returns — no `.json()`, no `.text()`, no `statusText`, no console write.

Driven in a real browser against a 500 carrying `Anthropic explain_diff failed: 529 {"type":"error",...}`:

```
'Anthropic explain_diff failed' / 'overloaded_error' / 'Anthropic'
'INTERNAL' / '529' / 'error.v1'                        -> 0 each
POS CTRL honest notice rendered                        -> 1
POS CTRL trigger present                               -> 1
CONTRAST fabricated string                             -> 0
runtime fetch URL                        -> /bff/cee/explain-diff
```

**This PR closes that surface rather than opening it** — the *previous* component surfaced raw text (its deleted test asserted `getByText(/Network error/)`). Every failure path now yields one honest string: *"Couldn't get an explanation for this change just now. Ask in the chat below and the assistant can talk it through."*

Credential non-leak verified independently, not inherited: `ASSIST_API_KEY` is read only into an outbound header via `Deno.env.get`, never into a response; zero secret tokens in the three client files, with `X-Olumi-Assist-Key` found 4× in the proxy as the positive control.

## The second lock
`/^\/assist\/v1\/explain-diff$/` — anchored both ends, no dynamic segment, and the check runs **before** key injection, so off-list paths 404 with no credential. No call site still uses `/bff/assist/explain-diff`. Discriminating pairs bind by identity: removing **this** entry REDs (M5) while removing a **different** entry stays GREEN (M7); removing the mount for all receipts REDs (M1) while gating a **different** receipt case stays GREEN (M2).

## Merge safety
Branch was **4 commits behind staging** and its green check never ran on the integrated tree. Staging had touched `InlineBlocks.tsx` — the file the mount spec reads from disk. Trial-merged in a throwaway worktree: **0 conflicts, 120/120 green across 7 named files, typecheck at exactly baseline** (3614/3654, ratchet 2252/2252). Counts asserted by name, never inferred from a total.

`Visual Regression (advisory)` is red and **not cited in either direction** — its noise floor measures ~110× its tolerance and its completeness guard reports zero screenshots captured. It is red on everything.

## Rung
**MOUNTED-by-code-path plus component-level visual evidence at 1280×800, 1440×900, 1512×982.** Explicitly **not** journey-witnessed — that rung is unreachable until this deploys. I will take the witness on staging.

## Three follow-ups rowed, none blocking
1. **No `AbortController` or client timeout** — proven with a never-settling fetch: the trigger stays disabled on "Asking…" indefinitely with no cancel. Bounded in practice by CEE's timeout and the proxy 504, so unreachable on the normal path — but the PR's own test named *"always leaves a terminal state"* covers **only rejection**, so its name is broader than its domain.
2. **`provenance_source` is parsed, preserved by a test, then dropped at render** — a `"hypothesis"` rationale is pixel-identical to a brief-grounded one. **The reviewer argues against rendering it as-is and I agree:** CEE types it `z.string().optional()`, free-form and model-authored, so a badge would present an unvalidated model-chosen label as a grounding attestation — worse than omitting it. The fix belongs on the CEE side, constraining the vocabulary first.
3. **No `aria-live` / `role="status"`** on the answer or the failure notice, so a screen-reader user gets no announcement when a ~5.6 s async result lands.

Also noted: explain-diff joins six LLM routes already unauthenticated at the edge. No new exposure class, but one more unauthenticated paid-model endpoint.

PR #827, opened 2026-08-24
fix(canvas): an invented target renders identically to a number the user wrote (B1-a)

Talchain:
## The defect, witnessed on the deployed build

UI `88cb7e37` · CEE `d1da670`, fresh disposable guest, 1280×800, 2026-08-24. On **one draw, on one factor** (`440d2e30` *Annual CRM Licensing Cost*) the product produced the discriminating pair, so nothing here is confounded:

| | invented | user-stated |
|---|---|---|
| `raw_value` | **22,500** | 45,000 |
| `unit` | — | `£` |
| `source` | **`cee_hypothesis`** | `brief_extraction` |
| `value_confidence` | `low` | `high` |
| `reasoning` | *"Model-chosen intervention level; **this amount is not stated in the brief**"* | *"…bound by edge `3d61e986` to `stated_items[3]`: our annual Salesforce licensing is £45,000"* |
| rendered | `Currently: 0.9 → [ 0.45 ]` | `Currently: 0.9 → [ 0.9 ]` |

Identical control, identical typography, no unit on either, no badge, tooltip, icon, colour or reasoning on either. **The only difference between a number Olumi invented and one the user wrote was the digits.**

Census over 41 rendered-text dumps and 16 mounted surfaces: `22,500` → 0 · `cee_hypothesis` → 0 · `not stated in the brief` → 0, against nine non-zero contrast controls in the same runs.

**The data was never wrong.** Every field above is correct on the wire. `unwrapInterventionValue` (`labelUtils.ts:633`) returned `{value, displayValue}` and discarded `source` at **the one point every render surface in the tree reads an intervention through**, so nothing downstream still had the fact to show.

## What changed

- **`labelUtils.ts`** — carry the producer's own `source` literal through the unwrap. Additive; the numeric and display halves are untouched.
- **`valueProvenance.ts`** — a **second** classifier, `classifyInterventionProvenance`, for the intervention vocabulary.
- **`model-tab-v2/{types,adapters}.ts`** — project each intervention's **own** stamp.
- **`model-tab-v2/ModelDetailRegion.tsx`** — mark it **inline, beside the number, in Plain**. And *"Where it came from"* no longer renders a heading with nothing under it (on an option it always did — the one element whose provenance is the finding).

**No stamp renders no mark** — not `"Not set"`, which is a claim about the *value*, and the value is set. Unknown stays unknown.

## ⚠ The thing worth reviewing: the two `source` fields are different questions wearing one name

Adding `cee_hypothesis` to `SOURCE_CLASSES` was tried **first**. It typechecks, it renders the right words, and it REDs `sourceClassesCompleteness` — **correctly**. The vendored contract says so itself (`@talchain/schemas@0.48.0`, `dist/orchestrator/edit-tool-ops.js:229-238`):

> *"`source` IS ONE OF ITS FIELDS (`cee-v3.ts:284` — an enum of `brief_extraction | cee_hypothesis | user_specified`, meaning HOW THE INTERVENTION WAS DETERMINED). **That is a different field from node `observed_state.source` wearing the same name**, and CEE exempts it for exactly this reason."*

- `OBSERVED_STATE_SOURCE_LITERALS` — 12 literals, "who put this value here". No `cee_hypothesis`.
- `InterventionV3.source` — 3 literals, "how this target was determined". Not exported by the package.

They **overlap on `brief_extraction`** and are otherwise disjoint, which is exactly what makes the merge look harmless. Merging them would have answered a question about a factor's observed value that no producer asks, and would have permanently blinded the one guard able to see contract drift. **The kind taxonomy is shared; the literal vocabularies are not** — CLAUDE.md trap 21.

Because the contract does not export the intervention literals, no derivation is available: the guard is a **hand-written corpus pinned to the contract's own quoted enum**, asserted as a set equality so it REDs if the map grows *or* shrinks (trap 12d).

## Evidence

**RED-first at pristine** — 21 collected, **15 RED**, 6 green (contrast controls + the unknown-stays-unknown twins, vacuous at pristine and proven non-vacuous by M3).
**Now** — `interventionProvenanceIsDisclosed.spec.tsx` **26/26**, asserted by name and count.

Every fixture is copied byte-for-byte from `drawA-11-autosave-raw.json`, the deployed capture — including two things a fixture written from the defect report would not contain: `raw_value: 0` invented targets (half the invented ones on that draw), and option `682a7e2d`, which is `provenance: from_brief` **and carries an invented target anyway**.

### Mutant kit (throwaway worktree outside the repo root; isolation proven by writing a sentinel and asserting the source unchanged — trap 9g; restores HEAD-relative — trap 9h; applied-check scoped to `src/`, control exactly 0)

| mutant | want | got |
|---|---|---|
| **M1** carrier severed at the drop point (the pristine defect restored) | RED | RED 11/26 |
| **M2** provenance read off the **option node** — plausible, confident, right on 2 of 3 options | RED | **RED 5/26 — exactly the discriminating cases** |
| **M3** unknown defaults to `'ai'` instead of rendering nothing | RED | RED 1/26 |
| **M4** carrier keyed on the value, so `raw_value: 0` loses its stamp | RED | RED 4/26 |
| **M5** labels inverted (the live deployed inversion, reproduced) | RED | RED 5/26 |
| **M6** the two vocabularies re-merged | RED | RED 1/26 **+ contract guard RED 1/4** |
| **M7** empty-heading gate removed | RED | RED 1/26 |
| **PAIR-B** a *different* value's rendering changed on the *same* panel | GREEN | **GREEN 26/26** |
| pristine + trailing control | GREEN | 30/30 both; applied-check 0 |

M2 ∧ PAIR-B is the discriminating pair: neither alone shows binding.

⚠ M2's first form produced invalid TS and collected **zero** tests — an unapplied mutation is indistinguishable from an equivalent one. Rebuilt as valid TypeScript and re-run; the table shows the rebuilt result.

Local: `pnpm typecheck` PASSED, ratchet **2252 / baseline 2252** · `pnpm lint` 0 errors, hooks ratchet exactly at baseline.

## Scope — what this PR does NOT do

**Defect 2 (the inversion) is not fixable in the UI and is deliberately untouched.** On the witnessed draw factor `440d2e30` carries `observedState.source: 'cee_inference'`, `provenance: 'ai_inferred'`, `extractionType: 'inferred'` — with `raw_value: 45000`, the user's own figure. Every provenance-bearing field on the factor says AI. The UI badges read it faithfully; correcting it client-side would be a second authority over a server fact (trap 21). The seam is CEE `src/cee/transforms/schema-v3.ts:329-330`, upstream of it the `extractionType` the model emits. Routed separately.

**The Inspector v2 Option panel** — the surface in the original finding — is outside this lane's seam and still shows `0.45` bare. `OptionPanel.tsx:118` already destructures `unwrapInterventionValue`, so the `source` this PR carries is one field away for whoever owns that directory.

### DIFF
--- a/src/canvas/domain/valueProvenance.ts
+++ b/src/canvas/domain/valueProvenance.ts
@@ -181,6 +181,96 @@ export function classifyValueProvenance(
 }
 
 /**
+ * ─────────────────────────────────────────────────────────────────────────────
+ * ⭐⭐ THE INTERVENTION VOCABULARY — A DIFFERENT FIELD WEARING THE SAME NAME
+ * ─────────────────────────────────────────────────────────────────────────────
+ *
+ * An option's `interventions[factorId].source` answers **HOW THIS TARGET WAS
+ * DETERMINED**. A node's `observed_state.source` answers **WHO PUT THIS VALUE
+ * HERE**. Two questions, two closed vocabularies, one field name — and the
+ * shared contract says so itself, at `edit-tool-ops.js:229-238` in the vendored
+ * 0.48.0 package:
+ *
+ *   *"`source` IS ONE OF ITS FIELDS (`cee-v3.ts:284` — an enum of
+ *   `brief_extraction | cee_hypothesis | user_specified`, meaning HOW THE
+ *   INTERVENTION WAS DETERMINED). That is a different field from node
+ *   `observed_state.source` wearing the same name, and CEE exempts it for
+ *   exactly this reason."*
+ *
+ * ⚠⚠ THIS IS WHY `cee_hypothesis` IS **NOT** IN `SOURCE_CLASSES`, AND ADDING IT
+ * WAS THE FIRST THING THIS LANE TRIED (B1-a, 2026-08-24). It typechecks, it
+ * makes the surface render the right words, and `sourceClassesCompleteness`
+ * REDs on it — correctly, because the guard's complement assertion exists to
+ * catch exactly this: a literal classified here that the contract's
+ * `OBSERVED_STATE_SOURCE_LITERALS` does not declare. Merging the two maps would
+ * have made `classifyValueProvenance('cee_hypothesis')` answer a question about
+ * a factor's observed value that no producer ever asks, and would have made the
+ * one guard that can see contract drift permanently unable to.
+ *
+ * The two vocabularies OVERLAP on `brief_extraction` and are otherwise
+ * disjoint, which is what makes the merge look harmless and is why it is not.
+ *
+ * ⚠ WHAT IS SHARED IS THE **KIND**, DELIBERATELY. Both classifiers return the
+ * same `ValueProvenanceKind`, so every surface keeps ONE taxonomy and ONE
+ * `Record<ValueProvenanceKind, …>` register. The concepts are named apart at the
+ * literal layer and joined at the kind layer — which is the shape CLAUDE.md
+ * trap 21 prescribes for two authorities answering different questions.
+ *
+ * ⚠ THE CONTRACT DOES NOT EXPORT THESE LITERALS. The vendored package "does not
+ * carry InterventionV3" (same comment, same file), so unlike `SOURCE_CLASSES`
+ * this map has NO canonical list to be checked against and no derivation is
+ * available. Stated rather than discovered later: the guard for this map is a
+ * HAND-WRITTEN CORPUS pinned to the contract's own quoted enum, which is the
+ * other half of trap 12d — where you cannot derive, a corpus is what notices the
+ * list is short. If a future schemas minor exports the intervention literals,
+ * replace the corpus with a derived guard and delete this paragraph.
+ */
+const INTERVENTION_SOURCE_CLASSES: Readonly<Record<string, ValueProvenanceKind>> = Object.freeze({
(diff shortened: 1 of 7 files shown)


Talchain:
Closing as **fully landed**, not abandoned.

#830 branched from this PR's head `b672ee04` and was merged as a squash (`289dd483`), so `b672ee04` never becomes an ancestor of `staging` and GitHub will not auto-close this. The content did land.

Verified before closing, with a positive control:

```
merge-base 88cb7e37 -> #827 contributes 7 files
merge simulation: merge b672ee04 into staging 289dd483
  files changed by the merge: 0
positive control: staging HEAD~1..HEAD touched 10 files
```

All seven files this PR contributes are already present on `staging`.

The one open review note on it — *"On the Model tab, yes. On the Inspector — the surface the defect was witnessed on — no"* — is **closed by #830**, which carries the mark on both Inspector value branches and pins the mount path via `InspectorModal`.

⚠ One finding worth carrying: the obvious reuse here, `getExtractionLabel`, classifies through the **node** `observed_state.source` vocabulary, where `user_specified` is not a member. It falls to the default arm and returns **"Estimated by Olumi" for a number the user typed** — the exact inversion this pair exists to prevent (trap 21). #830's `T-VOCAB` pins both halves, and it is currently the **only** test that REDs under that substitution.

PR #832, opened 2026-08-24
feat(constraints): show the user the hard limit they stated (chain step 6)

Talchain:
**Component:** 1 + 6, the ratified hard-constraint chain. **Rung reached:** step 6 built, tested, mounted-witnessed locally at three widths. Not deployed.

Steps 1-5 are closed. Step 6 — the user seeing their own limit on the results surface — was absent. This closes the **first half** of the closure condition. The second half (seeing a breach) is a **CEE dependency**, stated exactly below, not guessed at client-side.

---

## The three bounded answers

**1. Does `goal_constraints[]` reach the browser, and by what route?**
**Yes — this is a render change, no transport change.** Two live ingress routes write the store's `goalConstraints` slice: `applyDraftResult` (CEE `draft_graph.goal_constraints`) and `applyV5State` (`add_constraint` graph patches, upserted by id). It also persists to `scenarios.graph.goal_constraints`. The brief's "zero `goal_constraints` egress in 13 captured bodies" is about **egress**, which this lane does not need — and `applyV5State.ts:1210` records that the response-**root** read is deliberately absent because `responseParser` demotes unknown top-level keys to `__additive__`.

**2. Which surface is the right home?**
**`DecisionOverviewCard`** — mounted at `OutputsDock.tsx:3156`, already reads the same slice, already showed the count. Not the node inspector: every node panel ships inside a literal unconditional `<fieldset disabled>` (`InspectorRouter.tsx:221,334`), confirming the brief.

**3. Is the breach magnitude computable from data the browser already holds?**
**No — out of scope, returned as a dependency.** Swept for any per-option value keyed by the constrained node (`node_values|per_node|nodeValues|factor_values|node_outcomes`): **zero**, with two contrast controls firing non-zero in the same run. Options carry only `p10/expected/p50/p90` on the **goal** axis, normalised then `* scale`. The brief's example assumes the browser holds the £90,000; it holds the £50,000 cap and not the £90,000. This repo has already measured that subtracting a threshold from those numbers is a frame error — `SuccessTargetRow`'s L62 note records "£200,678 short" as a fabricated distance, removed for exactly this reason.

---

## What was built

| file | role |
|---|---|
| `src/components/results/decision-overview/statedLimits.ts` | pure projection: `selectStatedLimits`, `selectStatedLimitsKey`, `parseStatedLimitsKey` |
| `src/components/results/decision-overview/DecisionOverviewCard.tsx` | renders the limits in the expanded region |
| `__tests__/DecisionOverviewCard.statedLimits.spec.tsx` | 9 tests |

Reads **only** `label`, `operator`, `value`, `unit`. It does not accept or thread `probability` at all — the distrusted producer numbers stay gated.

---

## Evidence

**RED-first**, 9 collected by name, 5 failed at pristine:
- `shows the £50,000 budget cap, bound by its constraint id`
- `shows the 78% margin floor, bound by its OWN constraint id`
- `renders the region once both limits are present`
- `does not render the constraint probability that rode in on the fixture`
- `shows the boundary alone when the producer sent no label`

**Twins**

| direction | result |
|---|---|
| no stated limit | **byte-identical to pristine `4984ea4c`** — same sha256 `4276d6e8…`, 5,066 bytes; positive control confirms a real card, not an empty dump |
| a stated limit | `Programme cost ≤ £50,000` etc., bound by `constraint_id` |
| live browser twin | region present with limits → **absent** after clearing; chip reverts to "Nothing set as a hard limit"; `92%` appears nowhere in the document |

**Discriminating mutant pair** (worktree outside the repo root, isolation proven by **writing a sentinel** — source sha unchanged, inodes differ; restores HEAD-relative; applied-check scoped to `src/`, control exactly 0; clean tree before and after; trailing control 9/9):

| mutant | budget test | margin test |
|---|---|---|
| break render for **all** constraints | **RED** | — |
| break render for a **different row only** | **GREEN** | **RED** |

**Gate** — derived the required check's step sequence rather than assuming: `Staging Gate` needs `tsc` + `typecheck-selftest` + `vitest` + `vitest-summary` + `build`. All 8 `tsc` steps pass locally in order (**`lint` runs first**), `typecheck:selftest` 18/18, `build` passes, decision-overview suite 89/89.

**Mounted evidence** at **1280×800, 1440×900, 1512×982** — real app, fresh disposable guest, never the user's account. The `Visual Regression (advisory)` job is not cited in either direction, per the brief.

---

## CEE dependency — exact contract

To close the second half (a breach), CEE must forward on the CEE→UI hop:

1. **`constraints_status`** (`'computed' | 'unavailable' | 'skipped' | 'error'`) — already produced by PLoT #333 and already read defensively at `mapV5AnalysisToReport.ts:1329`, which is a **no-op today** because the key is absent from CEE `compose.ts` `P0B_SAFE_TRANSPORT_ENRICHMENT_KEEP`. *(The brief said an 18-key allowlist; the repo says **11**.)* Adding it needs **no UI change** to arrive.
2. **A per-option compliance verdict** — the eligibility decision, not a probability: constraint id + breach/comply/unevaluable, and where breaching, the option's value on the constrained node **in the constraint's own units and frame**. Without the frame attestation this is the L62 fabrication again.

Clause 3 ("say so and name what is missing") is **not** rendered, deliberately: saying "not evaluated" would be as false as "met" — evaluation happens upstream, its verdict just does not reach this surface.

---

## Corrected premises

1. **`SuccessTargetRow` cannot be reused — it has zero product call sites.** The brief says it "already exists and is switched OFF"; it is not merely starved by `PLOT_PER_OPTION_CONSTRAINTS_SUSPECT`, it is **unmounted**. Flipping that constant would still render nothing. Hosting step 6 there would have shipped it dark.
2. **My own mount probe produced a false absence, self-caught.** `<Component[ />]` cannot match a component whose props open on the next line. Re-derived: `SuccessTargetRow` 0 and `DraftNotes` 0 (both genuine), but `PreAnalysisPanel` 1 and `SuccessTarget` 1. `SuccessTarget` **does** render `{label} {operator} {value}` and is reachable — but only pre-run, only on the legacy panel (V3 supersedes it on `preAnalysisV3`), and only inside the **default-collapsed T3 Advanced accordion**. So the limit was reachable for a power-user *before* running and absent from the surface where the recommendation is read. The gap is real and narrower than "rendered nowhere".
3. **CEE's keep-list is 11 keys, not 18.**

## Observation, out of scope — not a verdict

`GoalPanel.tsx:405-424` renders `probability` as a percentage **and a `DataBar`** for each constraint, sourced from `results.report.goal_constraints[]` via `useGoalConstraints()`. That is a **different seam** from the per-option `constraint_analysis` block gated by `PLOT_PER_OPTION_CONSTRAINTS_SUSPECT`. I did **not** derive whether that seam is trustworthy, and did not touch it. Flagging it because if it is the same distrusted quantity, it is a live confident-wrongness surface in the chain this work belongs to.

### DIFF
--- a/src/components/results/decision-overview/DecisionOverviewCard.tsx
+++ b/src/components/results/decision-overview/DecisionOverviewCard.tsx
@@ -33,7 +33,7 @@ import { computeSuccessState } from '../../../canvas/components/pre-analysis-v3/
 import { computeGraphFacts } from '../../../canvas/components/pre-analysis-v3/selectors/graphFacts'
 import { ActionsMenu } from './ActionsMenu'
 import { REVIEW_BRIEF_ASK } from './actionsCatalogue'
-import { parseStatedLimitsKey, selectStatedLimitsKey } from './statedLimits'
+import { formatStatedLimitsNote, parseStatedLimitsKey, selectStatedLimitsKey } from './statedLimits'
 
 export type BriefStateOverride = 'thin' | 'contradictory' | 'unverified'
 
@@ -395,6 +395,30 @@ export function DecisionOverviewCard({ title, stateOverride }: DecisionOverviewC
         ? { line: OVERVIEW_COPY.thin, note: OVERVIEW_COPY.thinLiveNote }
         : STATE_COPY[state]
 
+  // ⭐ THE CLOSURE CONDITION, ON THE COLLAPSED CARD.
+  //
+  // Post-analysis `hasResult` is structurally true, so `autoExpand` reduces to
+  // `state === 'blocked'` — on the ordinary success path this card is
+  // COLLAPSED. The limits list below lives inside `{expanded && …}`, so
+  // rendering ONLY there meant a user who said "the budget cannot exceed
+  // £50,000" and got a successful analysis saw the word "constraints" and no
+  // £50,000. (That is the same criterion this lane used to disqualify
+  // `SuccessTarget` for sitting in a default-collapsed accordion.)
+  //
+  // The limit is therefore named in the note that already renders collapsed,
+  // on the one line it already occupies — NOT by lifting the list out of
+  // `expanded`, which would push the verdict down and regress ANSWER-FIRST.
+  //
+  // ⚠ `ready` ONLY, and that is a deliberate honesty boundary, not an
+  // oversight. Every other state's note is doing more urgent work than an
+  // orientation line: `blocked`/`contradictory` say "Resolve it before relying
+  // on the read", `needs_input` points at the questions, `thin` reports the
+  // missing success measure. A limit must never displace a warning about
+  // whether the result can be trusted. `blocked` auto-expands anyway, so the
+  // full list is visible there; on `thin`/`needs_input` the limit stays one
+  // click away, which is disclosed rather than hidden.
+  const collapsedLimitsNote = state === 'ready' ? formatStatedLimitsNote(statedLimits) : null
+
   // UI-SEM-077 (+ V7 L2, values-not-labels): decision-classification pill
   // inference. No producer classification contract exists; the only honest
   // client-side input today is the decision node's brief timeframe
@@ -570,7 +594,13 @@ export function DecisionOverviewCard({ title, stateOverride }: DecisionOverviewC
         />
         <span className="min-w-0 flex-1">
           <span className={`${typography.panelHeader} block text-text-header`}>{copy.line}</span>
-          <span className={`${typography.panelMeta} block text-text-light`}>{copy.note}</span>
+          <span
+            className={`${typography.panelMeta} block text-text-light${collapsedLimitsNote ? ' truncate' : ''}`}
+            data-testid={collapsedLimitsNote ? 'brief-bar-stated-limits' : undefined}
+            title={collapsedLimitsNote ? statedLimits.map((l) => l.text).join(' · ') : undefined}
+          >
+            {collapsedLimitsNote ?? copy.note}
+          </span>
         </span>
         <ChevronDown
           size={16}
--- a/src/components/results/decision-overview/DecisionOverviewCard.tsx
+++ b/src/components/results/decision-overview/DecisionOverviewCard.tsx
@@ -20,7 +20,7 @@
  * DS v4/5: bg-panel card, panel typography tokens only, Lucide, sentence
  * case, en-GB, no em dashes in prose.
  */
-import { useEffect, useState } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 import { ChevronDown } from 'lucide-react'
 
 import { useCanvasStore } from '../../../canvas/store'
@@ -33,6 +33,7 @@ import { computeSuccessState } from '../../../canvas/components/pre-analysis-v3/
 import { computeGraphFacts } from '../../../canvas/components/pre-analysis-v3/selectors/graphFacts'
 import { ActionsMenu } from './ActionsMenu'
 import { REVIEW_BRIEF_ASK } from './actionsCatalogue'
+import { parseStatedLimitsKey, selectStatedLimitsKey } from './statedLimits'
 
 export type BriefStateOverride = 'thin' | 'contradictory' | 'unverified'
 
@@ -83,6 +84,11 @@ export const OVERVIEW_COPY = {
   // risks (L3-BROWSER-TRUTH §9 C7). The sentence must be unambiguously about
   // what the MODEL has set, which is the only thing `constraintCount` knows.
   constraintsNoteEmpty: 'Nothing set as a hard limit',
+  // ⭐ STEP 6. The possessive is load-bearing and earned: every limit under
+  // this heading is one the USER stated and CEE recorded verbatim, never a
+  // value Olumi derived. The heading makes NO claim about whether a limit is
+  // met — that verdict does not reach this surface (see the render site).
+  statedLimitsHeading: 'Your stated limits',
   optionsNoteEmpty: 'No options mapped yet',
   // V6-RESPEC §4: empty classification fields fold into ONE muted aggregate
   // chip ("+N to set") — collapsed shows what IS, never a per-field inventory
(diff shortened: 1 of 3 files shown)


Talchain:
## FIX-FIRST addressed — `e457c0b1`

### The finding was correct, and the criticism lands

The limits rendered only inside `{expanded && …}`. Post-analysis `hasResult` is structurally true, so `autoExpand` reduces to `state === 'blocked'` — on the ordinary success path the card is collapsed and **the limit did not render**. All nine original tests passed because each calls `openBrief()` first.

**That is the same criterion I used to disqualify `SuccessTarget`** for sitting in a default-collapsed accordion, applied to the rival and not to itself. No defence offered.

### Fix taken: **option 1** — name the limit in the collapsed `brief-bar` note

The closure condition is **unchanged**. I am not demoting it to "one disclosure click".

Not lifted out of `expanded` — that would push the verdict down and regress Answer-First, exactly as you measured.

**Collapsed form is first-limit-in-full + "+N more", not a truncated join.** CSS truncation would cut a value mid-number ("≤ £50,0…") on the one surface whose job is showing the figure the user stated. First-plus-count guarantees one complete correct limit is always visible and ranks nothing — the order is the producer's extraction order.

⚠ **`ready` only, and that is a deliberate boundary I want on the record.** Every other state's note is doing more urgent work: `blocked`/`contradictory` say *"Resolve it before relying on the read"*, `needs_input` points at the questions, `thin` reports the missing success measure. A limit must never displace a warning about whether the result can be trusted. `blocked` auto-expands anyway, so the full list shows there; on `thin`/`needs_input` post-analysis the limit stays one click away. **Stated, not hidden, and pinned by a test.**

### Measured on the deployed-equivalent local build, fresh disposable guest

| measurement | result |
|---|---|
| `aria-expanded` on load, post-analysis | **`false`** — genuinely collapsed |
| collapsed note text, **no click** | **`Programme cost ≤ £50,000 · +2 more`** |
| `title` (full list) | `Programme cost ≤ £50,000 · Gross margin ≥ 78% · Compulsory redundancy rounds ≤ 1` |
| line count / truncated | **1 line**, `scrollWidth > clientWidth` = **false** |
| **card height with vs without limits** | **135px vs 135px — delta 0px** |
| probability leak (`92%`) anywhere in document | **false** |

Verified at **1280×800, 1440×900, 1512×982**. One pixel screenshot at 1280×800; the Browser pane stopped compositing after that, so the other two widths are **DOM-measured, not pixel-captured** — stated plainly rather than implied.

⚠ One honest note on process: my first browser check showed the fix absent. It was a **stale module in the tab**, proven by fetching the served module (`brief-bar-stated-limits`: 1 hit, matching disk) and inspecting the rendered span's attributes. I did not report the false negative as a result.

### Added, as asked: the gap that let this through

`POSITIVE CONTROL on the precondition: the card really is COLLAPSED here` — asserts `aria-expanded === 'false'` post-analysis and that the constraints chip is absent. **It passed at pristine, independently reproducing your finding inside my own suite.** Previously a flag move or a default-expansion change would have left all nine tests green while the user saw nothing.

### Evidence re-derived at the new tip (the old evidence was about a different tree)

- **Byte-identity, no-limit case:** re-run **collapsed *and* expanded** (the collapsed markup is now conditional). **Both IDENTICAL** to pristine `4984ea4c`. Positive control confirms distinct non-empty dumps (2,439 / 5,066 bytes).
- **Discriminating mutant pair on the collapsed note:** break it for all limits → **RED**; break only the `+N more` half → single-limit case **GREEN**, count-disclosure case **RED**. Isolation proven by writing a sentinel (source sha unchanged, inodes differ), HEAD-relative restores, applied-check scoped to `src/` with control exactly 0, clean tree before and after, trailing control 14/14.
- 14/14 this spec · 94/94 decision-overview · lint, typecheck, all five guards, build **pass**.

---

## ⚠ My CEE dependency was wrong — I withdraw it as written

**The keep-list is 18 keys, not 11.** My "correction" came from a stale UI-side comment (`mapV5AnalysisToReport.ts:1311,1323` + four doc sites) — a hand-maintained mirror, and I inverted a correct figure by trusting a comment over the producer. **The original brief was right and I was wrong.** Those five comment sites want fixing separately.

**"No UI change is needed" is refuted and withdrawn.** `widened.constraints_status` is **write-only — zero product rendering consumers**; the only readers are the debug drawer (reading the *raw PLoT response*) and a raw-V2 passthrough for CEE context. Contrast: `confidence_tier` renders in three live product files.

**Corrected dependency — three pieces, not one key:**
1. a **schemas release** adding it to `CEE_UI_ENRICHMENT_KEEP_LIST` (CEE's drift bolt asserts element-for-element equality, so a CEE-only change cannot land);
2. **CEE** adding it in lock-step;
3. a **UI rendering consumer**, which does not exist.

Also: **absent ≠ `'unavailable'`** — PLoT omits the field entirely when there are no constraints, so those are different states and a consumer must not conflate them.

⚠ **Name collision for whoever builds piece 3:** `report.constraints_status` is a **string enum**; `AnalysisInputsSummary['constraints_status']` is an **array of `{label, satisfied, probability?}`**. Same name, different concept.

**And your narrower reason for the clause-three omission is the better one, so I have adopted it:** a surface that never carries the verdict cannot honestly report the verdict's absence — and silence stops being correct the moment step 6 becomes the only place the limit appears. That is a further argument for option 1, which is what I took.

Ready for re-review.

Talchain:
## Merging — delta review MERGE. Constraint step 6 closes.

The blocking finding is closed and **verified by the instrument that found it**. Independent three-way probe at `e457c0b1`, deployed post-analysis state, no interaction:

```
POSITIVE  card mounts                    ✓
          aria-expanded on load          false   (genuinely collapsed)
TARGET    limit visible, NO CLICK        ✓  'Programme cost ≤ £50,000 · +2 more'
          full list region on load       false   (no Answer-First regression)
CONTRAST  one click reveals full list    ✓  'Gross margin ≥ 78%'
NEGATIVE  no limits -> note reverts, no testid
          single limit -> no '+N more'
```

**The control the lane added genuinely bites.** Run against the *old* product code at `f6187cfb` in an isolated worktree: the 2 closure tests **RED**, the precondition control passes at both tips (correctly — it pins the precondition, not the fix), the other 12 pass. **So the fix is verified by a route with no browser and no bundle in it**, which disposes of the stale-module question entirely.

**Answer-First is structurally safe, not merely measured.** The note **replaces** `copy.note` inside an existing span rather than adding a node — element count **18 vs 18**, brief-bar spans **4 vs 4**, identical tag shape. A 0px height delta is therefore **guaranteed at every width by construction**, not sampled at one.

**Twin re-measured at the new tip**, not carried over: collapsed and expanded both byte-identical to pristine `4984ea4c`, with a sensitivity control proving the probe discriminates (collapsed-with-a-limit differs). A model with no stated limit is unchanged.

**New mutant pair discriminates:** note always `null` → 2 RED; drop `+N more` keeping the first limit → **exactly 1 RED**, with the limit assertion staying GREEN. Two different tests under two mutants.

## ⚠ Two corrections recorded with the merge

**1. The state rationale cites a state that cannot occur.** `contradictory` and `unverified` arrive **only** via the `stateOverride` prop, and **no product call site passes it** — the live mount is `<DecisionOverviewCard title={overviewTitle} />`. They are test-only. The reachable set is five `liveState` values, and the limit surfaces in both states where the analysis is usable: **ready** (collapsed line) and **blocked** (auto-expands to the full list).

The reservation, rowed: in `thin` the displaced note names a genuine trust gap (no success measure). But **`needs_input` and `unassessed` notes are invitations, not trust warnings** — *"Work through the gaps with Olumi when you are ready"* is not more urgent than the user's own stated limit in the way `blocked` and `thin` are. Defensible (framing is incomplete, the limit is one click away, the full list renders there) but the stated rationale is weakest exactly there.

**2. ⭐ The code comment overclaims, and the measurement is fixture-bound.** `truncate` (`overflow:hidden; text-overflow:ellipsis; white-space:nowrap`) is applied to the note span **including the first limit** — a pixel-boundary cut with no content awareness. With a long label:

```
Total programme delivery cost including contingency and overheads ≤ £50,000
  ->  '…contingency and overheads ≤ £50,0…'
```

So *"guarantees at least one complete, correct limit is always visible"* is **false as an absolute**. The measured `truncation: false` is true of the fixture, not of the input space — **the corpus omits the class that breaks it** (trap 13d(c)).

**Non-blocking** — the value stays reachable via `title` and one click, and the prior state showed no limit at all. **Rowed with its fix:** split the span so the label truncates and `{operator} {value}` sits in a `shrink-0` sibling, making the number structurally uncuttable; soften the comment either way.

## Rungs
`Staging Gate` **success** at `e457c0b1`; 13 success, **0 running**, 14 total. The advisory job is red and **not cited in either direction**. Merge-base is `4984ea4c` and the branch is not rebased onto `463fc931`; staging drift **does not touch this card or its dependencies**, so the twin baseline stays valid. Decision-overview suite **94/94** (was 89, +5).

The lane conceded the original finding without defence — *"I applied the default-collapsed disqualification to `SuccessTarget` and not to my own change"* — and withdrew its keep-list correction (**18 keys, not 11**) and its "no UI change needed" claim. None of that was re-litigated.

PR #833, opened 2026-08-24
UI: the dock's primary navigation publishes which panel is active (affordance sweep CC7)

Talchain:
**Component 11 — Product Experience.** Closes affordance-sweep cross-cutting observation 7, which had no owner in either repo.

## The defect, derived at the deployed tip `463fc931`

The workspace dock's **primary navigation** — `Olumi | Analysis | Model` — carried no tab semantics. `WorkspaceShellTabStrip` rendered a `<nav>` of plain buttons; `OutputsDock` rendered a **second** `<nav>` of plain icon buttons for the collapsed rail. Neither carried `role="tablist"`, `role="tab"` or `aria-selected`. Which panel was fronted was conveyed **only** by `text-info` + a coloured bottom border + a `color-mix` fill.

**Contrast control** (so this is a real absence, not a blind search): the same repo already implements the full WAI-ARIA tabs pattern in `HeroLensTabs.tsx` (roving tabindex, arrow keys), and `ComparisonCanvasLayout`, `ScenarioListPage`, `ModelRowView` and `DebugPanelV2` all carry `aria-selected`. The instrument can see these attributes in this tree; it found none on the dock. **The product's *secondary* lens switcher was a proper tablist while its *primary* workspace navigation was not.**

This is a **truthfulness** defect before it is an accessibility one. "Which panel am I looking at" is state the product knows and did not publish — and the dock fronts itself **without user action** (the run-start auto-switch, `runReturnSignal.ts`, and the assistant's `open_panel` directive), so a user can be relocated between panels with the active state carried by colour alone. `aria-selected` on a plain button is inert, so `role="tab"` is what makes the fix real rather than decorative.

## What shipped

- **Both instances**, pinned separately by their own testids so fixing one cannot make the other's case pass: `role="tablist"`, `role="tab"`, `aria-selected` in **both** directions, roving tabindex, `Left/Right/Home/End` moving focus and selection together.
- The dock body is now the `role="tabpanel"` the strip controls, labelled from `effectiveActiveTab` so it cannot announce a tab the user has left.
- The collapsed rail keeps each icon's `aria-label` as **protected content** — it is the only label an icon-only control has.
- Removes an **inert** `aria-label="Outputs sections"` from a role-less wrapper `<div>`; a generic container cannot take an accessible name, so it did nothing while duplicating the `<nav>`'s name.

## ⚠ The correction worth reviewing

The first version put `role="tablist"` **on** the `<nav>`. An explicit role **replaces** the implicit one, so the `navigation` landmark ceased to exist and **18 assertions across three sibling suites went RED against a 73/73 green baseline at `463fc931`**. The tablist is therefore **nested inside** the `<nav>`: landmark outside, widget inside, every existing binding intact.

24 query sites in two sibling suites move from `getByRole('button')` to `getByRole('tab')`, **for the five tab names only**. This is required for **correctness**, not to make them pass — the absence assertions (`queryByRole(…, {name:'Compare'|'Journey'}).not.toBeInTheDocument()`) would otherwise have started passing **vacuously**. Contrast control: the `Collapse`/`Expand`/`Dismiss` button queries are deliberately untouched and still resolve as buttons.

## Evidence

**RED-first:** 17 cases, **17/17 failing at pristine**, 17/17 passing after. Collected count asserted by name.

**Mutants** — worktree outside the repo root, isolation proven by **writing a sentinel** (source hash byte-identical, worktree diverged; inodes compared for the APFS hard-link trap). Applied-check scoped to `src/`; leading **and** trailing controls at exactly zero, 17/17.

| Mutant | Result |
|---|---|
| **M1** expanded strip loses `aria-selected` | **7 RED — all expanded cases; rail cases GREEN** |
| **M2** the *different instance* — rail loses `aria-selected` | **2 RED — both rail cases; expanded cases GREEN** |
| **M3** naive fix, every tab claims selected | 3 RED (the opposite-direction twin bites) |
| **M4** key handler stops discriminating | 1 RED |

**M1/M2 are the discriminating pair**: each REDs only its own instance, proving the assertions bind to the named object and not to "any tab anywhere".

**M4 initially SURVIVED, and the test was fixed rather than the mutant.** The case started on the *first* tab, where `return surfaces[0]` and `return null` are the same observable. It now starts on a non-first tab. An earlier harness bug is also recorded in-file: a first isolation check reported PROVEN against a worktree that had failed to create, because an empty hash passed a `!=`.

**Opposite-direction twins:** inactive tabs must carry `aria-selected="false"` (not an absent attribute, and not the naive all-true); an unrelated key must be inert; the tabpanel label must *follow* the selection rather than stick to the first tab.

**Mount path asserted, not assumed:** every case drives `OutputsDock`, with `aiPanelV2` mocked ON **explicitly** so the suite REDs loudly if the deployed posture moves instead of passing vacuously.

**Mounted witness — fresh guest, three widths.** Driven on a **disposable localhost origin**, deliberately not staging: four sibling-lane tabs were live on the staging origin and clearing its storage would have destroyed their state.

At **1280×800, 1440×900 and 1512×982**: nav landmark preserved · `role="tablist"` present and named · exactly one `aria-selected="true"` · every tab declaring `true`/`false` · roving tabindex `-1,-1,0` · panel `role="tabpanel"` labelled by the **selected** tab · every tab's `aria-controls` pointing back at it · **zero** inert label claimants. A **real browser ArrowRight** moved selection Analysis→Model with focus following and the panel relabelling. The collapsed rail was measured with the same contract (`aria-orientation="vertical"`).

## Gates

- `pnpm typecheck` (`scripts/ci/typecheck-gate.sh`): **PASSED** — 3615/3655 files, ratchet 2252 = baseline 2252.
- `eslint` on touched files: **0 errors**; 4 warnings, **identical to the base commit** (parity checked at `463fc931`).
- `src/canvas/components/__tests__` + `workspaceShell`: **164 files / 1752 passed / 15 skipped**, up from 1734 + the 18 recovered. **All 25 specs that render `OutputsDock` or `WorkspaceShellTabStrip` live in that directory**, so the blast radius is covered.
- Repo-wide sweep: **zero** residual `role="button"` queries naming a dock tab.
- The full `vitest run` was still executing when this PR was opened; `Staging Gate` is the authority.

### DIFF
--- a/src/canvas/components/OutputsDock.tsx
+++ b/src/canvas/components/OutputsDock.tsx
@@ -56,6 +56,11 @@ import {
 import {
   WorkspaceShellCollapsedStrip,
   WorkspaceShellTabStrip,
+  DOCK_PANEL_DOM_ID,
+  DOCK_TABLIST_LABEL,
+  nextTabForKey,
+  railTabDomId,
+  tabDomId,
 } from './workspaceShell/WorkspaceShellTabStrip'
 import { AnalysisStateRegion } from '../../components/results/analysisState/AnalysisStateRegion'
 import { useAnalysisRunState } from '../../components/results/analysisState/useAnalysisRunState'
@@ -2538,10 +2543,28 @@ function OutputsDockBody({ sendMessage }: OutputsDockBodyProps) {
         {effectiveIsOpen && <AssistantOpenedNotice />}
       </div>
 
+      {/* The collapsed rail is the SECOND instance of the dock's tab strip, and
+          it carries the same contract as the expanded one: which panel is
+          fronted is state the product knows, so it is published rather than
+          left to colour. The rail is icon-only, so its `aria-label` is the
+          ONLY label each control has — it is protected content and the tab
+          semantics are added around it, never in place of it.
+
+          The tablist is NESTED INSIDE the `<nav>` rather than replacing its
+          role — an explicit role replaces the implicit one, and the sibling
+          suites resolve this element with
+          `getByRole('navigation', { name: 'Outputs sections' })`. Putting the
+          role on the `<nav>` deleted the landmark and turned eighteen sibling
+          assertions RED against a green baseline. Landmark outside, widget
+          inside: both are true and nothing that binds here has to change. */}
       {!effectiveIsOpen && (
-        <nav
+        <nav aria-label={DOCK_TABLIST_LABEL}>
+        <div
           className="flex flex-col items-center gap-2 py-3"
-          aria-label="Outputs sections"
+          role="tablist"
+          aria-orientation="vertical"
+          aria-label={DOCK_TABLIST_LABEL}
+          data-testid="outputs-dock-rail-tablist"
         >
           {OUTPUT_TABS.map(tab => {
             const Icon =
@@ -2558,6 +2581,17 @@ function OutputsDockBody({ sendMessage }: OutputsDockBodyProps) {
               <button
                 key={tab.id}
                 type="button"
+                id={railTabDomId(tab.id)}
+                role="tab"
+                aria-selected={effectiveActiveTab === tab.id}
+                tabIndex={effectiveActiveTab === tab.id ? 0 : -1}
+                data-testid={`outputs-dock-rail-tab-${tab.id}`}
+                onKeyDown={event => {
+                  const next = nextTabForKey(OUTPUT_TABS, effectiveActiveTab, event.key)
+                  if (!next) return
+                  event.preventDefault()
+                  handleTabClick(next.id)
+                }}
                 onClick={() => handleTabClick(tab.id)}
                 className={`flex items-center justify-center w-7 h-7 rounded-full border ${typography.caption} focus:outline-none focus-visible:ring-2 focus-visible:ring-info focus-visible:ring-offset-1 ${
                   effectiveActiveTab === tab.id
@@ -2572,6 +2606,7 @@ function OutputsDockBody({ sendMessage }: OutputsDockBodyProps) {
               </button>
             )
           })}
+        </div>
         </nav>
       )}
 
@@ -2604,6 +2639,14 @@ function OutputsDockBody({ sendMessage }: OutputsDockBodyProps) {
             `SHELL_INHERITED_BODY_TYPOGRAPHY` in the shell contract. */}
       {effectiveIsOpen && (
         <div
+          // The body IS the tab panel the strip controls, so it says so. The
+          // label is derived from `effectiveActiveTab` — the same value the
+          // strip renders its selection from — so the panel cannot announce a
+          // tab the user has left. A hardcoded id here would go stale on the
+          // first tab change and nothing would notice.
+          id={DOCK_PANEL_DOM_ID}
+          role="tabpanel"
+          aria-labelledby={tabDomId(effectiveActiveTab)}
           className={`${shellBodyClassName(surfaceFor(effectiveActiveTab))} ${typography.panelBody} text-text-header`}
           data-testid="outputs-dock-body"
           data-shell-scroll-owner={surfaceFor(effectiveActiveTab).scroll}
--- a/src/canvas/components/OutputsDock.tsx
+++ b/src/canvas/components/OutputsDock.tsx
@@ -58,7 +58,6 @@ import {
   WorkspaceShellTabStrip,
   DOCK_PANEL_DOM_ID,
   DOCK_TABLIST_LABEL,
-  nextTabForKey,
   railTabDomId,
   tabDomId,
 } from './workspaceShell/WorkspaceShellTabStrip'
@@ -2556,7 +2555,14 @@ function OutputsDockBody({ sendMessage }: OutputsDockBodyProps) {
           `getByRole('navigation', { name: 'Outputs sections' })`. Putting the
           role on the `<nav>` deleted the landmark and turned eighteen sibling
           assertions RED against a green baseline. Landmark outside, widget
-          inside: both are true and nothing that binds here has to change. */}
+          inside: both are true and nothing that binds here has to change.
+
+          ⚠ WHAT THE RAIL DOES NOT TAKE FROM THE EXPANDED STRIP: roving
+          `tabIndex` and arrow-key traversal. The reason is on the button
+          below — read it before adding either. The short version is that the
+          rail's activation path expands the dock, so an arrow key would
+          unmount the rail rather than move along it, and a roving index
+          without traversal simply deletes two icons from the tab order. */}
       {!effectiveIsOpen && (
         <nav aria-label={DOCK_TABLIST_LABEL}>
         <div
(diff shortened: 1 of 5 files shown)


Talchain:
## FIX-FIRST — the expanded strip is genuinely good; the collapsed rail regresses keyboard access

**What a user gains** (dock expanded): a screen-reader user hears *"Analysis, tab, selected, 2 of 3"* instead of three identical "button" announcements, and the dock body announces itself as that tab's panel. Left/Right/Home/End work with focus following selection. Real, and it works.

**What a user loses** (dock collapsed — **the first-use rail a fresh guest lands on**): two of three rail icons are removed from the Tab order, and the replacement navigation is on the wrong axis. Before this PR they could Tab to each icon and press Enter.

### F1 — the rail publishes a VERTICAL tab set and implements a HORIZONTAL one · new regression

Driven, not reasoned:

```
BASE       ariaOrientation:"vertical"  tabIndexes:[olumi:-1, results:0, model:-1]
                                       railTabsHaveAriaControls:[null,null,null]
ArrowDown  selected:["results"]  activeTabInStore:"results"     ← inert
ArrowRight railStillMounted:false  tabs:[GONE,GONE,GONE]
           focusIsBody:true  focusTag:"BODY"
```

`tabIndex={active ? 0 : -1}` is an **added** line on the rail button (~`OutputsDock.tsx:2587`); pre-PR these were plain `<button>`s, all Tab-reachable. The expanded strip solves this properly (`tabRefs.current.get(next.id)?.focus()`); **the rail has no equivalent**.

So a keyboard user on the first-use rail can Tab to **one** icon. The key the published orientation tells them to press does nothing. The key that does work expands the dock, unmounts the rail, and **drops focus on `<body>`**. **The spec has zero rail keyboard cases**, so nothing pins any of it.

**Minimal fix:** orientation-aware `nextTabForKey` accepting ArrowUp/ArrowDown on the rail; then either focus the newly-fronted expanded tab after `handleTabClick`, or drop the roving `tabIndex` on the rail and leave all three tabbable. Plus a rail-keyboard case with an opposite-direction twin.

### F2 — the tabpanel names a control that does not exist · reachable today

Driven through the real production route (`ReactFlowGraph.tsx:836 handleOpenCompare` → `OutputsDock.tsx:1942`, **no flag, no contract check**):

```
compareSurfaceOnScreen:true   panelRole:"tabpanel"
panelAriaLabelledby:"outputs-dock-tab-compare"
labelledbyResolvesToAnElement:FALSE
tabCount:3  selectedCount:0
```

`compare` is `presentedAsTab:false` — confirmed **in the deployed bundle** (`presentedAsTab:!0` ×3, `:!1` ×2). `shellContract.ts:374-388` documents this state itself and defers closing it to a separate change.

**The PR's headline claim inverts here:** the navigation publishes that **no** panel is active while a panel is on screen, and that panel points at a phantom. Pre-PR this was a silent gap; **post-PR it is an explicit false statement in the ARIA graph.**

You need not close the routing hole — but do not *assert* into it. Either omit `role="tabpanel"`/`aria-labelledby` when `effectiveActiveTab ∉ surfaces`, or pin it in an explicit known-state test asserting exactly this set, so it REDs if the set grows.

### F3 — cosmetic
`WorkspaceShellTabStrip.tsx`: *"The name lives on the tablist, once."* The code puts `DOCK_TABLIST_LABEL` on **both** the `<nav>` and the tablist. The nesting decision is right and well-evidenced; the sentence describing it is not.

### The mutant pair discriminates in both directions — this is above the estate's average

Isolation proved by **writing** a sentinel (source SHA-256 identical before/after; inodes differ). Applied-check scoped to `src/`.

| Mutant | Applied | Result |
|---|---|---|
| CTRL-LEAD | **0** | 17 passed |
| M0 revert both production files | 2 | **17 failed** — RED-first holds |
| M1 `aria-selected` off **expanded only** | 1 | 7 expanded RED · **all 4 rail GREEN** |
| M2 `aria-selected` off **rail only** | 1 | 2 rail RED · **all 12 expanded GREEN** |
| M3 panel label hardcoded to first tab | 1 | 2 tabpanel RED |
| M4 `nextTabForKey` default | 1 | the unrelated-key case RED, as its comment claims |
| CTRL-TRAIL | **0** | 17 passed |

M1/M2 prove binding by identity. The mount-path case, opposite-direction twins and store-sourced precondition are all genuine.

### Gate at head `ae1b3edd`
14 checks, `running=0`, **`Staging Gate` success**. Sole failure is `Visual Regression (advisory)`, which `staging-full-tests.yml:300` puts **deliberately outside** the Staging Gate aggregator's `needs`. **Corrected premise: `mergeStateStatus` is UNSTABLE, not CLEAN — solely because of that advisory red.** Trial-merge into `5113f5a0` clean, 0 conflicts; staging has not touched either production file since the merge-base. Zero overlap with `codex/c11-strategic-workspace-entry` @ `8b42bcc8`.

### Rung
**TESTED** (jsdom 17/17, mutation-proven), mount path derived at the deployed bundle. **Not MOUNTED, not WIRE-WITNESSED.** jsdom proves attributes and structure only — **no visibility, focus-ring or real-AT verification was performed**, and `aiPanelV2` OFF was not measured.

### Instrument failures the reviewer disclosed — worth reading
- Its **first bundle crawl fetched 2 of 82 chunks**: zsh does not word-split unquoted expansions, so the BFS queue died after round one and every dock marker read zero — **including markers that are present**. Re-run with typed arrays: 82 chunks, positive control non-zero, fabricated contrast zero. **Those first zeros were instrument blindness, not absence.**
- Its first mutant harness read `local ec=0` — the *assignment's* status, not vitest's — so every `EXIT=` printed 0.
- `rg -c` on minified single-line bundles counts **lines, not matches**.

### Two things noted out of scope
- `src/canvas/components/__tests__/zz.probe2.spec.tsx` — a 44-line probe file committed in #779, still in the suite.
- `netlify.toml:156` claims *"Compare tab — renders live on staging today"* while `presentedAsTab:false` in the deployed bundle says otherwise — a hand-maintained mirror telling readers the opposite of the contract.

Talchain:
## Mutant evidence for the new focus guard

Recording this in the thread, since the commit message refers to it here.

**Guard added:** `arrowRightMovesFocusAndNotOnlySelectionAlongTheExpandedStrip`, with its precondition
pinned **in-test** — focus is placed on `olumi` explicitly (`fireEvent.click` does not focus in jsdom)
and the destination is asserted **not** to hold focus first. So it cannot pass on a tree where focus
never moved.

**Why it was needed:** `WorkspaceShellTabStrip.tsx:178` (`tabRefs.current.get(next.id)?.focus()`) was
**entirely unpinned** — deleting it left **71/71 green**, and the word `focus` appeared **zero times**
in the 532-line spec. The roving-focus half of the WAI-ARIA contract was implemented correctly and
guarded by nothing.

**Mutant run:**
- Isolation proven **by writing**, not by locating: worktree outside the repo root, **distinct inodes**
  (`713355625` vs `713430333`), and a sentinel appended to the worktree copy left the source SHA-256
  byte-identical while the sentinel provably landed. *(`cp -R` can preserve APFS hard links, so
  locating a path proves nothing.)*
- Clean tree asserted **before**; applied-check scoped to `src/` returned **exactly 1** changed file.
- **RED on deletion:** `1 failed | 17 passed (18)`, failing at spec line 399 on `toHaveFocus()` —
  **my case and only my case.** Discriminating, not a blanket red.
- Restored from a **pristine archive** (`cp`), never `git checkout -- <path>` — that restores from the
  **index** and can write the defect back.
- Clean tree asserted **after**; trailing control re-ran **18/18 green**.

**Named limit, carried honestly:** jsdom does not implement sequential focus navigation, so **no test
here presses Tab**. The central fix-1 claim — all three rail icons are Tab-reachable again — rests on
the *absence* of `tabIndex` on `<button>` elements, which is a spec-level inference. The original
review's finding was jsdom-only in the same way. Only a real browser can witness the tab order.

PR #834, opened 2026-08-25
Step 6b: show whether the crowned option met the user's stated limits

Talchain:
## What a user sees differently

Where a user stated a hard limit, the results card now says **whether the recommended option met it — in the producer's own words** — instead of listing the limit and saying nothing.

---

## Phase 1 first: this began as a reachability question, and the prior was wrong

The lane was dispatched on the prior that **CEE drops** `robustness.recommended_option_compliance`. **Refuted.** CEE forwards it intact. The fields died *inside this repo*.

| Hop | Verdict | Evidence |
|---|---|---|
| PLoT emits | ✅ passes | `routes/v2/run.ts:3608-3609`, unconditional — no flag gate in the 200 preceding lines |
| CEE keep-list | ✅ passes | `robustness` is a member (`compose.ts:723`); projection `:1079` is `out[key]=stripInternalKeysDeep(src[key])` — shallow top-level **keep** + deep **deny**-strip of internal carriers only, so additive nested members ride through |
| CEE withheld-crown projection | ✅ passes | `projectLeaderDesignatingMembers` is a **deny**-list; `keyDesignatesLeadingOption`'s pattern `/^recommend(?:ed\|ation)_option(?:_(?:id\|label\|name))?$/i` is anchored with a **closed** suffix group these names do not match. Executed, with positive controls (`recommended_option_id`/`_label`/`leading_option`/`top_option_id` all `true`) and contrast controls (`display_verdict`/`score` `false`) |
| schemas pin skew (0.40.0 → 0.48.0) | ✅ **not** the breaking hop | `EnrichmentRobustnessSchema` is `.passthrough()` at the **shipped 0.48.0 vendored tarball bytes**. 0.48.0 does not *know* the fields (0 files; contrast `recommended_option_id` 6 files) but passthrough tolerates unknown keys |
| **UI mappers** | ❌ **BREAKING HOP** | `mapV5AnalysisToReport.ts` and `responseMapper.ts` both rebuild `robustness` as **key-by-key allow-lists**; neither listed them |
| **UI consumer** | ❌ **absent** | zero references |

**Absence evidence, contrast-controlled on both instruments:**
- Source sweep (repo HEAD == deployed `/version.json` commit `39162243`): target **0** files in CEE/UI/schemas; contrast `recommended_option_id` **12 / 70 / 11**.
- **Deployed-bundle sweep** (83 chunks, 5,950,662 bytes, BFS from `/assets/index-CUPZ6R6f.js`): both targets **0 chunks**; six same-family contrast symbols (`display_verdict`, `display_verdict_reason`, `recommended_option_id`, `conditional_winners`, `flip_thresholds`, `near_tie`) at **3–4 chunks each**; negative control **0**.

Settled by **derivation + deployed-artefact sweep**. **No live wire capture was taken** — that bound is not upgraded anywhere in this PR.

The V5 mapper's own comment already recorded this exact failure for `display_verdict`: *"ON-WIRE on Seam A … but previously dropped here."* This is that story one field on.

---

## The six-state rendering table

| enum | renders? | tone | treatment |
|---|---|---|---|
| `not_applicable` | **no** | — | Its sentence is "no limits were set for this decision" — it would contradict the user's own stated-limits list above it. Silence, and not a dead end: the limits still render |
| `compliant` | yes | positive | producer reason, verbatim |
| `uncertain` | yes | **unknown** | ⚠ never binarised — ISL publishes no breach threshold |
| `unverified` | yes | **unknown** | the scale was untrusted → no claim in *either* direction |
| `not_assessed` | yes | **unknown** | ⚠ **not** "no limits were set" — that would recreate the falsehood PLoT #338 fixed one layer up |
| `no_eligible_option` | yes | negative | ⚠ no `recommended_option_id` in this state, so the **reason** carries the message rather than an empty leader slot |
| field absent / unknown token | **no** | — | fail-closed; an older or future producer leaves the surface silent |

The UI **authors no copy for a verdict**. Every sentence is PLoT's, verbatim (`isl-to-ui.contract.ts:281`).

---

## Reuse, not a second vocabulary

- Tone classes are `DecisionOverviewCard`'s **own** `STATE_DOT_TONE` (`bg-success` / `bg-danger` / `bg-text-light` — already this card's "unassessed" tone), not a new palette.
- The renderer extends **UI #832's** live `stated-limits` region rather than minting a surface.
- `SuccessTargetRow.tsx` was **avoided deliberately** — confirmed zero product call sites (all four importers are specs; contrast control `TargetProbabilityBars` *does* return a product importer).
- Selectors return **primitives**, honouring the card's primitive-selector contract (`primitiveSelectors.spec`).
- The card's "deliberately silent about COMPLIANCE" note is **withdrawn in place, with its reason**: it was correct about `constraints_status` (which CEE genuinely strips) and does not hold for this field.

---

## Tests

**RED-first**, at pristine, by name:
- `mapV5AnalysisToReport.crown-compliance.spec.ts` — 7 collected, **3 RED**. ⚠ **Correction:** the first run printed 4 failures, but one was the positive control failing on a fixture id I had *guessed wrong* — my own spec defect, not the defect under test. The honest RED-first signature is **3/7**. (The control earned its place by catching it.)
- `DecisionOverviewCard.crownCompliance.spec.tsx` — 11 collected, **7 RED**.
- `crownCompliance.spec.ts` — written *after* its module (new code, no defect to reproduce). Disclosed in-file; its discrimination rests on the mutant pair, not a pristine RED.

**Discriminating mutant pair** (throwaway worktree at an absolute path outside the repo root; isolation proven by **writing** a sentinel with a positive control on the write, inodes differing; restores `git checkout HEAD --`):
- **Mutant A** — break rendering for **all** states → **14/30 RED**.
- **Mutant B** — break rendering for **`compliant` only** → the `no_eligible_option` assertion stays **GREEN** (binds by identity), while its own `compliant` twin **REDs** (proving the mutant applied and bites).
- Applied-check scoped to `src/`: leading control **0**, mutants **1**, trailing control **0**, markers remaining **0**, source clone dirty **0**.

**Opposite-direction twins** throughout: a compliant run still reads compliant; an unknown never becomes a breach (asserted in *both* directions — `not_negative` alone would also pass for a positive render).

**Precondition pinning**: every render case asserts the stated limit is on screen *before* asserting anything about compliance, so a case cannot reach the right verdict through a card that failed to render the limits region.

**Mount path, asserted against the DEPLOYED BUNDLE** (post-resolution inlined value, superseding `netlify.toml` and the dashboard):
- `VITE_FEATURE_DECISION_OVERVIEW:"1"` is inlined, and the resolver is `return"1"===t||"true"===t` → **TRUE**.
- The surface itself ships: `stated-limits`, `stated-limit-`, `Your stated limits`, `brief-bar-stated-limits` all present in the deployed bundle.
- Negative control: `crown-compliance` (this PR's new testid) **0 chunks** pre-deploy.

⚠ **Bound on that claim, stated rather than implied.** This establishes **"deployed, flag-gate proven open"** — *not* MOUNTED in the status-ladder sense. Mounting also requires the runtime data conditions `!W && mt && Me` (a completed analysis carrying a recommendation), which no bundle grep can settle. A MOUNTED rung would need a DOM witness, and this PR does not have one.

**Gate**, run in the required order (lint gates the rest):
- `npm run lint` — **0 errors**; hooks ratchet *"229 known violation(s) … exactly matching the baseline. No new conditional hooks."*
- `npm run typecheck` — **PASSED**, ratchet **2252 = 2252 baseline**, 3635/3675 files covered.
- Regression: **1801 passed / 0 failed** across 116 files (decision-overview, v5, adapters/plot).

---

## Review round 2 — the named change is in

**Gate corrected: `goalConstraintCount > 0`, not `statedLimits.length > 0`.** The review found my disclosed gap was narrower and worse than I framed it. `selectStatedLimits` (`statedLimits.ts:96-98`) **skips any constraint whose `value` is non-finite or whose `operator` is empty**, so `statedLimits` can be empty while `goalConstraints` is populated — and `not_assessed` (*"we could not check every limit you set on this run"*) is **exactly the state a malformed or withheld constraint produces**. The gate was quietest precisely where the producer was speaking loudest. One predicate answering two questions (trap 21): *"can we format the limits?"* vs *"did the user set limits the producer could speak to?"*

The row is now a **sibling** of the limits list rather than a child of it.

⚠ **The other direction is preserved deliberately.** PLoT auto-synthesises a `'Goal target'` constraint (`auto_goal_threshold`, `run.ts:6035-6042`) when the user set no limits, which can return `compliant` carrying *"this option met every limit **you set**"* — about limits nobody set. That synthesis happens inside PLoT's run handler and never reaches this store, so `goalConstraints` is genuinely empty there and the gate keeps it off screen. **Derived at PLoT staging `e19ac506` — the PLoT-side wording fix has NOT landed** (staging unchanged since this lane began), so the suppression test is load-bearing and says so in-file.

**Two-directional discriminating pair** — each way of breaking the predicate reds a *disjoint* set:
- revert the gate to pre-fix → **exactly 2 RED**, the two malformed-limit cases; all three suppression tests + control **GREEN**
- **Mutant C**, remove the gate entirely (the naive widening) → **exactly 2 RED**, the two suppression tests; both widening tests **GREEN** and the **discriminating control GREEN**, proving the reds are the gate and not the card breaking wholesale

Isolation re-proven by writing a sentinel with a positive control on the write; inodes differed; applied-check leading 0 / mutant 1 / trailing 0; markers remaining 0; source clone dirty 0. Regression **1807 passed / 0 failed** across 116 files, including #832's stated-limits suite and the primitive-selector contract. Lint 0 errors, hooks ratchet at baseline; typecheck ratchet **2252 = 2252**.

Three of the review's own findings strengthened the record and are adopted: PLoT's emission is more unconditional than I stated (no top-level returns, no try/catch, no flag in the span) and `CROWN_COMPLIANCE_REASONS` is **total over the union**, so a verdict cannot arrive without a reason; the withheld-crown projection survives because `projectLeaderDesignatingMembers` is a **keep-by-default deny-filter, not an allow-list rebuild** — the load-bearing fact, which I under-cited by crediting only the regex's closed suffix group; and the pin skew is safer than I said because **validation is not applied on the transport path at all** (`enrichment` is `z.record(z.unknown())`).

---

## ⚠ Routed, not fixed here — a CEE residual

`compose.ts:1017` drops any string leaf containing `[REDACTED]` **under any key**. `recommended_option_compliance_reason` is free prose: **if PLoT ever redacts into it, the key vanishes silently with no alarm**, and this card renders nothing where a verdict genuinely exists. Out of scope for a UI lane; flagged for the CEE owner.

---

## Not merging

This is for review. Please do not merge without an independent adversarial review — the seam it touches is a trust surface.

### DIFF
--- a/src/adapters/plot/v2/responseMapper.ts
+++ b/src/adapters/plot/v2/responseMapper.ts
@@ -571,6 +571,13 @@ export function mapV2ResponseToReportV1(
         // and falls back to this mapped slot, then normalises fail-closed.
         ...(typeof robustnessRaw?.display_verdict === 'string' ? { display_verdict: robustnessRaw.display_verdict } : {}),
         ...(typeof robustnessRaw?.display_verdict_reason === 'string' ? { display_verdict_reason: robustnessRaw.display_verdict_reason } : {}),
+        // Hard-constraint chain step 5 (producer PLoT #338): crown compliance
+        // verdict + producer reason pass through VERBATIM so saved/hydrated
+        // reports keep them — the render seam prefers the raw response and
+        // falls back to this mapped slot, then normalises fail-closed. Same
+        // treatment as `display_verdict` directly above.
+        ...(typeof robustnessRaw?.recommended_option_compliance === 'string' ? { recommended_option_compliance: robustnessRaw.recommended_option_compliance } : {}),
+        ...(typeof robustnessRaw?.recommended_option_compliance_reason === 'string' ? { recommended_option_compliance_reason: robustnessRaw.recommended_option_compliance_reason } : {}),
       }
     })() : undefined,
     // P0 Fix: Pass through robustness_status for gating logic
--- a/src/adapters/plot/v2/types.ts
+++ b/src/adapters/plot/v2/types.ts
@@ -338,6 +338,29 @@ export interface V2RobustnessActual {
    */
   display_verdict?: string
   display_verdict_reason?: string
+  /**
+   * CROWN COMPLIANCE (hard-constraint chain step 5; producer PLoT #338,
+   * staging `e19ac506`) — may the crowned option be badged as leading, given
+   * the limits the user actually stated? Emitted UNCONDITIONALLY on every
+   * /v2/run (`routes/v2/run.ts:3608-3609`), so ABSENCE here means an older
+   * producer build, never "no limits".
+   *
+   * Producer enum (`routes/v2/crown-eligibility.ts`):
+   *   not_applicable | compliant | uncertain | unverified | not_assessed |
+   *   no_eligible_option
+   * Typed `string` (not the union) at the trust boundary — same tagged
+   * passthrough convention as `display_verdict` / `level` / `label` above,
+   * and for the same reason: the vendored @talchain/schemas pin (0.48.0) does
+   * not know these members, and a future producer value must not crash an old
+   * UI build. `normaliseCrownCompliance` narrows FAIL-CLOSED at the render
+   * seam.
+   *
+   * The reason is the producer's own claim-safe phrase and is rendered
+   * VERBATIM — never re-authored here
+   * (`contracts/isl-to-ui.contract.ts:281`).
+   */
+  recommended_option_compliance?: string
+  recommended_option_compliance_reason?: string
   /** Tipping-point analysis: how much each factor must change to flip the recommendation */
   flip_thresholds?: Array<{
     node_id?: string
--- a/src/components/results/decision-overview/DecisionOverviewCard.tsx
+++ b/src/components/results/decision-overview/DecisionOverviewCard.tsx
@@ -34,6 +34,12 @@ import { computeGraphFacts } from '../../../canvas/components/pre-analysis-v3/se
 import { ActionsMenu } from './ActionsMenu'
 import { REVIEW_BRIEF_ASK } from './actionsCatalogue'
 import { formatStatedLimitsNote, parseStatedLimitsKey, selectStatedLimitsKey } from './statedLimits'
+import {
+  CROWN_COMPLIANCE_DOT_TONE,
+  selectCrownComplianceReason,
+  selectCrownComplianceVerdict,
+  selectCrownComplianceDisclosure,
+} from './crownCompliance'
 
 export type BriefStateOverride = 'thin' | 'contradictory' | 'unverified'
 
@@ -292,6 +298,17 @@ export function DecisionOverviewCard({ title, stateOverride }: DecisionOverviewC
   // drag, a producer re-sync — does not re-commit the whole card.
   const statedLimitsKey = useCanvasStore((s) => selectStatedLimitsKey(s.goalConstraints))
   const statedLimits = useMemo(() => parseStatedLimitsKey(statedLimitsKey), [statedLimitsKey])
+  // Step 6, second half — the PRODUCER's compliance verdict on the crown, read
+  // as two PRIMITIVES to keep the card's primitive-selector contract (U1), then
+  // composed. Fresh raw response wins over the saved/hydrated report; the
+  // narrowing is fail-closed, so an older producer or an unknown future token
+  // leaves the surface silent rather than guessing.
+  const crownComplianceVerdict = useCanvasStore(selectCrownComplianceVerdict)
+  const crownComplianceReason = useCanvasStore(selectCrownComplianceReason)
+  const crownCompliance = useMemo(
+    () => selectCrownComplianceDisclosure(crownComplianceVerdict, crownComplianceReason),
+    [crownComplianceVerdict, crownComplianceReason],
+  )
   const briefPresent = useCanvasStore((s) => Boolean(s.currentBriefText?.trim()))
   // Mirrors ResultsBody's UI-SEM-065 blocker read (graphHealth is the
   // engine-critique carrier; blocker severity never reaches uncertainties).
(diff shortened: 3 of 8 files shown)


Talchain:
## MERGE — step 6's second half. The chain now reaches a user.

`Staging Gate` **success** at `15b73c32`, 14 checks, `running=0`. Sole failure is `Visual Regression (advisory)` — verified pre-existing, measured failing on staging `39162243`, `5113f5a0` and `684b82a4` while `Staging Gate` passed on all of them, so the contrast discriminates.

**User-visible outcome:** where a user stated a hard limit, the results card now prints **the producer's own sentence** about whether the crowned option met it — including *"no option met the limits you set"* when the crown is withheld — instead of listing the limit and saying nothing.

### The named change, and why the gate had to move rather than disappear

The row was gated on **`statedLimits.length > 0`**, and `selectStatedLimits` (`statedLimits.ts:96-98`) **skips any constraint with a non-finite `value` or an empty `operator`**. So the limits list can be empty while the constraints exist.

⭐ **And the correlation was the sharp edge: `not_assessed` means *"we could not check every limit you set on this run"* — exactly what a malformed or withheld constraint produces. The gate was quietest precisely where the producer was loudest.**

One predicate answering two questions — *"can we format the limits?"* is not *"did the user set limits the producer could speak to?"*

**Now gated on `selectGoalConstraintCount(s) > 0`**, with the row a **sibling** of the limits list rather than a child.

### The other direction is preserved — and it was verified, not assumed

PLoT auto-synthesises a `'Goal target'` constraint when the user sets no limits, which can yield `compliant` with the reason *"this option met every limit **you set**"*. The old gate suppressed that by accident.

**The lane re-derived PLoT staging as `e19ac506` — unchanged, so the auto-synthesis fix has not landed — and confirmed at the bytes that `auto_goal_threshold` is synthesised INSIDE PLoT's run handler (`run.ts:6035-6042`) and never reaches the UI store.** So `goalConstraints` is genuinely empty in that case and the falsehood stays off screen. The suppression test is load-bearing and says so in-file.

**Two-directional discriminating pair, each direction REDing a disjoint set:**
- revert the gate → **exactly 2 RED** (both malformed-limit cases); all three suppression tests plus control GREEN
- remove the gate entirely → **exactly 2 RED** (both suppression tests); both widening tests GREEN **and the discriminating control GREEN**, so the reds are the gate rather than the card breaking wholesale

Isolation re-proven by writing a sentinel with a positive control on the write; applied-check leading 0 / mutant 1 / trailing 0.

### Corrections recorded rather than quietly fixed

**RED-first is 3/7, not 4/7.** The fourth failure was the lane's **own positive control failing on a fixture id it had guessed wrong** (`option_2` against the real `opt_hire`) — its own spec defect, not the defect under test. 7/11 reproduces exactly.

**And the support for one conclusion was thinner than the evidence allowed.** The withheld-crown projection was credited to a regex's closed suffix group; the load-bearing property is that it is a **keep-by-default deny-filter, not an allow-list rebuild** — had it been an allow-list, the suffix group would not have saved the fields. Right conclusion, better fact now cited.

### Both bounds stated rather than implied

**Rung is "deployed, flag-gate proven open" — NOT MOUNTED.** The flag gate is genuinely open (83 chunks discovered and 83 fetched, controls firing, the gate preceding all three render markers), but mounting also needs `!W && mt && Me` — runtime data conditions **no bundle grep can settle**. A MOUNTED rung needs a DOM witness this PR does not have.

**The CEE residual is routed, and its severity is honestly narrowed.** `stripInternalKeysDeep` drops any string leaf containing `[REDACTED]` **by value, under any key, at any depth, silently**. But the reason is a lookup into a **frozen six-entry table with no interpolation and no user data**, so **there is no path today by which a marker enters it**. It is a **latent hazard for future free-prose disclosures, not a live defect** — routed as a defect class (make the drop loud), not as a fix to one field.

Good judgement on the near-duplicate task chip: left in place rather than destroying another lane's work on a guess.

### Where the chain stands
Steps **1–5 CLOSED and deployed**; step 6 first half closed (#832); **this is the second half.** The remaining PLoT-side item — the auto-synthesis making `constraintsStated` true when the user set no limits — is commissioned separately and is **not** a blocker here, because that constraint never reaches the UI store.

PR #844, opened 2026-08-25
C11 guest copy — the draft banner no longer claims where an unsaved draft lives

Talchain:
Follow-on to **#841** (merged). **Body describes HEAD `02dca687`.**

## What #841 missed, and why

`GuestDraftImportBanner` read:

> "Save it to keep working on it — **otherwise it stays only in this browser**."

The same sentence #841 removed from the entry screen, false for the same reason: a guest's graph **also exists server-side**, so *"only in this browser"* tells the user nothing leaves their machine when it does. It is shown to someone who has just signed in, so it is live and reachable.

⚠ **#841's regression pin did not catch it.** The pin was scoped to one component's rendered output, so it only guarded the surface someone remembered to opt in. That is the hand-maintained-mirror defect, and it is exactly how this twin survived a PR written to kill this class.

## The fix is the discovery, not the detector

The guard now globs **tracked source** rather than being handed a list of surfaces — the same design move as `claimDrift/claimDriftWalker.ts`. The banned patterns have **one definition**, imported by both this sweep and #841's entry-page pin. Two copies of an absence assertion's own definition is trap 13 with a delay fuse.

## ⭐ It immediately found two more — and neither is fixed here

Both are **adjudicated with a written reason**, because blanket-fixing them would be wrong:

- **`DecisionRecordModal.tsx` is TRUE AS WRITTEN.** `persistenceNote` names a *split* — choice/confidence/expectation/review date to the account; rationale/assumption/revisit trigger local. That is a precise claim about **named fields**, not a blanket claim about the user's model, and its own docblock says it was written to claim neither more nor less than is true. Its `guestNote` twin has zero consumers — rowed.
- **`OnboardingOverlay.tsx` matches on the wrong half, and the right half is worse.** The match is about *share links* ("stay on your device unless you copy them out"), plausibly true. The sentence **before** it — *"Everything runs locally for single-user mode"* — is **materially false**: every turn goes UI → CEE → PLoT → ISL. That is a claim about product architecture, a bigger finding than this guard was built for. **Rowed, not fixed** — fixing it inside this seam would be the "while we're here" expansion the estate bans.

**The adjudication list cannot rot.** The sweep asserts every entry *still matches* and is still tracked, so a fixed or renamed file fails loud instead of quietly widening the hole.

## Discrimination, measured

| mutant | result |
|---|---|
| control | 0 applied, 6 green |
| reintroduce the sentence in the fixed file | **1 RED** |
| **new tracked file nobody opted in** | **1 RED**, naming file and pattern |
| stale allowlist entry | **2 RED** |
| harmless copy edit | GREEN |
| trailing control | 0 applied, 6 green |

## ⚠ A survivor I chased rather than explained away

The new-file mutant **first read GREEN**. Cause: `git ls-files` lists **tracked** files and the mutant's file was untracked. Re-run staged, it REDs.

So coverage is *tracked source* — the right scope for CI, where everything under review is tracked. The limit is now written into the helper rather than left as an overclaim, because "covered by construction" without the word **tracked** is precisely the scope-generalisation this estate keeps paying for.

Also stated in the helper: patterns are a **proxy** for "makes a promise about where a guest's work lives", not a proof. Runtime-assembled or server-loaded copy is invisible to a source sweep.

Lint clean. **Named typecheck gate PASSED** — ratchet identical at baseline (2252). 23 tests green across both specs.

Rung: **TESTED**.

### DIFF
--- a/src/components/auth/GuestDraftImportBanner.tsx
+++ b/src/components/auth/GuestDraftImportBanner.tsx
@@ -55,9 +55,17 @@ export function GuestDraftImportBanner() {
         <p className={`${typography.label} text-text-header`}>
           Save your draft to your account
         </p>
+        {/*
+          ⚠ NO CLAIM ABOUT WHERE THE UNSAVED DRAFT LIVES. This read
+          "otherwise it stays only in this browser", the same sentence removed
+          from the entry screen in #841 and false for the same reason: a guest's
+          graph also exists server-side, so "only in this browser" tells the user
+          nothing leaves their machine when it does. The actionable half was
+          always the true half, and it is what remains.
+        */}
         <p className={`${typography.bodySmall} mt-1 text-text-body`}>
-          You built a draft before signing in. Save it to keep working on it —
-          otherwise it stays only in this browser.
+          You built a draft before signing in. Save it to your account to keep
+          working on it.
         </p>
         {error && (
           <p className={`${typography.bodySmall} mt-2 text-danger`} role="alert">
--- a/src/pages/ScenarioListPage.tsx
+++ b/src/pages/ScenarioListPage.tsx
@@ -367,6 +367,14 @@ export default function ScenarioListPage() {
   // (secondary). Guest mode is the POC's primary flow and #/canvas works fully
   // as guest, so this branch must never be a dead end.
   //
+  // ⚠⚠ AND MY OWN REPLACEMENT WAS FALSE TOO. #841 put "Sign in to create a saved
+  // workspace." here. The mechanism is now settled: a guest's model is stored on
+  // OLUMI'S SERVERS and re-fetched every load from a 36-byte UUID pointer — so it
+  // is ALREADY saved, and an invitation to sign in "to save" tells the user their
+  // work is at risk when it is not. I fixed one falsehood and shipped another in
+  // the same sentence, which is why the licensed wording is now pinned centrally
+  // rather than reasoned out per surface.
+  //
   // ⚠ THE COPY MAKES NO CLAIM ABOUT WHERE GUEST WORK LIVES, deliberately.
   // A previous version read "Without an account, your work stays only in this
   // browser." Both halves were wrong at once:
@@ -390,7 +398,7 @@ export default function ScenarioListPage() {
             Olumi turns messy strategic work into a living visual model while keeping your judgement visible.
           </p>
           <p className={`${typography.bodySmall} text-text-light mt-3`}>
-            Sign in to create a saved workspace.
+            Sign up to keep your models across devices.
           </p>
           <div className="mt-6 flex flex-col items-center gap-3">
             <button
--- a/src/pages/__tests__/ScenarioListPage.spec.tsx
+++ b/src/pages/__tests__/ScenarioListPage.spec.tsx
@@ -57,7 +57,7 @@ function expectStrategicReasoningEntry() {
       'Olumi turns messy strategic work into a living visual model while keeping your judgement visible.',
     ),
   ).toBeInTheDocument()
-  expect(screen.getByText('Sign in to create a saved workspace.')).toBeInTheDocument()
+  expect(screen.getByText('Sign up to keep your models across devices.')).toBeInTheDocument()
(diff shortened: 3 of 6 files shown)


Talchain:
⛔ **DO-NOT-MERGE — hold on the claim list, requested by the Primary lane.**

The **machinery** in this PR (derived scoping, the adjudication list that cannot rot, the discriminating mutants) is not in question. The **claim list itself** is.

`GUEST_STORAGE_CLAIM_PATTERNS` pins what may be said about where a guest's work lives. The Primary lane is settling the guest-restore mechanism now, and the answer changes what this list should permit: **if restore turns out to be server-side, a guard that permits "saved locally" would pin a falsehood into the suite** — the exact failure mode this PR was written to end, one level up.

So: hold, deliberately and visibly, rather than merging on the assumption my patterns are right.

**What is NOT affected and stands on its own evidence:**
- the banner copy fix — *"otherwise it stays only in this browser"* is false regardless of how restore works, because a guest's graph exists server-side either way;
- the discovery-by-globbing design, and the finding that a per-surface opt-in is what let this twin survive #841;
- the two adjudicated entries (`DecisionRecordModal` true-as-written; `OnboardingOverlay`'s *"Everything runs locally for single-user mode"*, which is materially false and rowed separately).

I am marking this rather than quietly editing the list, so the hold is visible instead of implicit. Unblock expected in minutes, not hours.

Talchain:
### ✅ HOLD LIFTED — mechanism settled, claim list aligned, and it falsified copy I had already merged

The Primary lane settled it: **guest restore is server-side.** The model lives on Olumi's servers, re-fetched every page load from a **36-byte UUID pointer** in localStorage. The browser holds nothing else — largest stored value 47 bytes against a ~95 KB model, and **no `olumi-canvas-autosave` key exists on the deployed build.** Decisive control: a pointer-only transplant into a brand-new profile restored the model **5/5**; the same profile without the pointer gave 0 nodes and no graph call **4/4**.

That also reconciles the two readings that appeared to contradict each other. **One store, not two** — *"0 nodes after restart"* was a restart **inside** the 30–90 s write-back window; *"work survives restart"* was one **after** it.

### ⚠ It falsified my own merged copy

#841 replaced *"your work stays only in this browser"* with **"Sign in to create a saved workspace."** The work is **already saved**. Inviting the user to sign in *"to save"* tells them their work is at risk when it is not.

**Second time in this file that I fixed one falsehood and shipped another.** That is precisely why the licensed wording is now pinned **centrally** rather than reasoned out per surface. Replaced with the licensed *"Sign up to keep your models across devices."*

### ⭐ The sweep found a third instance — and my pattern was at fault, not the copy

`ServerVersionsSection` says *"Sign in to save **shared** versions"* — correctly scoped, and plausibly **true**: version capture is owner-scoped by DB design and short-circuits pre-RPC for guests.

**I did not tune the regex to exclude it.** A predicate over natural language that needs "just one more rule" is the four-round oscillation this estate has already paid for. Adjudicated with a written reason instead.

**And that surfaced a question I cannot answer.** The same sentence continues *"the local history above still works in this browser"* — a locality claim, when the browser holds 36 bytes. Either local version history uses a store nobody has named, or that half is false. **Rowed for the mechanism owner; not patched by me.**

### Two guard changes this round

- **A licensed-sentence positive control.** A ban list drifts towards forbidding everything, and **over-blocking reads as safe** — nothing fires, nothing logs, and a surface that says less looks tidier than one that lies. The sanctioned sentences now fail the suite if a pattern grows to eat them.
- **My own stale control was blessing the false sentence.** It asserted *"Sign in to create a saved workspace."* was clean. **Inverted, not deleted**, so the record shows this file once vouched for it.

Three bans added that a reasonable person would write, one of which I did: `sign in/up to save`, `create a saved workspace`, and `clearing browser data deletes your work` — the last deletes the **pointer**, not the model.

### Not asserted, deliberately

No privacy or exclusivity claim anywhere. A guest's model — including their brief in their own words — is readable by anyone holding the scenario UUID. **That is the founder's call**, and nothing here implies otherwise.

29/29 green · full-scope `eslint .` 0 errors · named typecheck gate PASSED at exactly baseline 2252 / 556.

PR #845, opened 2026-08-25
C11 — surface the producer's robustness caveat, gated on the owned leader claim [IN-class: cross-model review]

Talchain:
⚠ **CONTRACT CLASS: IN.** This changes what the product asserts about the user's ranking, so it wants **cross-model review**, unlike #840/#841/#844 which were presentation-class. Flagging rather than routing around the gate.

## What it closes

`robustness_caveat` had **zero production readers**. It carries the producer's own sentence about how far the ranking held — *"This ranking held up under the perturbations tested. That is not a guarantee — defaulted or uncertain inputs can still change the result."* — with a `basis` token licensing it.

Paired with #840's defaulted assumptions, a user can now see both **that a ranking may be fragile** and **which values Olumi had to invent**.

## The gate is the increment

`robustness_caveat` is one of **three leader-ranking members** CEE strips on a withheld turn. Per `ownedLeaderClaim`: *"Absence of the owned claim IS the withheld signal — there is no separate flag, by design."* The text presupposes a ranking exists.

The estate already shipped the defect next door: `deriveDecisionVerdict`'s **Authority 3** reconstructed a leader from win probabilities on a withheld run, so the hero printed *"X is slightly ahead"* beside CEE's own *"no option can be put forward yet"*.

So this surface **consumes** `hasLeadingOption` — *"the single boolean every surface must gate on"* — resolved once by `ResultsBody`. Presence is never read as evidence a leader exists; absence is never rendered as reassurance.

**The prop is required, not defaulted.** A default of `true` would silently re-open the claim for every future caller. It cost 5 type errors at existing call sites — the prop doing its job. Fixed at source; **`--update-baseline` not touched**; ratchet back to exactly 2252.

`basis` is required on the payload: a caveat with no stated basis is an unattested claim about the user's ranking.

## ⚠ My browser test was wrong, and the product was right

The first version asserted the caveat was **visible**, and failed at all three widths. Cause: the component resolves `deriveDecisionVerdict(report, { visibleOptionIds, ... })`, and the measure seeds **build-vs-buy** then replays **T3**, whose leader `opt_phased` is not on that canvas. Olumi correctly withheld; my gate correctly closed; I read a correct refusal as a failure.

My diagnostic probe disagreed with the component because it called the authority **bare**, without the options — two resolutions of one verdict differing by an argument, the two-authorities shape in the instrument.

Making it "pass" would have needed a matching graph **or a weakened gate** — and weakening the gate is the defect. The measure now witnesses the **invariant**: caveat on screen **iff** the verdict permits a leader.

**NAMED LIMIT:** all three viewports exercised the **withheld** arm. The permitted arm is **jsdom-only**. The arm is printed on every run so a product that silently stopped naming leaders cannot keep this green.

## Evidence

RED-first 8 collected by name / 6 failed. **46/46 green.** Lint clean. Named typecheck gate **PASSED**.

Discriminating pair verified by *which* test breaks:
- remove the gate → *"renders NOTHING when the leader claim is not permitted"*
- rename an adjacent group title → only a title-presence test

Plus: drop `basis` → RED · drop the prose identity proof → RED · unclassify in the anti-dark guard → **2 RED**, forcing `DECLARED_DARK` → `RENDERED_HERE` bookkeeping. Controls 0-applied and green at both ends.

Measured near-miss: the producer writes *"perturbations"*; `\bperturbation\b` is a banned glossary term that does **not** match the plural. **0 of 3 captured texts trip it, by one character** — pinned by test.

Rung: **TESTED**.

### DIFF
--- a/e2e/geometry/decisionBriefAssumed.measure.ts
+++ b/e2e/geometry/decisionBriefAssumed.measure.ts
@@ -121,7 +121,49 @@ for (const vp of VPS) {
     expect(new Set(signatures).size, `two groups rendered identical content: ${JSON.stringify(signatures)}`)
       .toBe(signatures.length)
 
-    // 5. Geometry, measured not inferred.
+    // 5. THE ROBUSTNESS CAVEAT — the INVARIANT, not one arm of it.
+    //
+    // ⚠ MY FIRST VERSION OF THIS ASSERTED THE CAVEAT WAS VISIBLE, AND FAILED —
+    // correctly. The component resolves the verdict as
+    //   deriveDecisionVerdict(report, { visibleOptionIds, rawHeadlineBanded })
+    // and this file seeds the BUILD-VS-BUY starter then replays the T3 capture,
+    // whose leader is `opt_phased` — an option that is not on that canvas. So the
+    // product withholds the leader claim, and the gate correctly closes. The
+    // product was right and the test was feeding it an incoherent state.
+    //
+    // Asserting "caveat visible" would therefore have required either a matching
+    // graph or a weakened gate. What is actually worth witnessing is the RULE:
+    // the caveat is on screen if and only if the live verdict permits a leader.
+    // That holds under either arm, and the arm reached is REPORTED so a silent
+    // drift to always-withheld cannot masquerade as a pass.
+    const permitted = await page.evaluate(async () => {
+      const w = window as unknown as { useCanvasStore: { getState: () => Record<string, unknown> } }
+      const report = (w.useCanvasStore.getState().results as { report?: unknown } | null)?.report ?? null
+      const nodes = (w.useCanvasStore.getState().nodes ?? []) as Array<{ id: string, data?: { kind?: string } }>
+      const mod = (await import(/* @vite-ignore */ '/src/lib/decisionVerdict.ts')) as {
+        deriveDecisionVerdict: (r: unknown, o?: unknown) => { hasLeadingOption: boolean }
+      }
+      return mod.deriveDecisionVerdict(report, {
+        visibleOptionIds: new Set(nodes.filter(n => n.data?.kind === 'option').map(n => n.id)),
+      }).hasLeadingOption
+    })
+
+    const caveat = section.getByTestId('decision-brief-robustness-caveat')
+    if (permitted) {
+      await expect(caveat, 'leader permitted, so the producer caveat must be on screen').toBeVisible()
+      await expect(caveat).toHaveText(brief.robustness_caveat.text)
+    } else {
+      await expect(
+        caveat,
+        'leader withheld, so a caveat ABOUT THE RANKING must not appear',
+      ).toHaveCount(0)
+      // And the absence must not be dressed as reassurance anywhere in the card.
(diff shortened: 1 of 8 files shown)


Talchain:
### Derivation for the `postRunFactorValue` ruling — posted here because the requesting session is no longer reachable

Primary asked: *does `postRunFactorValue`'s commit terminate in a `WIRE_SYSTEM_EVENT_TYPES` member with a CEE `'mutating'` classification?*

**YES, on both halves, derived at the bytes with contrast controls.**

**UI** — `factor_value_edit` is a `WIRE_SYSTEM_EVENT_TYPES` member, `src/canvas/conversation/types.ts:974` at `origin/staging`. Its comment names it a value-carrying **sibling** of `direct_graph_edit` (ROADMAP 1.346), deliberately not a field on it, because `direct_graph_edit.target_id` is *"a REPRESENTATIVE singular (the first changed id in a batch)"*.

**CEE** — `src/orchestrator-v5/system-events/dispatch.ts:270` at `ref=staging`:

```ts
factor_value_edit: 'mutating',
```

Contrast controls from the **same table**, so the classification discriminates rather than defaulting:

| event | classification |
|---|---|
| `patch_accepted` · `patch_dismissed` · `direct_graph_edit` · `chip_click` | `'ack_and_commit'` |
| `factor_value_edit` · `edge_strength_edit` · `structural_delete` | `'mutating'` |

`direct_graph_edit` being `'ack_and_commit'` is the relevant one: it is a turn row and **no graph write**.

It reaches a real writer, not an acknowledgement — `dispatch.ts:465-470`:

```ts
if (handling === 'mutating' && payload.event.kind === 'factor_value_edit') {
  return await dispatchFactorValueEdit(payload, payload.event, requestId, startedAt)
}
```

### ⭐ The part nobody asked for, and it splits the ruling

`dispatch.ts:433-439` degrades `'mutating'` to `'reader_only_refusal'` — **for `edge_strength_edit` only**:

```ts
const handling =
  payload.event.kind === 'edge_strength_edit' &&
  declaredHandling === 'mutating' &&
  config.features.graphCas.rpcEnforce !== true
    ? 'reader_only_refusal'
    : declaredHandling
```

**`factor_value_edit` carries no such gate.** It does not depend on `CEE_V5_GRAPH_CAS_RPC` — the single un-alarmed dashboard variable whose flip back to `shadow` silently returns `edge_strength_edit` to dark with no code change, no test and no red anywhere.

**So these two are not the same risk and should not share one decision.** `factor_value_edit`'s commit path is durable independently of that dashboard value; `edge_strength_edit`'s is not.

### Scope, stated precisely

I read `dispatch.ts` at `ref=staging` via the contents API. I did **not** read `dispatchFactorValueEdit`'s body (`factor-value-edit.ts`, 29 KB). *"Reaches a dedicated writer"* is a **routing** claim, not a claim about what that writer does on every branch. The independently reported fresh-guest journey witness (`graph_patch {"status":"applied"}`, `provenance: user_set` surviving reload, 3/3 trials) covers the happy path; neither of us has bounded its failure modes — and a nonsense value (`-999999999`) was reportedly applied without refusal, so the destination does not validate input.

PR #847, opened 2026-08-25
C11 — the ACT step: review a factor's value, gated on a usable destination [IN-class]

Talchain:
⚠ **CONTRACT CLASS: IN** — it changes what the product invites a user to do about their model. Cross-model review, not Opus review alone.

## What it closes

Resolve-next tells a user which unknown is most worth resolving — and then offered them **nothing to do about it**. NOTICE and EXPLAIN with no INTERVENE. This adds the act: one control per row, taking them to where that factor's value can be reviewed and replaced with their own judgement.

## Why this signal and not `defaulted_assumptions`

The obvious candidate is blocked **twice**, and I only knew about the first:

1. **Identity** — it carries `factor_label` and **no id** (26/26 fixtures). Binding to it means matching on a label, and a label is not a durable key.
2. ⭐ **Consequence** — a rerun after editing a factor the product shows as **"Not set"** *refuses* (`baseline_scale_unresolved`, 3/3 on a fresh-guest journey). That is exactly the defaulted/unset class. **It could never have closed the loop even if it had been bindable.**

`VoiRankingRow` already carries what the other lacks: `factorId`, a resolved canvas label (*"a row without one is dropped, never id-shaped"*), and `canFocus`. Identity already resolved, already validated, already fail-closed. **No label matching, no new producer contract, no wait on PLoT #340.**

## ⚠ The gate is the increment, and I was one step from shipping without it

`canFocus` proves a node **exists**. It does not prove that node has an **editor** — `FactorsSection` renders an inert *"Not set"* when the value is not a finite number. A control pointing there advertises an action terminating in nothing: the shape this estate has already shipped once.

So it gates on `canReviewValue`, decided by the **one** exported authority `factorHasConfirmableValue`, whose docblock exists because *five surfaces were answering this question four ways*. A sixth spelling would be the defect that module was created to end.

The gate also excludes precisely the class whose rerun refuses — **the control only appears where the loop can actually close.**

⚠ Note `canFocus` is **always true on the live path** (`nodeLabelMap` is built from canvas nodes, as that call site already documents). Gating on it would have gated on nothing.

## What the copy may claim

- **"Review this value" is provenance-neutral on purpose.** These rows rank by *value of information*, not defaultedness — the number may be one the **user** set — so *"review Olumi's estimate"* would be false on exactly those rows.
- **It promises no consequence.** The destination accepts input without checking plausibility (`-999999999` was committed without refusal in a live trial). Nothing claims Olumi validates the number, or that entering one improves the analysis.
- **It is separate from the crosshair**, which honestly focuses the canvas. `focusOnCanvasCopy.ts` is this estate's ruling that an "Edit" label on that handler was a **false promise**. The act navigates to the Model tab first — the one shipped results→Model-tab precedent.

## Evidence

RED-first: 7 collected by name, all failing at pristine. Wider area **275 files / 4856 tests, 0 failures**. Named typecheck gate **PASSED** at exactly baseline 2252/556; `--update-baseline` **not touched**.

| mutant | result |
|---|---|
| control | 0 applied, 7 green |
| **drop the gate** — offer on unset rows | **1 RED** |
| **own-object** — always send row 1's id | **1 RED** (proves identity binding) |
| provenance-claiming copy | **4 RED** |
| **caller ignores the authority** | **survived** → closed, see below |
| trailing control | 0 applied, 7 green |

**⚠ A mutant survived and it was a real hole.** Mutating the *caller* to `canReviewValue: true` left the component suite **7/7 green** — a component spec cannot see its caller, and the whole safety property lives there. Closed with a structural wiring guard (bounded to the resolver block); the mutant now REDs.

**⚠ And my "adjacent object" control never applied** (`applied-check 0`) — it mutated a string living in a different component. An unapplied mutation is indistinguishable from an equivalent one without an applied-check; the check caught it, not my reading of the result.

`canReviewValue` is **required, not optional** — it cost seven fixture updates across five specs, every call site now stating its answer. A default of `true` would silently license the control for every future fixture.

## Named limits

- The wiring guard cannot see a caller that imports the authority then ignores its result — that needs the hook under test, not proportionate here.
- **No browser witness on this branch.** Rung: **TESTED**.
- Whether a `factor_evppi` row can point at a canvas node that is unset is **unsettled** — the gate is correct either way, but it decides how often this control appears.


Talchain:
### ⚠ A coupling for whoever fixes `baseline_scale_unresolved` — this gate will need revisiting

Recording here because the reviewing lane is not reachable directly and this is easy to miss.

This PR's control is gated on `factorHasConfirmableValue` — the factor must carry a finite value or no control renders. I chose that gate for **two** reasons, both true today:

1. `FactorsSection` renders an inert **"Not set"** with no editor for such factors;
2. an unset factor drives the rerun into **`baseline_scale_unresolved`** (witnessed 3/3 on a fresh-guest journey), so the loop cannot close there anyway.

**A fix for (2) changes it without changing (1).** When that fix lands:

- if unset factors become **rerunnable but still uneditable** → this gate stays correct, nothing to do;
- if they also gain a **usable editor** → this gate becomes **too strict**, and it will suppress the control on exactly the rows where the intervention is most valuable.

⚠ **The failure mode is silent.** Nothing fires, nothing logs; the surface just looks like it has fewer rows worth acting on. That is the same shape as the glossary defect that cost ten of thirteen factor labels invisibly, which is why it is written down now rather than discovered later.

**I am not loosening it speculatively.** A control landing on a row with no editor is the dead end the gate exists to prevent, and guessing in that direction is worse than being one increment behind.

---

### Also relevant: `factor_id` has landed, verified emitted

PLoT staging `3a3bee58`, `buildDefaultedAssumptions` pushes `factor_id: f.factor_id` — **emitted, not merely typed**. Its docblock names the use precisely: *"a consumer can then tell 'this label IS the id' from 'this label is a real name', instead of guessing from the string's shape."*

That does **not** unblock an ACT step on `defaulted_assumptions`: it solves the *identity* half, while the *consequence* half stays blocked for exactly that class until `baseline_scale_unresolved` is fixed. It does unblock the separate live defect where an unprefixed id such as `hiring_cost` renders as though it were a human factor name — now detectable by identity rather than by guessing from string shape.

Talchain:
## Independent adversarial review — **MERGE WITH NAMED FOLLOW-UPS**, no blocking defect

Re-derived at `a91d64de` (the head moved mid-review from `36c85666`). 14/14 checks completed, 0 running, sole failure `Visual Regression (advisory)`.

### The gate premise is SETTLED, and the second correction is also wrong — same direction
**(a) is FALSE.** The new comment asserts `ModelTabBody` mounts both v2 and v1 `FactorsSection`, that "where v1 renders my original sentence is still literally true", and that the question "cannot be settled by reading source". **All three are refuted:**

- `ModelTabBody.tsx:120` — `const LEGACY_DETAILED_EDITOR_MOUNTED = false`
- the only `<FactorsSection` mount sits inside `{LEGACY_DETAILED_EDITOR_MOUNTED && (` (`:917`–`:1039`)
- three lines above it: *"B5: retained source, deliberately NOT mounted. The v2 outline above is the sole visible/editable Model route."*

It is a **source constant, not a deployed flag** — so it *is* settleable by reading source, the opposite of what the comment now tells the next reader. Deployed bundle corroborates (staging `01755479`, 82/82-chunk closed crawl): `model-tab-v1-stack` **0**, `LEGACY_DETAILED` **0** (tree-shaken) against `Change this value` **1**, `model-row-v2-` **11**. **v1 renders nowhere. There is no "except on v1" carve-out.**

**So the gate rests entirely on (b) — the rerun refusing `baseline_scale_unresolved` — which is what you now say, and that is correct.** Two things follow: the gate is **stricter than a "can the user act?" test would license today**, and when (b) is fixed it will suppress rows that *can* complete the loop, silently.

### F2 — the ACT step's wiring has ZERO coverage; it can be made entirely dark with the suite green
- **M7** delete `onReviewValue={onReviewValue}` at `AnalysisHeroContainer.tsx:191` → control never renders for anyone → **746/746 PASS**
- **M8** wrong destination tab → **746/746 PASS**

The estate already owns the instrument: `analysis-hero/__tests__/containerFocusWiring.spec.tsx`, whose header says it exists precisely because panel-level tests inject a `vi.fn()` above the container. **Extend it with the `onReviewValue` twin — that kills both.**

### F3 — the caller guard is defeated by a mutant satisfying its own text conditions
`canReviewValue: node ? factorHasConfirmableValue(node.data) || true : true` — imports the authority, calls it by name, evades the literal regex. **10/10 targeted and 746/746 green.** You named this limit; it is now **demonstrated**.

### What could NOT be broken
Identity binding proven by a **discriminating pair** (M3 misroute-all → RED; M4 misroute-unreachable-index → GREEN). Component gate M1/M2/M5 all RED. Collection asserted by name: 7/7 and 3/3, full sweep 40 files / 746 tests. **The required-field change is clean — the only production producer is derived, every other occurrence is a fixture answering `false` honestly; no call site answered wrongly to compile.** Copy claims no provenance, validation or improvement, and nothing reaches the >100%-ratio boundary. **The commit is canonical, not local-only.**

### Smaller, named
**F5** the route reaches the Model tab but not the row — v2 holds `selectedId` locally with no canvas subscription; the v1 scroll-into-view equivalent is in the unmounted branch. **F6** a suppressed row is byte-identical to a build without the feature — "we withheld this" and "no such control exists" are indistinguishable. **F7** below-resolution rows get no ACT step (pre-existing).

**Scope note, not a defect:** this puts no ACT step on the *"No starting value was provided"* notice — it binds to EVPPI `resolveNext.resolved`. Your reason is sound (`defaulted_assumptions` carries a label and no id, 26/26). Flagging so it isn't read as closing the defaulted-assumption gap.

### Follow-ups
1. Correct the comment — v1 is not mounted, the constant is source-level, the question is settled. **Keep the gate and the (b)-only justification.**
2. Extend `containerFocusWiring.spec.tsx` — kills M7 and M8.
3. Record F3 as *demonstrated* in the guard's header, or bind the caller with a hook-level test.
4. Consider selecting/scrolling the v2 row.
5. Disclosure for the suppressed case, **before** (b) is fixed.

*Reviewer limits: not journey-witnessed — deployed-bundle bytes and code, never rendered DOM. Reason (b) inherited, not re-measured.*

Talchain:
## Post-open derivation — the gate's *other* direction, and why it holds

The body reasons about one direction of the gate (rows with no finite value are excluded, because their rerun refuses). A reviewer should have the other direction too, because I did not state it and it is the one that could have made this a false promise.

### The question

`factorHasConfirmableValue` (`src/canvas/domain/valueProvenance.ts:351-356`) is exactly:

```ts
const value = (d?.observedState ?? d?.observed_state)?.value
return typeof value === 'number' && Number.isFinite(value)
```

So every row **with** a finite value gets the control. CEE **#1103** (open, `1963afff`) derives that the `baseline_scale_unresolved` refusal fires on values the **edit seam wrote**: `normalise-factor-value.ts:154-165` returns `{raw_value: x, value: x}` when capless + frameless, and `recoverScaleFrame` (`scale-frame.ts:39-52`) requires `raw > value` — so `raw === value` fails **by construction**.

That reads as a live false promise: a row that *has* a value, whose value came from the editor, would show "Review this value" and then refuse on rerun. **It does not, and the reason is upstream.**

### Why it holds — derived at ISL, not inferred from the UI

#1103's refusal carries an EVIDENCE conjunct: *"the factor's options must carry normalised interventions. A baseline can only be incoherent **with** something."* That is an **option-controlled lever**.

ISL's own contract states that `factor_evppi` **suppresses** exactly that class. From `openapi.json` (ISL @ `28fe0c95`), in the EVPC description:

> "EVPC is the value of CONTROL — the mirror of factor_evppi's value of INFORMATION … **Because control is the point, option-controlled levers are NOT suppressed here (unlike factor_evppi).**"

Contrast control: the same sentence is *discriminating*, not blanket — it names EVPC as the exception precisely because `factor_evppi` is the rule.

**So the refusing class and the ranked class are disjoint at the producer.** A row whose rerun would refuse for this reason never reaches this surface to be gated.

### The finding that follows, and it is not a comfortable one

**This PR's safety on that direction rests on an upstream suppression rule that nothing in this repo pins.** The gate does not enforce it; the tests do not assert it; the body did not know about it. If ISL ever stopped suppressing option-controlled levers from `factor_evppi` — the way EVPC already does not — this control would begin appearing on refusing rows, and **no guard anywhere would red.**

That is an unpinned precondition (CLAUDE.md 13b, third face: a guard whose discrimination depends on something nothing pins).

### What I am not claiming

- I have **not** driven this on the authenticated product. This is a producer-contract derivation, not a journey witness.
- I have **not** verified that `baseline_scale_unresolved` has no *second* trigger outside #1103's mechanism. #1103's table is byte-verified for the path it names; a different path to the same refusal would reopen the question.

### Suggested disposition

Not a blocker as I read it, but the reviewer should decide between:
- **(a)** merge as-is and row the unpinned precondition, or
- **(b)** add one test pinning the assumption — that the ranking cannot contain an option-controlled lever — so the dependency fails loud instead of silently.

I lean **(b)**: it is small, and the whole point of the gate was to avoid advertising an action that terminates in nothing.

⚠ Still owed regardless: this is Opus-authored and Opus-derived. **Cross-model IN-class review is not satisfied by this comment.**

Talchain:
## ⚠ Correcting my own comment above — it defends the wrong axis

An independent adversarial review returned **FIX-FIRST** with two HIGH findings. I re-derived both at the pinned tip `01755479` rather than adopting them. **Both mechanisms are confirmed.** My comment above is not wrong, but it is **beside the point**, and a reviewer should not spend time on it.

I argued the gate's second direction is safe because ISL suppresses option-controlled levers, so the `baseline_scale_unresolved` refusal class cannot reach this surface. That argument still holds **on its own axis** — but it is a *CEE/ISL rerun* argument, and the real defect is one hop earlier and entirely inside this repo.

### F1 (HIGH) — the gate reads `value`; the destination reads `raw_value`. Confirmed.

- Gate: `factorHasConfirmableValue` → `observedState.value` (`src/canvas/domain/valueProvenance.ts:351-356`).
- Destination: `getPrimaryValue` returns **`null` whenever `raw_value === undefined`** — it never reads `value` at all (`src/canvas/components/model-tab/utils.ts:64-72`, verified verbatim).

**Leg A — false promise.** A capped factor carrying `{value: 0.7}` with no `raw_value` passes my gate, so the control appears; the destination row has no value to show. That is the precise shape this gate was written to prevent.

**Leg B — silent gap.** A high-EVPPI factor with no `observedState.value` gets no control, yet `adapters.ts:448/463/479` mark factor rows `editable: true` and `proposeFactorValue` carries no confirmable-value precondition. The loop **can** close there; I hid the control on exactly the highest-information rows.

Two opposite harms under one predicate — they cannot share one window (CLAUDE.md 22b). And `valueProvenance.ts:337-345` **already documents this `value`/`raw_value` divergence**: I imported the function whose own docblock names the hazard without checking which side the destination reads.

### F2 (HIGH) — the destination does not resolve to the factor. Confirmed.

`onReviewValue` switches the dock tab and calls `focusModelTarget`, which acts on the **canvas** store, not the outputs dock. The only Model surface that reacts to canvas selection is the v1 `FactorsSection` scroll bridge — and **v1 is not mounted**: `src/canvas/components/ModelTabBody.tsx:120` `const LEGACY_DETAILED_EDITOR_MOUNTED = false`, wrapping the whole v1 stack at `:917`.

⭐ **This also settles a question I recorded as unsettleable.** I wrote that whether a user gets the v1 or v2 editor "cannot be settled from source — it needs a driven probe." **It is a `const … = false` on line 120.** That was not an honest UNKNOWN; it was an unchecked one.

### Two details in the review I did NOT confirm

- Its file paths omit the `src/canvas/` segment (`src/components/ModelTabBody.tsx` → `src/canvas/components/ModelTabBody.tsx`). Abbreviation, not fabrication — the files exist.
- It cites `ModelRowView.tsx:490` rendering `{display ?? 'Not set'}`. At this tip that is **past tense**: `:431` reads *"Every row **used to** render `display ?? 'Not set'`"* and `:204` says the row now shows **nothing**. The failure text differs; the divergence it rests on does not.

### Disposition

**I accept FIX-FIRST. Not merge-ready.** The fix must give the two harms **separate treatments**, not one retuned predicate:
1. bind `canReviewValue` to the projection the destination actually renders; and
2. give the no-value class its own honest affordance rather than silence — that is where value-of-information is highest.

F2 needs the destination to actually arrive at the row.

Cross-model IN-class review still owed, and now more so — the review that caught this was Opus, i.e. the same model that wrote the defect.

PR #849, opened 2026-08-25
fix(canvas): gate boot's analysis verdict on graph ACCEPTANCE — a refused graph must not become the wire authority (A3)

Talchain:
## The defect — LIVE on staging at `01755479`

At boot the UI wrote CEE's analysis verdict into `analysisStateV1` **before it knew whether the graph carrying that verdict had been accepted**. In `serverGraphHydration.ts` the call sat at **:192** and the `merge.accepted` gate at **:242** — fifty lines apart.

`analysisStateV1` is **feature-detected** by the selector: a non-null value takes the **WIRE branch**, where the local dirty overlay is not consulted (`analysisStateSelector.ts:551-554`). So when the merge refused the server's graph, that refused graph's verdict still became **authoritative over the user's own local graph** — the graph the refusal exists to protect.

The user was then told **"Model changed since this analysis"** about a model that analysis never ran on. This is a System-A truth defect: the product asserts something false about the user's own model, which is worse than saying nothing.

**Trigger** (Codex, reviewing `2c83021b`, re-evaluated on deployed merge `55807813`): authenticated reload + non-empty local canvas + valid zero-overlap server graph + `complete_stale`.

## Blast radius is wider than the trigger — three reachable legs, not one

`mergeServerGraphOnHydrate` refuses on four named reasons. I measured which are reachable through this leg:

| Refusal | Reachable here | Why it produces the same lie |
|---|---|---|
| `zeroOverlap` | **yes** | Codex's trigger. Two unrelated graphs; the verdict describes *theirs*, the canvas holds *ours*. |
| `importUnregistered` | **yes** | **The sharpest case.** The canvas holds an import the server has **never seen** (ROADMAP 2.467/2.503), so CEE's verdict is *necessarily* about an older model. Not an edge case — a guaranteed mismatch. |
| `emptyServerGraph` | **yes** | A verdict about a graph CEE does not have. |
| `unusableShape` | **no** | The adapter filters a non-object `graph` at `scenarioGraph.ts:267,274` and returns `'unusable'` **before this call site exists**. |

Fixing only `zeroOverlap` would have left two legs producing the identical user-visible lie, with the fix's own tests unable to see the survivor.

**`unusableShape` was measured, not assumed.** My spec's first draft asserted `mergeRefused` for it and RED-ed with `'unusable'` — that discrimination is what established the boundary. It is pinned as an explicit test rather than deleted, so an adapter change that later makes the fourth leg reachable turns that test red instead of passing silently.

## The remedy — and a material correction to the prescribed one

Codex prescribed *"apply the analysis verdict only after graph acceptance."* **Taken literally that drops the commonest boot of all.**

The `unchanged` short-circuit returns **before the merge runs**, so a literal "move it below `merge.accepted`" would leave the verdict absent on every re-boot of an unmoved scenario — exactly the sessions the original placement note was right to worry about, and exactly when a user is most likely to open the panel. That would trade a false verdict for a missing one: the mirror harm.

`unchanged` **is** an accepted path, for a structural reason rather than a conventional one: the identity token it matches on is recorded **only after a merge was accepted** (the `!merge.accepted` return precedes `setServerGraphIdentity`). A token match is therefore **proof of a prior acceptance** of that exact server graph under that exact projection version.

So the implemented rule is **acceptance, not position**: the verdict is applied at each of the two exits that represent an accepted graph (`merged`, `unchanged`), and at neither refusal. The code change is small — the existing call is wrapped in a `restoreVerdict()` closure and invoked at those two exits.

A refusal writes **nothing** — not `null`. Writing `null` would replace a standing belief with a claim of ignorance, a second falsehood rather than the absence of the first (`applyScenarioAnalysisRead.ts:402-405` draws the same distinction on the decline side).

### The behavioural change is provably confined to one exit

Not an assurance — a complete manifest. `hydrateCanvasFromServer` has **ten** exits. The old call site sat at `:192`, so only the exits **below** it could ever have written a verdict, and there are exactly three:

| Exit | Old | New |
|---|---|---|
| `skipped` (bad id), `absent`, `notReadable`, `unavailable`, `refused`, `unusable`, `skipped` (scenario moved) — 7 exits **above** `:192` | no verdict | no verdict — **untouched** |
| `unchanged` (`01755479:209`) | verdict written | verdict written — **unchanged** |
| `merged` (`01755479:262`) | verdict written | verdict written — **unchanged** |
| `mergeRefused` (`01755479:247`) | **verdict written** | **nothing written** ← the whole change |

So `mergeRefused` is the *only* exit whose behaviour moves. That is the argument the accepted-positive tests exist to hold up, and it is checkable rather than asserted.

**No `HydrationOutcome` value changes.** The fix removes a *side-effect*, never a return value — `mergeRefused` is still returned on exactly the same conditions. So every consumer that switches on the outcome is untouched by construction, including `draftRecovery.ts`, which maps `mergeRefused` to `notRecovered` and is unaffected either way. `absentGraphRetry.ts` and `draftRecovery.ts` have **zero** changes in this PR.

## RED-first, by name, at pristine `01755479`

New spec `src/canvas/hydrate/__tests__/bootVerdictGraphAcceptance.spec.ts` — **12 tests collected** (asserted by name and exact count, not inferred from a suite total).

**Pristine: 4 failed | 8 passed.**

| RED at pristine | Signature |
|---|---|
| `zeroOverlap` | `expected { run_state: {…} } to be null` |
| `importUnregistered` | `expected { run_state: {…} } to be null` |
| `emptyServerGraph` | `expected { run_state: {…} } to be null` |
| a refusal does not DESTROY a standing verdict | `expected '2026-08-25T09:00:00.000Z' to be '2026-08-24T08:00:00.000Z'` |

**After the fix: 12 passed.** The 8 that were green at pristine stayed green.

## Both directions, deliberately

One predicate here guards two opposite harms, so both doors are pinned:

- **refusal-negative** — a refused graph's verdict must not land (the lie this closes);
- **accepted-positive** — an accepted graph's verdict **is** applied, unchanged from today (re-opening #842's gap is the mirror harm). `merged`, `unchanged` and the **idempotent** `accepted: true, changed: false` case are all named tests.

Plus a **contrast control**: the *same* verdict IS restored when the graph is accepted. Every refusal assertion claims an absence; the contrast reads PRESENT in the same run, so a null cannot be mistaken for a fixture that never worked.

`complete_current` is still declined on an accepted merge, and the 404 leg still writes nothing — the reordering does not weaken the kind-level declines.

## Discriminating mutant pairs

Worktree **outside** the repo root (absolute path to `worktree add`), mutating only **committed** state, applied-check scoped to `src/` with the baseline control reading **exactly 0**, clean-tree assertion before and after each mutant, restores via `git checkout HEAD --` (never `git checkout --`, which restores from the index), and a trailing control. **Isolation proven by writing** a sentinel into the worktree and asserting the source's SHA-256 was unchanged; inodes compared (`718488056` vs `718491950`).

Every run asserts **12 collected** — a shrunk collect is a hard error, not a pass.

| Mutant | RED | Proves |
|---|---|---|
| CONTROL (no mutation) | none (12 green, applied=0) | baseline |
| **M1** drop restore at MERGED exit | contrast control, `merged`, idempotent | the merged exit binds |
| **M2** drop restore at UNCHANGED exit | `unchanged` **only** | the unchanged exit binds **independently** — M1/M2 are the pair |
| **M3** REVERT to the pre-fix position | **7 tests**, incl. all 4 original REDs | reverting the fix turns the suite red |
| **M4** gate on `merge.changed` not `.accepted` | idempotent **only** | gating on the right field |
| **M5** loosen for ALL refusals | all 4 refusal tests | the gate is what closes all three legs |
| **M6** loosen for `zeroOverlap` ONLY | `zeroOverlap` (+ the standing-verdict test, which uses that fixture) | per-leg binding |
| **M7** loosen for `emptyServerGraph` ONLY | `emptyServerGraph` **only** | per-leg binding |
| **M8** loosen for `importUnregistered` ONLY | `importUnregistered` **only** | per-leg binding |
| TRAILING CONTROL | none (12 green, tree clean) | no mutation leaked |

M6/M7/M8 are the discriminating set: loosening **all** legs REDs each test (M5), loosening a **different** leg leaves it GREEN. Each test binds to its own leg, not to a shared predicate.

## Named limits — what I did NOT prove

- **No live staging witness.** Everything here is local and wire-level. The rung reached is TESTED, not JOURNEY-WITNESSED. The defect's live status rests on the deployed tip matching this source, not on my having driven it.
- **jsdom proves no visibility.** These are store-level and outcome-level assertions. That the *rendered* panel changes its sentence is **not** shown here — only that the wire authority the panel reads is no longer poisoned.
- **`unusableShape` unreachability is a claim about this leg through this adapter**, at this tip. It is not a claim about `mergeServerGraphOnHydrate`'s other callers.
- **The write-log instrument is test-local.** It proves the boot leg does not call `setAnalysisStateV1`; it does not prove no other writer touches the field during boot.
- The three reachable legs are the three I could reach **from the wire plus the store**. I did not enumerate refusal causes reachable only through states I did not construct.

## Two instrument defects caught and disclosed

Both were caught by controls, not inspection, and neither would have errored:

1. **The write-log wrapper re-wrapped itself.** Reading `setAnalysisStateV1` inside `beforeEach` wrapped the previous test's wrapper; because every layer closed over the same module-level binding, the write count tracked the **test index** (read 8 and 9 for a leg that writes at most once). Only an **exact-count** assertion exposed it — `toHaveLength(1)` rather than "at least one". Fixed by capturing the pristine action once at module load; the reason is written into the spec so it cannot be re-introduced quietly.

2. **The first mutation harness could not report a failure at all.** Its `sed 's/.*spec.ts > //'` stripped the `✓`/`×` marker along with the path, so the RED list was **always empty** and all nine mutants reported "RED: (none)" — *including a full revert of the fix*. The collected-count assertion still read 12, so nothing errored and every mutant agreed with every other. It was caught only by the **uniformity**: a per-item probe returning an identical answer for every item is reporting on itself. Fixed to preserve the mark; the numbers above are from the corrected harness.

## Salvage disposition — ADAPT-SHAPE-ONLY, not adopt

`salvage/review842-adversarial-spec-2026-08-25` @ `abe2cc24` was **not** adopted, and two things in the brief describing it need correcting for the next reader:

- The brief's "9 files / +1901" is the **branch-vs-stale-base** diffstat (base `fa2113aa`, 8 commits behind tip). `abe2cc24` itself is **spec-only, 116 lines**.
- The spec is **not usable as a RED-first basis**. Its test 1 asserts the **buggy** behaviour, so it *passes at pristine* — a test that passes before the fix exists is not a test. Its test 2 asserts a `blocked` verdict becomes the wire authority, which is **false at this tip**: `BOOT_RESTORABLE_RUN_STATE_KINDS` is `['complete_stale']` only, so `blocked` is already declined upstream as `not_restorable`.

I took its *shape* (the fixture and attack framing) and wrote a fresh inverted spec.

## Gate

Run in the required check's real step order, derived at this tip — `lint → typecheck → guards → 4 vitest shards → build` — not a hand-rolled subset.

## Boundaries

Touches only `src/canvas/hydrate/`. No files under `components/results/**`, `decision-brief/**`, `mapV5AnalysisToReport.ts`, `responseMapper.ts`, `ScenarioListPage.tsx`, and no CEE files.

### DIFF
--- a/src/canvas/hydrate/__tests__/bootVerdictGraphAcceptance.spec.ts
+++ b/src/canvas/hydrate/__tests__/bootVerdictGraphAcceptance.spec.ts
@@ -1,0 +1,482 @@
+/**
+ * BOOT'S ANALYSIS VERDICT IS GATED ON GRAPH ACCEPTANCE — the A3 truth spec.
+ *
+ * ═══════════════════════════════════════════════════════════════════════════
+ * THE DEFECT, DERIVED AT `01755479` (deployed staging tip)
+ * ═══════════════════════════════════════════════════════════════════════════
+ * `hydrateCanvasFromServer` calls `applyBootAnalysisVerdict`
+ * (`serverGraphHydration.ts:192`) BEFORE it knows whether the graph that
+ * carried that verdict was accepted — the `merge.accepted` gate is fifty lines
+ * further down (`:242`). So on every boot where the MERGE REFUSES, CEE's
+ * verdict is still written into `analysisStateV1`.
+ *
+ * That store field is FEATURE-DETECTED by the selector: a non-null
+ * `analysisStateV1` takes the WIRE branch, where the local dirty overlay is not
+ * consulted (`analysisStateSelector.ts:551-554`). So the refused server graph's
+ * verdict becomes AUTHORITATIVE over the user's own local graph — a graph it
+ * has, by the refusal's own definition, nothing to say about.
+ *
+ * The user is then told "Model changed since this analysis" about a model the
+ * analysis never ran on. This is a System-A truth defect: the product asserts
+ * something false about the user's own model.
+ *
+ * ═══════════════════════════════════════════════════════════════════════════
+ * THE FOUR REFUSALS, AND WHY EACH ONE MATTERS HERE
+ * ═══════════════════════════════════════════════════════════════════════════
+ * `mergeServerGraphOnHydrate` refuses on four named reasons
+ * (`mergeServerGraph.ts:103-131`), and this spec exercises ALL FOUR rather than
+ * only the one Codex's trigger names — a fix aimed at `zeroOverlap` alone would
+ * leave three live doors, and the predicate guards one property, not one case:
+ *
+ *   `zeroOverlap`         Codex's trigger. Two unrelated graphs; the server's
+ *                         verdict describes THEIRS, the canvas holds OURS.
+ *   `importUnregistered`  The sharpest one. The canvas holds an import the
+ *                         server has NEVER SEEN (ROADMAP 2.467/2.503), so
+ *                         CEE's verdict is necessarily about the older model.
+ *   `emptyServerGraph`    CEE has no graph; a verdict about nothing.
+ *   `unusableShape`       Not a graph at all.
+ *
+ * ═══════════════════════════════════════════════════════════════════════════
+ * ⚠ BOTH DIRECTIONS, DELIBERATELY (trap 22b)
+ * ═══════════════════════════════════════════════════════════════════════════
+ * One predicate here guards two OPPOSITE harms:
+ *
+ *   too narrow → a refused graph's verdict still lands  (the LIE this closes)
+ *   too wide   → an ACCEPTED graph's verdict is dropped (re-opens #842, the
+ *                defect whose fix this is layered on top of)
+ *
+ * A fix that closes the gap must be re-measured against the defect it was
+ * written to close, so the accepted-positive block below re-asserts #842's
+ * `merged` AND `unchanged` restores with the SAME fixtures. They are not
+ * decoration: they are the other door.
+ *
+ * ⚠ `unchanged` IS AN ACCEPTED PATH, AND THAT IS THE SUBTLE PART. It
+ * short-circuits BEFORE the merge runs, so "apply after `merge.accepted`" read
+ * literally would drop it. It is accepted because the identity token it matches
+ * on is only ever recorded AFTER a merge was accepted
+ * (`serverGraphHydration.ts:253`, gated on `!merge.accepted` returning above) —
+ * so a token match is proof of a PRIOR acceptance of that exact server graph.
+ * Pinned here so that reasoning is a test rather than a comment.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import type { AnalysisStateV1 } from '@talchain/schemas/boundary'
+
+import { useCanvasStore } from '../../store'
+import { hydrateCanvasFromServer } from '../serverGraphHydration'
+
+const SCENARIO_ID = '11111111-2222-4333-8444-555555555555'
+const CEE_TOKEN = 'c'.repeat(63) + '9'
+
+/** The exact instant CEE stamped the verdict — the IDENTITY this spec binds to. */
+const COMPUTED_AT = '2026-08-25T09:00:00.000Z'
+
+function envelope(value = CEE_TOKEN, projection = 'identity.v1') {
+  return {
+    kind: 'graph_identity_hash',
+    value,
+    algorithm: 'sha256',
+    projection_version: projection,
+    graph_schema_version: 'graph_v3',
+    normaliser_version: '1',
+  }
+}
+
+function verdict(
+  runState: AnalysisStateV1['run_state'],
+  over: Partial<AnalysisStateV1> = {},
+): AnalysisStateV1 {
+  return {
+    run_state: runState,
+    readiness: { status: 'ready', blockers: [] },
+    leader_claim: { permitted: false, withheld_reason: 'separation_unavailable' },
+    robustness: {},
+    usable_for_prose: false,
+    usable_for_chips: false,
+    usable_for_followup: false,
+    requires_rerun: true,
+    blocked_unusable: false,
+    contradictions: [],
+    ...over,
+  } as AnalysisStateV1
+}
+
+/**
+ * `complete_stale` is the ONLY boot-restorable kind
+ * (`BOOT_RESTORABLE_RUN_STATE_KINDS`), which is what makes it the right fixture:
+ * every other kind is already declined upstream by `applyBootAnalysisVerdict`,
+ * so a spec built on one of those would pass at pristine for the WRONG REASON —
+ * the verdict would be absent because the KIND was declined, not because the
+ * graph was refused. Codex's trigger names this kind for exactly that reason.
+ *
+ * ⚠ `cause`, NOT `stale_cause` — the contract schema is `.strict()` and the
+ * wrong spelling arrives as `null`, indistinguishable from the fix working.
+ * The positive control below is what proves it parses.
+ */
+const STALE_VERDICT = verdict({
+  kind: 'complete_stale',
+  computed_at: COMPUTED_AT,
+  cause: 'graph_changed',
+})
+
+/** A server graph that SHARES node ids with the seeded canvas — merge ACCEPTS. */
+function overlappingGraph() {
+  return {
+    nodes: [
+      { id: 'factor-1', kind: 'factor', label: 'Spend', value: 250 },
+      { id: 'goal-1', kind: 'goal', label: 'Profit', value: 9 },
+    ],
+    edges: [],
+  }
+}
+
+/** A server graph sharing NO node ids with the seeded canvas — merge REFUSES. */
+function unrelatedGraph() {
+  return {
+    nodes: [
+      { id: 'unrelated-a', kind: 'factor', label: 'Other' },
+      { id: 'unrelated-b', kind: 'goal', label: 'Elsewhere' },
+    ],
+    edges: [],
+  }
+}
+
+function okBody(over: Record<string, unknown> = {}) {
+  return {
+    schema: 'scenario_graph.v1',
+    scenario_id: SCENARIO_ID,
+    graph: overlappingGraph(),
+    graph_present: true,
+    brief_text: null,
+    graph_identity_hash: envelope(),
+    layout_present: false,
+    request_id: 'req-boot-acceptance',
+    ...over,
+  }
+}
+
+function jsonResponse(status: number, body: unknown): Response {
+  return {
+    ok: status >= 200 && status < 300,
+    status,
+    json: async () => body,
+  } as unknown as Response
+}
+
+/**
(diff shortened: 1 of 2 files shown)


Talchain:
## Discharging a disclosed limit: the complete writer manifest for `analysisStateV1`

The PR body names a limit — *"the write-log proves the boot leg does not write, not that no other writer touches the field."* That limit is now **measured** rather than merely disclosed.

**Method:** fresh blobless clone at `56a89a60`, `rg -a` (binary-safe, so the NUL-blindness trap cannot apply), `src/` excluding tests.

Every occurrence of `analysisStateV1` in non-test `src/`:

| Location | What it is |
|---|---|
| `store.ts:521` | type declaration — not a write |
| `store.ts:1986` | initial state — `null` |
| `store.ts:2996` | `importCanvas` — clears to `null` |
| `store.ts:3335` | `resetCanvas` — clears to `null` |
| `store.ts:4342` | `loadScenario` — clears to `null` |
| `store.ts:4891` | **`setAnalysisStateV1`** — `set({ analysisStateV1: state })` |

**Exactly one writer can set a non-null verdict:** `setAnalysisStateV1` (`store.ts:4890-4892`) — which is precisely the action this PR's write-log wraps. Nothing performs a direct `setState({ analysisStateV1: <verdict> })` that would bypass it. The other three writes are deliberate clears on graph replacement, which is the **safe** direction: a clear cannot introduce a false verdict.

**Contrast control, same run:** 9 non-test files mention the field — non-zero, so the probe was not blind. An absence claim with a contrast that reads present is a measurement; without one it is indistinguishable from a broken probe.

### ⚠ Two claims, two instruments — please do not collapse them

This is the load-bearing part, and it is stated against my own interest, because collapsing them would make this PR look stronger than it is:

- **"The boot leg does not write on a refusal"** — proved by the **mutants** (M5–M8 loosen the gate per-leg and each REDs only its own leg).
- **"Nothing else writes a verdict"** — proved by the **manifest above**, *not* by the write-log.

The write-log watches the **action**. It would be **blind** to a future writer that bypasses it via direct `setState`. That is exactly why the second claim is carried by an enumeration rather than by the instrument. Attack the manifest, not the write-log, if you want to falsify the second claim.

**Scope, stated precisely:** `src/`, excluding tests, at this SHA. Not a claim about test code, other repos, or any writer reachable only through code paths this sweep could not see.

Talchain:
## Independent adversarial review: APPROVE — with four corrections to the record

Merged on this verdict. Recording the corrections here because three of them make claims in this PR's own body **more accurate rather than less**, and the next reader should inherit the corrected version.

Reviewed at `56a89a6087db5caf443bf494d2e6d401d132c223`, verified via `git ls-remote refs/pull/849/head` against `gh api`. Merge base was the staging tip `01755479` exactly.

### All four claims survive
- **Three reachable refusal legs** — the refusal domain was re-derived independently, not accepted: `mergeServerGraph.ts` has exactly one `refused()` helper with exactly four call sites (`unusableShape:194`, `emptyServerGraph:203`, `importUnregistered:223`, `zeroOverlap:245`) and one `accepted: true` site (`:394`), so `!merge.accepted` ⟺ one of four, complete. `unusableShape` is genuinely unreachable through this leg (`parseOk` returns `{status:'unusable'}` at `scenarioGraph.ts:274,281` before the graph branch exists). Pinning it as `'unusable'` rather than deleting the case is a correct tripwire — an adapter change REDs it.
- **Accepted paths preserved** — both families plus the idempotent `(accepted:true, changed:false)` case.
- **No `HydrationOutcome` value change** — union byte-identical, 9 members, both extractions asserted non-empty first so this is not two empty files agreeing.

### Corrections
1. **The writer manifest was incomplete.** A seventh site exists: `store.ts:6043`, `updates.analysisStateV1 = null` in `hydrateGraphSlice` — property-assignment form, invisible to a `analysisStateV1:` key grep. It writes **null**, so *"exactly one writer sets a non-null verdict"* stands. The cause is syntactic-form blindness in the sweep, which is worth knowing for the next enumeration.
2. **"The revert REDs 7" describes the naive double-write revert.** A *faithful* revert REDs **4**, and 4 is the honest "this fix closes the defect" figure. Both mutants are legitimate; 7 overstates the blast radius.
3. **"One exit moves" is understated, not wrong — two behaviours move.** The `merged` exit's *ordering* inverts, and an eleventh, abrupt exit (a throw from the merge) previously left a verdict written and now does not. Both are safe-direction, and the code comment is honest about the ordering; only the summary was not.
4. The idempotent test binds by `kind` rather than `computed_at` and carries no write-count assertion. It bites M2/M4, so it is not vacuous — just inconsistent with its siblings.

### The stated limit is DISCHARGED, not merely disclosed
The PR disclosed that its write-log watches the store action and so could not speak to other writers. The reviewer settled it: a real second non-null writer exists at `applyScenarioAnalysisRead.ts:243`, reached from `useProvisionalAnalysisDelivery` — and **the boot leg cannot arm it**, because it arms exclusively on `run_state.kind === 'running'` and `running` sits in `BOOT_DECLINED_RUN_STATE_KINDS`, so boot can only ever restore `complete_stale`. No direct-`setState` bypass and no persist-rehydration path exists; the three clears are unconditional. Controls: 49 non-test `useCanvasStore.setState(` sites and 23 non-test `analysisStateV1` occurrences, both probes demonstrably sighted, intersection empty.

### Mutants re-derived independently
Nine runs with a JSON reporter (no marker-stripping `sed`), applied-check scoped to `src/`, control reading exactly 0, restore from `HEAD`, clean tree asserted before and after, isolation proven by writing a sentinel with distinct inodes. `M0=0 · M1=4 · M2=1 · M3=1 · M4=3 · M5=4 · M6=2 · M7=1 · M8=1` — **every count predicted before the run and matched**, and the counts are non-uniform, which is the discrimination a blind instrument cannot fake.

### Reviewer's own disclosed limit
`R7` is **not** discriminating: under M5 it stays green because `complete_current` is declined at the kind level before reaching the store, so two independent declines produce one observable. Reported rather than counted.

### Rung, stated plainly
**TESTED, not journey-witnessed.** jsdom shows the store field is no longer poisoned; it does not show what the user reads on screen. The rendered sentence was not exercised, and no live staging witness was taken before merge.

### Rowed, not fixed here
The provisional-delivery seam (`applyScenarioAnalysisRead.ts:243` via `useProvisionalAnalysisDelivery`) applies CEE's verdict with **no graph-acceptance gate** of its own. It is unreachable from boot, but reachable after a refused boot *plus* a new run — the same harm class as this PR, through a different door. Tracked as follow-up work, not smuggled into this change.

PR #851, opened 2026-08-25
fix(auth): the scenario routes carry a verified token, so CEE can stop trusting the body

Talchain:
UI half of the scenario-ownership fix. **Review gates 1 and 2 addressed.**

## ⚠⚠ WHAT THIS DOES AND DOES NOT CLOSE — read before reviewing

**This is the necessary PRECONDITION for the CEE strip, not the fix.**

An earlier version of this body called it "additive and guest-safe"; a later one over-corrected and implied it closes the forgery. Neither is right.

- **CEE already prefers the verified `sub`** (`route-v2-preflight.ts:188-206`, deployed) — gated only on a token arriving. So the moment this deploys, a signed-in user's ownership is decided by their token subject.
- **But the forgery stays open.** `netlify/edge-functions/cee-proxy.ts` injects the assist key **unconditionally, with no JWT gate**. An anonymous browser that simply omits the header still reaches `service_legacy` and forges via body `user_id`. **The probe table below still returns 200 after this merges.** Closing it is CEE-side (#1109).

Two derivations that looked contradictory answer **different questions**: `:188-206` answers *"token present, who wins?"* → the JWT sub. The strip answers *"token absent, can the caller forge?"* → yes, still. Both true at once.

## The problem, measured

`CEE_REQUIRE_USER_JWT` is **ON**: boot log `"require_user_jwt": true`, and a JWT-shaped invalid Bearer answers `401 {"validator":"user_jwt"}` — reachable only with the flag on. These adapters sent no `Authorization` at all, so every signed-in caller resolved `service_legacy`, where the body `user_id` is the only identity.

| probe (deployed staging) | result |
|---|---|
| **anonymous, no token, `user_id` = owner** | **200 — full graph, brief and analysis; canary present** |
| anonymous, *different* `user_id` | 404, no canary |
| anonymous, **no** `user_id` | 404, no canary |
| anonymous, malformed `user_id` | 404, no canary |

The same channel let an anonymous caller **create** a scenario attributed to an arbitrary id.

## Merge gates addressed

**Gate 1 — a sixth call site.** `useProvisionalAnalysisDelivery.ts` passed the adapter *by reference* (`read: fetchScenarioGraph`), invisible to the grep that found the other five. It sent `user?.id ?? null` and no token. Fixed, and `accessToken` is now **required** on the schedule's deps so it cannot be omitted silently.

**And the enumeration mirror behind it.** The five-route table was hand-maintained — a sixth route would get no coverage and nothing would go red. It is now checked against a set **derived from source** (every exported `async function` in the three adapter modules), with a positive control proving the deriver sees functions, and a failure that **names** the uncovered route. Mutation-verified: renaming a row's key REDs with `expected [ 'restoreModelVersion' ] to deeply equal []`.

**Gate 2 — two sign-in refusals, different casing.** `SIGN_IN_REQUIRED` (guest) vs `sign_in_required` + `validator: "user_jwt"` (JWT). The UI matched UPPER only, so the JWT refusal became *"could not be saved right now. Try again."* on a `retryable: false` response — and a silent hydration failure on the graph read. **This PR creates that reachability** (`refused` needs a token to be *presented*), so it is fixed here. One predicate with two disjuncts: `validator === 'user_jwt'` (robust, casing-immune) plus a case-insensitive code match. Applied in `sharedRefusal`, so LIST gains a branch it never had. Refusal fixture is **verbatim from the wire**.

## What this PR does

Sends the token on all five routes via the existing `buildTurnAuthHeaders`. **Both identity fields come from one `getSessionIdentity()` read at every call site** — six now. Keeps the body `user_id`, pinned by a test, until CEE strips it. Folded five copies of the guest-identity rule into `src/lib/guestIdentity.ts`, **mint included**. Corrects the stale "flag is off" comments.

Guests remain byte-identical: no session → both null → no headers.

## Evidence

- `scenarioRoutesUserJwt.spec.ts` **32** · `signInRefusal.spec.ts` **9** · `useServerGraphHydration.spec.tsx` **9** — counts asserted **by name**. Corpus includes the classes the signed-in/guest pair is blind to: **(real id, null token)** and **(real id, another session's token)**.
- **Mutants, discriminating pairs in both directions.** Dropping `...authHeaders` REDs exactly its own route's tests; always-emit REDs all five guest tests. Worktree outside the repo root, isolation proven by **writing** a sentinel.
- **Negative control on the instrument:** with the Supabase env vars unset the hook spec dies at collect while the run prints `Tests 20 passed` and zero failures.
- Lint **0 errors** · typecheck at exactly baseline (**2252**, zero drift) · **223 files / 2769 tests** green across every touched area.

## ⚠ Not established — do not let these travel as settled

- **The 56-of-64 orphan-owner figure is INHERITED from the coordinator's brief, not corroborated by review.** The reviewer could not authenticate with found credentials (correctly — that constraint is absolute) and did not re-derive it.
- **No verified-mode wire probe exists.** That a valid token is *honoured* on these five routes is derived at the bytes only. The refusal branch is wire-witnessed; the acceptance branch is not.
- **The real-A-owns / real-B-works-on shape is undetectable in the data.** My low-risk reading is an inference from how the UI sources `opts.userId`, not a measurement.

## ⚠ Witness owed — merge is held

**No signed-in journey has been driven against this.** Account minting is outside what this lane may do. Held for the founder.

### DIFF
--- a/src/adapters/cee/__tests__/scenarioRoutesUserJwt.spec.ts
+++ b/src/adapters/cee/__tests__/scenarioRoutesUserJwt.spec.ts
@@ -36,6 +36,8 @@ const USER_ID = '99999999-8888-4777-8666-555555555555'
 
 /** A JWT-shaped token. Distinctive so an assertion cannot match by accident. */
 const ACCESS_TOKEN = 'eyJhbGciOiJFUzI1NiJ9.eyJzdWIiOiJzMy1zcGVjIn0.c2lnbmF0dXJl'
+/** A token from a DIFFERENT session — distinct bytes, so a mix-up is visible. */
+const OTHER_TOKEN = 'eyJhbGciOiJFUzI1NiJ9.eyJzdWIiOiJvdGhlci1zZXNzIn0.b3RoZXJzaWc'
 
 let fetchSpy: ReturnType<typeof vi.fn>
 
@@ -155,6 +157,49 @@ describe('scenario routes — a guest is byte-identical to before this change', 
 
       expect(sentBody().user_id).toBeUndefined()
       expect(sentHeaders()['X-User-Id']).toBeUndefined()
+    })
+  }
+})
+
+/**
+ * ── THE TWO CLASSES THE SIGNED-IN / GUEST PAIR CANNOT SEE ──────────────────
+ *
+ * The pair above varies both fields together: (id + token) or (null + null).
+ * That is structurally blind to every input where the two DISAGREE — and
+ * disagreement is exactly what the call sites were fixed to prevent, so the
+ * corpus has to be able to observe it or the fix is unguarded at rest.
+ *
+ * Neither class is a defect in the adapter. The adapter is a transport and
+ * cannot know which session a value came from. They are pinned so the wire is
+ * documented, and so a future change to either channel has to face them.
+ */
+describe('identity classes the signed-in/guest pair is blind to', () => {
+  for (const route of ROUTES) {
+    it(`${route.name}: a real id with NO token sends the body id and no header`, async () => {
+      fetchSpy.mockResolvedValue(jsonResponse(200, route.ok))
+
+      // The pre-#851 world, and the world any signed-in user is in whenever
+      // the token read fails or the session has gone. CEE resolves this as
+      // `service_legacy`, where the body id is the only identity — which is
+      // precisely the channel the CEE-side strip will close.
+      await route.call({ userId: USER_ID, accessToken: null })
+
+      expect(sentBody().user_id).toBe(USER_ID)
+      expect(sentHeaders().Authorization).toBeUndefined()
+    })
+
+    it(`${route.name}: a real id with ANOTHER session's token sends both verbatim`, async () => {
+      fetchSpy.mockResolvedValue(jsonResponse(200, route.ok))
+
+      await route.call({ userId: USER_ID, accessToken: OTHER_TOKEN })
+
+      // Both go out as given. The adapter does not and cannot arbitrate; the
+      // guarantee that these two never disagree lives at the CALL SITE, which
+      // is why every call site reads both fields from one session object.
+      // CEE is the arbiter: with the flag on, the verified `sub` wins and the
+      // body id is ignored, so a mismatch is telemetry and not a privilege.
+      expect(sentBody().user_id).toBe(USER_ID)
+      expect(sentHeaders().Authorization).toBe(`Bearer ${OTHER_TOKEN}`)
     })
   }
 })
--- a/src/adapters/cee/__tests__/scenarioRoutesUserJwt.spec.ts
+++ b/src/adapters/cee/__tests__/scenarioRoutesUserJwt.spec.ts
@@ -1,0 +1,160 @@
+/**
+ * The scenario routes carry a VERIFIED TOKEN — UI half of the ownership fix.
+ *
+ * ── WHY THIS SPEC EXISTS ────────────────────────────────────────────────────
+ * `CEE_REQUIRE_USER_JWT` is ON in deployed staging (measured at the boot log:
+ * `require_user_jwt: true`; and on the wire: a JWT-shaped invalid Bearer
+ * answers 401 `validator: "user_jwt"`, a branch only reachable with the flag
+ * on). These three adapters nevertheless sent NO `Authorization` header, so
+ * every signed-in caller resolved `service_legacy` at CEE and the body
+ * `user_id` was their ONLY identity — a field an unauthenticated caller can
+ * forge, demonstrated end-to-end against deployed staging.
+ *
+ * CEE cannot strip that field until the token arrives, or it would refuse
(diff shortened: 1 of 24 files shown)


Talchain:
## Green at `6ba5c0ea` — rebased, and one real regression caught by CI

**Required context `Staging Gate`: SUCCESS.** 14 checks, running=0, 13 success, 1 failure — `Visual Regression (advisory)`.

**That advisory does not gate, and here is the control rather than an assertion:** on the untouched staging base `3953d958`, `Visual Regression (advisory)` = **failure** while `Staging Gate` = **success**. Same pairing, no change of mine involved.

### CI caught a real regression, and it was mine

The first head failed `Full Test Suite (shard 2/4)`: `useServerGraphHydration.spec.tsx` pinned that the hook passes `useAuth()`'s user id, and the hook now reads identity from `getSessionIdentity()`. Locally green, red in CI — exactly the split this repo's env-var trap produces.

**The hook's behaviour is correct and stayed.** `useAuth()` remains the effect *dependency*; the value *sent* is read at request time because tokens rotate. Taking both fields from one atomic read is what guarantees the body `user_id` and the `Authorization` header describe the same session — splitting them across two sources would be two answers to one identity question, which is the defect this whole change exists to remove.

So the spec was updated to mock the real source and **strengthened from one assertion to three**: user id through, token through, and the opposite-direction guest twin asserting neither is sent. Mock uses an `importOriginal` spread, not a hand-listed factory.

### Instrument check, with a negative control

`VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` sourced from `DecisionGuideAI/.env.local`.

- **Negative control (vars unset):** `useServerGraphHydration.spec.tsx` dies at collect — while the run still prints `Tests 20 passed` and **zero failures**. The aggregate looks perfectly healthy. That is the trap, reproduced deliberately so the instrument is known to be able to fail.
- **Positive (real vars), counts asserted BY NAME:** `scenarioRoutesUserJwt.spec.ts` = 20/20, `useServerGraphHydration.spec.tsx` = 9/9, failed=0. Not inferred from a suite total.

### Rebase overlap — re-derived at the new tip, not inherited

Staging moved twice and #849 touched `serverGraphHydration.ts`, the same file. Re-measured against the **current** tip: their lowest hunk starts at line **172**; my hunks are at **55** and **109**. Disjoint, 69-line gap. Rebase applied with no conflicts; both changes verified intact afterwards.

### Local gates at this head

Lint **0 errors**, hooks ratchet exactly at baseline · typecheck **PASSED** at exactly baseline (2252, zero drift) · **92 files / 1060 tests** green across every touched area.

### ⚠ Still owed, unchanged

**No signed-in journey has been driven against this.** Minting a disposable authenticated account is outside what this lane may do. Before the CEE strip lands: sign in, confirm a signed-in user reads and registers their own scenario, and confirm CEE resolves them `verified` rather than `service_legacy`.

Talchain:
## Four review fixes in at `872aa75d` — `Staging Gate` SUCCESS

`total_count=14 · running=0 · 13 success · 1 failure` (`Visual Regression (advisory)`, which does not gate — control: on untouched staging base it fails while `Staging Gate` succeeds).

**(1) Atomic read at all four remaining sites.** `useImportRegistration` + all three `ServerVersionsSection` handlers now take the body id from the same `getSessionIdentity()` read as the token. The review's point was exactly right — `useAuth().user.id` is React state defaulting to `'guest'`, so the property this PR claims was true at 1 of 5 sites.

**(2) Deploy-order note corrected — the PR body is rewritten.** This is the cutover, not its preparation: CEE's precedence flip at `route-v2-preflight.ts:188-206` is already deployed and gated only on a token arriving. Body now leads with that, plus the residual-risk statement (orphan rows cannot lock anyone out; the real-A-owns/real-B-works-on shape is undetectable in the data and my low-risk reading is an inference, not a measurement).

**(3) Corpus rows added** for **(real id, null token)** and **(real id, other-session token)** — the classes the signed-in/guest pair is structurally blind to, because it varies both fields together. Spec now 30 tests, count asserted by name.

**(4) Sole source-discriminator named — and I re-measured it rather than taking it on trust.** Precise mutation (token still atomically read, body id re-sourced to `useAuth()`): `passes the signed-in user id through` stays **GREEN**, only the guest twin goes **RED**. Confirmed exactly as reported. Commented so a tidy-up cannot silently un-pin the source.

**Bonus, since all copies were in files already touched:** folded `GUEST_USER_ID` ×4 and the sanitise predicate ×3 into `src/lib/guestIdentity.ts` — and the **mint** in `AuthContext` imports it too, otherwise the shared constant is just a sixth copy that happens to match today.

`ServerVersionsSection.spec.tsx` needed the same treatment as its subject: it mocked `useAuth` but not the real identity source, so it was asserting against a source the component no longer reads.

**`X-User-Id` noted and left alone** — redundant, zero readers in CEE, delete it at the strip rather than churn this PR.

Local: lint 0 errors · typecheck at exactly baseline (2252) · 224 files / 2779 tests green. Witness still owed and unchanged.

PR #886, opened 2026-08-26
D2 — remove the duplicate Inspector authority; keep the policy in its enforced form

Talchain:
## What disappears, what survives

**DISAPPEARS:** `NODE_SETTER_AUTHORITY` (21 keys) + `EDGE_SETTER_AUTHORITY` (5) — 26 keys of prose with **zero code consumers**. Outside their own definition and `mutationAuthority.spec.ts`, every reference to either was a *comment*. Nothing branched on them, so they could neither drift nor enforce.

**SURVIVES:** the policy in its **enforced** form — `InspectorRouter`'s unconditional `<fieldset disabled>` around every panel, node (`:334`) and edge (`:221`), beneath `INSPECTOR_READ_ONLY_REASON`. Strictly stronger than the manifests: it disables descendants structurally, so a setter added tomorrow is inert without anyone remembering to classify it.

Contrast control at time of deletion: `useNodeMutations`/`useEdgeMutations` were imported by 30+ files — the sweep that returned zero for the tables was demonstrably able to see consumers.

## Scope — 5 files, deliberately

| File | Why it is here |
|---|---|
| `useInspectorMutations.ts` | the deletion itself |
| `mutationAuthority.spec.ts` | **forced** — it *imports* all four deleted symbols; without this edit the tree does not compile. The single real code consumer, and the exception the sweep above names in advance. |
| `inspectorAuthorityBinding.spec.tsx` | the replacement guard |
| `mutationAuthority.ts` | docstring names the question it answers |
| `ModelTabV2Panel.tsx` | corrects a false mount claim |

An earlier revision of this PR also redirected 11 files' worth of prose citations of the deleted symbols. Those were prose-only, not forced by the deletion, and have been **reverted** — an authority removal should be reviewable as one thing. One of them is a genuine finding and is rowed separately: `analyticalNodeFields.ts` has a `note:` **runtime string** that ships to browsers naming `NODE_SETTER_AUTHORITY` (measured: one occurrence in the deployed bundle at `assets/store-*.js`, commit `f287c012`; `EDGE_SETTER_AUTHORITY` zero).

## The guard that makes the deletion safe

`inspector-v2/__tests__/inspectorAuthorityBinding.spec.tsx` — 12 tests. It pins what nothing covered:

1. the notice renders the **constant**, not a substring of it;
2. `aria-describedby` **resolves to that notice by identity** — nothing asserted the binding, so it could be deleted silently, leaving the control inert and the user never told why;
3. no editing control inside the boundary is **effectively** enabled (the deleted completeness check, replaced by a DOM claim that holds however many setters exist).

### Discriminating mutant matrix
Worktree outside the repo root; isolation proven by a sentinel whose landing was asserted (worktree=1, source=0, inodes differ); pristine-archive restore; applied-check scoped to `src/` = 1 per mutant, control = 0, trailing control clean.

| Mutant | New guard | Other region | Pre-existing guard |
|---|---|---|---|
| A node `aria-describedby` removed | node BINDS **RED** | edge GREEN | — |
| B edge `aria-describedby` removed | edge BINDS **RED** | node GREEN | — |
| C node `disabled` removed | node inert + escapes **RED** | edge GREEN | — |
| D node copy gutted, **old substring kept** | **RED** | edge GREEN | **GREEN — blind** |
| E edge copy gutted, **old substring kept** | **RED** | node GREEN | **GREEN — blind** |

D and E are the added-value demonstration: the pre-existing substring assertion stays fully green while the copy is gutted.

## A ruling was inverted by measurement — read this before reviewing

The original plan was to wire `CANONICAL_EDIT_AUTHORITY` into `useModelEditAuthority`. **That is withdrawn**, and the docstring now records why.

`CANONICAL_EDIT_AUTHORITY` is a **presentation** authority. 14 of its 21 keys have production consumers and *every one* reads it into a `*_CONNECTED` boolean that decides what to **render**. It answers *"may this control **look like** a shared-model edit?"* — not *"may this write happen?"* Wiring it into the writer would use the first answer to gate the second, and would have turned off `proposeOptionIntervention` and `proposeFactorConfirmation`.

Both are **genuinely working writes**: `interventions` and `observedState` are `purposes: ['stale']`, *"persisted by hash-by-default"*, absent from the three-field ephemeral denylist, and read by `hasAnalyticalNodeChange` (live in `store.ts`, `analyticalChange.ts`, `graphChangeDiff.ts`, `applyPatch.ts`). Disabling them would have removed the only way to say what an option does and the only way to confirm a factor value.

**Scope limit, stated because it bounds the claim:** there is **no wire witness** of `interventions` reaching PLoT. `islRequestAdapter`'s only callers are `useISLConformal.ts` and a debug tab — the conformal path, not the main run. The persistence and staleness findings stand on their own and do not depend on it.

## A false comment corrected

`ModelTabV2Panel.tsx` asserted that `ModelTabBody` *"renders `FactorsSection` unconditionally"*. **It does not** — `LEGACY_DETAILED_EDITOR_MOUNTED = false` (`ModelTabBody.tsx:120`) gates the whole v1 stack at `:917` and esbuild folds it out.

Verified at the **deployed bundle** (staging `f287c012`, 81 chunks crawled): `factor-card-`, `factors-add-cta`, `attribution-stability-pill`, `range-derivation-badge` all **zero**, while positive controls fired in the same sweep (`option-card-` ×2 — both from live *results* components — `model-scientific-transparency`, `model-tab-v2-panel`) and a fabricated control read zero.

The comment described the call site (`:972`/`:981`) without its enclosing guard, which is exactly how it read as true. It now names the guard.

## Gate — the required check's real step sequence, all of it

`TypeScript + Lint` is **seven** steps, not two. All run locally at this head, plus the separate self-test on a clean tree:

| Step | Result |
|---|---|
| `pnpm run lint` (eslint + hooks-ratchet) | exit 0 |
| `pnpm run typecheck` | exit 0 — 2251 diagnostics = baseline exactly |
| `pnpm run ci:guard:zustand` | exit 0 |
| `pnpm run ci:guard:starters` | exit 0 |
| `pnpm run check:vendor` | exit 0 |
| `pnpm run ci:guard:schemas-version` | exit 0 |
| `pnpm run ci:guard:ds:enforce` | exit 0 |
| `pnpm run typecheck:selftest` | exit 0 |

Tests over every touched area: **132 spec files, 1892 passed / 0 failed**, with the new spec's collected count asserted **by name** (12). No stray `.only`/`.skip` in the changed specs, detector's own positive control firing.

## Review notes

Please check the **breadth** of the new guard's claims rather than only that they pass. Instrument defects caught and disclosed during this work: a grep that ate leading-dash strings as flags (false zero on every item at once), a `sed` that stripped the vitest pass/fail marker, a single-byte capture of a multibyte marker, and un-anchored mutation patterns that silently hit the wrong region — only the final matrix is trustworthy, and it is the one above.

---

## Review remedies R1–R6 applied at `2168b51e` (27 Aug 2026)

**R1 — the guard that could not fail.** Reproduced the reviewer's mutant M5 (an enabled `<input aria-label="ESCAPED_EDITOR">` in the `InspectorShell` header, inside the Inspector and outside the fieldset, applied-check 1 file / 1 insertion) **before** replacing anything: the shipped suite stayed **12/12 GREEN**. Both cases built their candidate set from `fieldset.querySelectorAll(...)`, so `isEffectivelyDisabled` was true **by construction**. Replaced with a query over the whole Inspector region minus the boundary minus an identity-pinned deliberately-outside set, widened to `[role="button"], [contenteditable], a[href]`. The exemption list is fenced **both ways** — every entry must match something, and every match must resolve outside the boundary — so it cannot quietly become the answer. New guard: **17/17 GREEN clean, 3 RED under M5**, naming `<input> ESCAPED_EDITOR`. Node/edge discriminating pair re-run and still clean.

**R2 — "strictly stronger" withdrawn.** `<fieldset disabled>` inerts form-associated descendants only. Reworded to "every descendant **form control**" and dropped "a setter added tomorrow is inert", in both places that carried it. The reviewer's bound is carried exactly and **not** upgraded: **no write escapes**, and user reachability was explicitly not claimed (its store write used a synthetic `fireEvent.change`). The non-inerted set is now pinned in the suite as `NOT_INERTED_BY_THE_FIELDSET_NODE`, so it REDs if it grows **or** shrinks — the bound lives in the tests, not in a comment.

**R3 — "every reference … was a COMMENT" was false** and is corrected in all three places that carried it (not only the one flagged). `analyticalNodeFields.ts:158` names the table in a runtime string literal; verified it is a `note:` property value and that all four named importers exist. The narrow claim — zero code **consumers** — is true and is what the deletion rests on.

**R4 —** the docstring disavowed "a bare `document.querySelector('fieldset')`" while `readBoundary()` is exactly that. Corrected to the real mechanism (one region mounted per `describe`) and named the refactor that would destroy it.

**R5 — framing only; nothing deleted.** Both mutants re-run here: inverting the constant's meaning while keeping the guarded substring leaves **68/68 GREEN**; gutting the constant REDs **exactly two, both pre-existing** (`InspectorRouter.spec.tsx:347`, `FactorExternalPanel.priorRangeHonesty.spec.tsx:705`). This file is the weaker of the two on meaning and now says so, and says keep both.

**R6 — 12 sites, not 11.** Every line number re-derived at this tip. The extra site is `FactorExternalPanel.priorRangeHonesty.spec.tsx:76`, which also carried a third dead citation (`useInspectorMutations.ts:127`). `analyticalNodeFields.ts` deliberately untouched — owned follow-up.

### Shape of the delta, proven rather than asserted
13 files, +415 / −106. Transpiling base and head with comments stripped, **12 of 13 files emit byte-identical code**: the only executable change in the whole delta is the spec this remedy was about. No shipped behaviour or copy changed.

### ⚠ Instrument disclosure — the part worth keeping
Two independent attempts at that comment-only proof returned a **uniform IDENTICAL across all seven files** while measuring nothing: both sides hashed the empty string (`da39a3ee5e6b4b0d3255bfef95601890afd80709`, byte counts 0/0) because esbuild's CLI was never installed, and a second attempt did the same after `node` left `PATH`. Each was caught only by a contrast control required to DIFFER. **Uniformity across items that ought to differ is evidence about the instrument, not the world — and an absence proof that flatters its author is the one nobody re-checks.**

### Gate at this tip
`typecheck` PASSED (556 files / 2251 errors, baseline 2251, Δ0) · `typecheck:selftest` PASSED (18/0) · `lint` PASSED (0 errors; hooks ratchet 229/13 exactly at baseline) · targeted vitest 9 files / **120 passed**, target spec collected **by name** (17). Full suite left to CI by instruction.

### DIFF
--- a/src/canvas/model-tab-v2/ModelTabV2Panel.tsx
+++ b/src/canvas/model-tab-v2/ModelTabV2Panel.tsx
@@ -145,16 +145,34 @@ export function ModelTabV2Panel({
    * the branch above is structurally exclusive and the spec pins the outline rows
    * absent in queue mode, with a before-click contrast control.
    *
-   * What is NOT yet true is the SURFACE. `ModelTabBody.tsx` renders
-   * `FactorsSection` unconditionally, outside this panel, in the same scroll — so
-   * a queued factor appears in the queue AND in its v1 factor card. The
-   * consolidation's invariant *"there is only ever one rendering of a row"* is a
-   * goal at this tip, not an achieved state, and it stays that way until the
-   * duplicate section stack is deleted.
+   * ⚠⚠ THE PARAGRAPH THAT USED TO SIT HERE IS FALSE AND IS CORRECTED BELOW
+   * (26 Aug 2026). It read: *"`ModelTabBody.tsx` renders `FactorsSection`
+   * unconditionally, outside this panel, in the same scroll — so a queued
+   * factor appears in the queue AND in its v1 factor card."* **It does not.**
    *
-   * A spec that renders THIS COMPONENT can never see that (trap 3b — bound to a
-   * component, not to the surface the deployed tab mounts), which is exactly how
-   * the overstatement passed a green suite.
+   * `ModelTabBody.tsx:120` declares `const LEGACY_DETAILED_EDITOR_MOUNTED =
+   * false`, and `:917` gates the ENTIRE v1 section stack behind it
+   * (`{LEGACY_DETAILED_EDITOR_MOUNTED && (<section data-testid="model-tab-v1-stack">…)}`).
+   * esbuild folds the constant, so Goal/Options/Factors/Relationships/Risks are
+   * dead-code-eliminated — they are not merely hidden, they are not shipped.
+   * Verified at the DEPLOYED bundle (staging `f287c012`, 81 chunks crawled):
+   * `factor-card-`, `factors-add-cta`, `attribution-stability-pill` and
+   * `range-derivation-badge` all read ZERO, while positive controls in the same
+   * sweep fired (`option-card-` ×2 — both from live RESULTS components —
+   * `model-scientific-transparency`, `model-tab-v2-panel`). So this panel IS
+   * the sole rendering of a row, and the consolidation invariant HOLDS.
+   *
+   * ⚠ HOW THE FALSE VERSION SURVIVED, because that is the reusable part: the
+   * JSX for those sections really does sit at `ModelTabBody.tsx:972`/`:981`
+   * with no local condition. Reading the call site WITHOUT its enclosing guard
+   * at `:917` produces exactly the sentence above — and a later reader who
+   * greps for `<FactorsSection` reproduces the same mistake. A comment that
+   * describes a mount MUST name the guard, not just the call site.
+   *
+   * A spec that renders THIS COMPONENT can never settle either version (trap 3b
+   * — bound to a component, not to the surface the deployed tab mounts), which
+   * is why the deployed-bundle sweep above, not a green suite, is the evidence
+   * cited here. Re-derive it rather than inheriting this paragraph.
    */
   const [activeQueue, setActiveQueue] = useState<MountedQueueId | null>(null)
   /**
--- a/src/canvas/mutations/__tests__/mutationAuthority.spec.ts
+++ b/src/canvas/mutations/__tests__/mutationAuthority.spec.ts
@@ -146,9 +146,15 @@ describe('mutation authority is exhaustive and fail-closed', () => {
    * They asserted over `NODE_SETTER_AUTHORITY` / `EDGE_SETTER_AUTHORITY` —
    * that every Inspector setter was classified, and that none was classified
    * `'server_graph'`. Those two tables were the only readers of themselves:
-   * outside this file and their own definition, every reference was a COMMENT.
-   * So the tests proved a hand-maintained list agreed with itself, which is
-   * not the same as proving the Inspector is read-only.
+   * outside this file and their own definition, every reference was a comment
+   * or a documentation string, and none was a consumer. So the tests proved a
+   * hand-maintained list agreed with itself, which is not the same as proving
+   * the Inspector is read-only.
+   *
+   * ⚠ This read "every reference was a COMMENT" until 27 Aug 2026. That was
+   * false — `canvas/domain/analyticalNodeFields.ts:158` names the table in a
+   * runtime string literal that ships to browsers. The claim the deletion
+   * rests on is the narrow one: zero code CONSUMERS.
    *
    * What actually keeps it read-only is `InspectorRouter`'s unconditional
    * `<fieldset disabled>` around every panel. That is now pinned, per region
--- a/src/canvas/mutations/__tests__/mutationAuthority.spec.ts
+++ b/src/canvas/mutations/__tests__/mutationAuthority.spec.ts
@@ -5,12 +5,6 @@ import {
   hasServerGraphAuthority,
   type MutationAuthority,
 } from '../mutationAuthority'
-import {
-  EDGE_SETTER_AUTHORITY,
-  EDGE_SETTER_FIELDS,
-  NODE_SETTER_AUTHORITY,
-  NODE_SETTER_FIELDS,
-} from '../../ui/inspector-v2/useInspectorMutations'
 
 /**
  * Frozen, independently-written audit contract from Brief B3. This is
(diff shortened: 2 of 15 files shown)


Talchain:
## Independent review at `3b286bc0` — **FIX-FIRST**, six mechanical remedies. **The deletion itself survived refutation.**

Behaviour is sound; every finding is a claim in **shipped prose** or a guard weaker than its own name. Nothing here asks for the deletion to be re-litigated.

### Established, and more strongly than I had it

- **The absence claim holds**, multi-line-import-aware, at base **and** at the current tip: importers = **1** for both symbols, against a contrast of `useCanvasStore` at **711 importers / 6151 hits / 933 files** in the same invocation. Barrel exports only three symbols; **zero `export *` laundering paths**. Complete classified manifest: NODE 15 hits = 1 def + 6 test-code + 7 comments + **1 runtime string**; EDGE 13 = 1 def + 4 test-code + 8 comments.
- **Discriminating pair proven**: deleting node-`disabled` REDs node cases and leaves **edge all GREEN**; deleting edge-`disabled` REDs edge and leaves **node all GREEN**. `applied=1` each, leading and trailing controls `applied=0`.
- **Scope contained by measurement**: comment-stripped diff gives **0 non-comment lines** for `ModelTabV2Panel.tsx` (324→324) and `mutationAuthority.ts` (33→33).
- All seven gate steps plus `typecheck:selftest` pass on a clean tree (baseline 2251, Δ0). Zero intersection with #887/#888/#890, contrast firing.
- ⭐ **My withdrawal of the RED is confirmed with a stronger contrast than mine:** the two "failures" belong to the **cancelled** run; required contexts are `["Staging Gate"]` only, SUCCESS on the live run — and **staging tip `16c55158` itself fails Visual Regression**, so it is a standing advisory red, not this PR's.

---

## ⛔ The gate that matters — a completeness guard replaced by one that cannot observe an escape

Mutant **M5** added an enabled `<input aria-label="ESCAPED_EDITOR">` to the `InspectorShell` header — **inside the Inspector, outside the fieldset**. `applied=1`. **12/12 GREEN.**

Both tests query `fieldset.querySelectorAll(...)`, so `isEffectivelyDisabled` is true **by construction**; `escaped` is non-empty only when the fieldset loses `disabled`, which the sibling test already measures. **It duplicates that test under a name it does not earn** — and its docstring calls it *"THE REPLACEMENT FOR THE DELETED COMPLETENESS GUARD"*.

This PR deletes a completeness guard and installs a guard that cannot fail in the way its name claims. **Remedy (tested):** query the whole rendered container, subtract the deliberately-outside set, include `[role="button"], [contenteditable], a[href]`. That version REDs on M5 **and** on the finding below.

## ⛔ "Strictly stronger" is refuted by execution

`<fieldset disabled>` inerts only **form-associated** descendants. Measured inside the Decision panel's boundary: one `[role="button"]` div (`EmptyDescriptionPrompt`, `tabindex=0`) **takes focus, fires its handler, and opens an editor** (textareas 0→1) — while, contrast in the same run, the two real `<button>`s there **are** disabled. That shape is used by 6 panels; `ConnectionRow` by 7.

⭐ **The scope is precise and I am not overstating it: no write escapes.** The opened `<textarea>` is inside the fieldset and natively disabled, and the reviewer explicitly **declined to claim user reachability** because its store write came from a synthetic `fireEvent.change` that bypasses the browser's own gating. Reword to *"every descendant **form control**"*, and drop *"a setter added tomorrow is inert"*.

## The other four, all one-line

3. **`useInspectorMutations.ts:102-105` ships a false sentence** — *"every reference … was a COMMENT"*, but `analyticalNodeFields.ts:158` is a **runtime string literal** in a file imported by `useAutosave`, `graphChangeDiff`, `analyticalChange` and `useGraphEditEvents`, so it ships to browsers. The narrow *"zero code consumers"* is true; say *"a comment or a documentation string, and none was a consumer."*
4. **The stated binding mechanism is not the code** — the docstring disavows *"a bare `document.querySelector('fieldset')`"*, which is exactly what `readBoundary()` is. The discrimination is real but comes from each `describe` rendering one region.
5. **The copy guard is weaker than the pre-existing one it critiques** — inverting the fragment's meaning to *"You may freely edit anything here"* leaves **all 53 GREEN**, while gutting the constant REDs the two **pre-existing** guards. No coverage is lost; the framing invites the opposite reading.
6. **11 present-tense references to the deleted symbols survive in `src/`**, several citing `(:119)`/`(:143)` that will point at nothing.

---

## ⭐⭐ Sequencing — and this changes the follow-up from "queued" to "owed"

`analyticalNodeFields.ts:158`'s browser-shipped `note:` **currently names a live table and is TRUE. After this merges it names a symbol that does not exist.** So this PR does not merely leave stale prose behind — it **converts a true runtime string into a false one**, in a file that ships.

The ordering is right: you cannot write *"the table was deleted"* before it is. **So I am naming the window rather than leaving it open** — I hold the follow-up specified and will land it in this same session, immediately after this merges, covering `:158`'s runtime string, the `//` block above it (keeping the correction and its reason, removing only the dead symbol citation), and the 11 reversions. Line numbers to be re-derived at the tip, not inherited from this comment.

Take remedies 1–6 and I will merge on the next green.

⭐ One note worth keeping, and credit to the reviewer for disclosing it: its first three-spec comparison passed an unquoted list under zsh, vitest found no files and **exited 0** — caught by its own *"no summary ⇒ PROBE FAILED, not a measured zero"* rule. That guard is the only reason the run was not laundered into a verdict.

Talchain:
## Merging. Delta verified at the bytes, in both directions.

13 files since the reviewed head. Every non-spec source file reads **0 non-comment added lines** — except two that my first filter flagged at 7 each, which I read rather than waved through: both are **JSX comment blocks** (`{/* … */}`), which a `//`-and-`*` filter cannot see. Pure prose. The control fired at **98** on `inspectorAuthorityBinding.spec.tsx`, so the zeros are measurements.

⭐ **And the R6 corrections are the right shape**: they keep the claim, delete only the dead citation, and record *why* — including the load-bearing point that **the fieldset was always the whole enforcement**, so nothing about these controls' inertness changed with the deletion. A correction in place with its reason visible, not evidence tidied away.

**Visual Regression discharged by NAME, not by colour:** 11 failing at this head, 13 at the reviewed base, **in-mine-but-not-in-base = empty** — a strict subset, two fixed. ⭐ Worth noting one of the 11 at *both* tips is the harness's own positive control, *"visual harness self-test — proves the instrument can fail"* — so the suite demonstrably can fail.

`Staging Gate` **completed/success**, `not_completed=0`, single run with no cancelled twin.

---

## ⭐⭐ The disclosure in the body is the most valuable thing in this PR

Four blind probes in one lane, and **every one agreed with what the author wanted to be true**: two independent comment-only proofs returning a uniform IDENTICAL across all seven files while hashing the **empty string** — once because esbuild's CLI was never installed, once because `node` left `PATH`; and a name-diff that "agreed" because both sides were empty. Each was caught only by asserting the inputs were non-empty **before** believing the agreement.

*Uniformity across items that ought to differ is evidence about the instrument, not the world — and an absence proof that flatters its author is the one nobody re-checks.*

**The follow-up is mine and lands immediately**: `analyticalNodeFields.ts:158` carries a **runtime string that ships to browsers** and is true until this merges, false after. Line numbers re-derived at the tip, not inherited.

PR #890, opened 2026-08-27
fix(versions): teach the modelVersions adapter CEE's v2 list and restore

Talchain:
## The defect (live, user-facing, on the deployed build)

**Version history listed nothing and restore could not complete, while save still succeeded** — so a user could save versions they could then neither see nor restore.

CEE bumped LIST and RESTORE to v2 in `4c29c5b5` (26 Aug 14:52Z), an ancestor of the deployed staging tip `3c3d3d53` (`/healthz` → `build: 3c3d3d5`):

| call | CEE (deployed) | UI (staging `16c55158`) | result |
|---|---|---|---|
| list | `model_versions_list.v2` | demanded `…v1` | `unusable` |
| restore | `model_version_restore.v2` | demanded `…v1` | `unusable` |
| **save** | `model_version_save.v1` | `model_version_save.v1` | **works — untouched** |

The adapter's checks are hard equality, so both reads fell straight through to `unusable`.

This is a **REST contract bump landed without its reader** — not schema-pin skew. No pin alignment would have prevented it.

It fails closed to an honest retry state, so this is a **broken capability, not a trust defect**. The fail-closed behaviour and its copy are deliberate and are **unchanged**.

## The fix

The bump moved the **shape**, not only the label. Every field below was derived at the bytes from CEE `3c3d3d53` (`assist.v1.scenario-versions.ts`, `history-v2.ts`, `mutation-receipt.ts`) and cross-checked against **CEE's own route test**, which pins the exact top-level key set.

**restore** — `graph`, `version`, `undo_version_id` are gone from the top level and now live inside a nested `receipt`; the ordinal is `sequence`, not `version_number`.

**list** — rows are `version_id` / `sequence` / `full_hash` with structured `actor` / `creation` / `lineage`, replacing `id` / `version_number` / `graph_identity_hash` / `provenance`.

`parseReceiptWriteOutcome` is deliberately **named apart** from `parseWriteOutcome`: they answer different questions (v2 receipt `sequence` vs v1 save-envelope `version_number`). Collapsing them would be the similar-names/different-questions defect this codebase has paid for before.

## Decision: v2-only, not a compatibility path

A whole-repo sweep of CEE finds **zero** v1 list/restore literals (contrast controls in the same sweep: 7 v2, 2 `save.v1`), and the route does **no content negotiation** — one hardcoded discriminator per response. There is no live v1 producer, so a v1 branch would be a compatibility path with nothing to serve it and no deletion condition. Two tests pin the refusal, so this **REDs if a v1 producer ever appears**.

## ⚠ Contract narrowing — reported, not papered over

**v2 carries no replay signal on the restore wire.** CEE computes `replayed` and only *logs* it; the response schema is `.strict()` and CEE's own test pins the exact key set. So `deduped` is `false` by construction on the v2 path.

Consequence: on an idempotent replay the panel no longer says *"already at that version — nothing changed"*. It falls to the `changedNothing` branch — *"Restored on the server. If the canvas looks unchanged, reload the page…"* — plus a warn. **Vaguer, still true, fails safe.** No signal is invented to fill the gap.

## Evidence

**RED-first, by name** — the break reproduced before any fix:
- `listModelVersions — v2 > parses a real model_versions_list.v2 envelope…` → `expected 'unusable' to be 'list'`
- `restoreModelVersion — v2 > reads graph, version and undo id from the nested receipt` → `expected 'unusable' to be 'restored'`

**Fixture preconditions pinned in-test** — the fixtures are asserted to carry the v2 label *and* the nested `receipt`, *and* to lack the v1 field names, so they cannot pass for the wrong reason.

**Discriminating mutant pair** (throwaway worktree outside the repo root; isolation proved by *writing* a sentinel and confirming the source was unchanged; HEAD-relative restores; applied-check scoped to `src/`):

| mutant | expect | got |
|---|---|---|
| list guard v2→v1 | RED | RED (4 failed) |
| restore guard v2→v1 | RED | RED (2 failed) |
| graph read receipt→top-level | RED | RED (2 failed) |
| `receipt.sequence`→`version_number` | RED | RED (2 failed) |
| list `version_id`→`id` | RED | RED (3 failed) |
| list `full_hash`→`graph_identity_hash` | RED | RED (3 failed) |
| **save guard mutated — v2 tests only** | **GREEN** | **GREEN** |
| same mutant, full suite | RED | RED (1 failed — the save test) |

The last two are the pair: the same *kind* of mutation on a **different object** leaves the v2 seam green, and the full-suite run proves that mutation was genuinely applied and detectable. Failure counts vary across mutants — real discrimination, not a uniformly-firing blind probe.

**Collected count asserted by name**: `modelVersions.spec.ts` 13 → **23**, and 23 in every mutant run. Never inferred from a suite total.

**All seven `TypeScript + Lint` steps** run at this tip: lint, typecheck, `ci:guard:zustand`, `ci:guard:starters`, `check:vendor`, `ci:guard:schemas-version`, `ci:guard:ds:enforce` — **all rc=0**.

**Full suite, sharded 4-way as CI does**: shards 1–2 clean; shards 3–4 each carry **one pre-existing failure**, both **discharged by control** — a pristine worktree at `origin/staging` `16c55158` with these changes absent reproduces both identically (`aiPanelTranche1.spec.tsx` focus-ring, `useSensitivityRanking.plotBearer.spec.tsx` proxy path). Neither file is touched here.

## Three tests that would have passed for the wrong reason

Once the adapter went v2-only, three pre-existing tests would have kept passing via the **schema guard short-circuiting before the property under test ever ran**. All are converted, each keeping its **original text in an adjacent comment**. One is re-aimed at a stronger property: a v2 payload carrying a good graph at the old v1 top-level position but no `receipt` must still be refused — the adapter must never fall back to the old location.

## Scope

Disjoint from the in-flight PRs — **measured**, intersection = 0 (#886 `canvas/mutations` + `inspector-v2`; #887 `pre-analysis-v3/**`).

Do not merge or deploy on my account — handing back for review.

### DIFF
--- a/src/adapters/cee/__tests__/modelVersions.spec.ts
+++ b/src/adapters/cee/__tests__/modelVersions.spec.ts
@@ -2,9 +2,18 @@
  * modelVersions adapter — the UI's client for CEE's versions wiring slice.
  *
  * Server contract (olumi-assistants-service, assist.v1.scenario-versions.ts):
- *   POST /bff/cee/scenarios/{id}/versions          → model_versions_list.v1
+ *   POST /bff/cee/scenarios/{id}/versions          → model_versions_list.v2
  *   POST /bff/cee/scenarios/{id}/versions/save     → model_version_save.v1
- *   POST /bff/cee/scenarios/{id}/versions/restore  → model_version_restore.v1
+ *   POST /bff/cee/scenarios/{id}/versions/restore  → model_version_restore.v2
+ *
+ * ⚠ WHY THE v1 EXPECTATIONS IN THIS FILE CHANGED (2026-08-27). CEE bumped LIST
+ * and RESTORE to v2 in commit `4c29c5b5` (merged 26 Aug 14:52Z, an ancestor of
+ * the deployed staging tip `3c3d3d53` / `/healthz` build `3c3d3d5`). This spec
+ * previously pinned v1 acceptance as CORRECT; that expectation is now false
+ * about the deployed server, so it is changed here deliberately rather than
+ * left to fail. Every converted case keeps its ORIGINAL text in an adjacent
+ * comment so the change is auditable and nothing is silently rewritten.
+ * SAVE remains v1 on both sides and is untouched.
  *
  * What is pinned here:
  *  - the SAME transport rules as scenarioGraph.ts: literal same-origin
@@ -30,22 +39,111 @@ const USER = '0f8a1b2c-3d4e-4f50-9a6b-7c8d9e0f1a2b'
 const VERSION_A = 'aaaaaaaa-1111-4111-8111-aaaaaaaaaaaa'
 const HASH_A = 'a'.repeat(64)
 
-function wireSummary(overrides: Record<string, unknown> = {}) {
+// ─────────────────────────────────────────────────────────────────────────────
+// v2 CONTRACT — list + restore
+//
+// PROVENANCE OF EVERY FIXTURE FIELD BELOW: derived at the bytes from CEE
+// staging `3c3d3d53b5202526e09f2c770c5bcc467d83856d` (== deployed /healthz
+// build `3c3d3d5`), namely
+//   src/routes/assist.v1.scenario-versions.ts
+//     :115 MODEL_VERSIONS_LIST_SCHEMA   = "model_versions_list.v2"
+//     :118 MODEL_VERSION_RESTORE_SCHEMA = "model_version_restore.v2"
+//     :152-161 AtomicRestoreRouteResponseSchema (.strict())
+//     :169-241 summaryV2() — the row→wire mapping
+//   src/orchestrator-v5/model-management/history-v2.ts   (list + summary shapes)
+//   src/orchestrator-v5/model-management/mutation-receipt.ts  (receipt shape)
+// and cross-checked against CEE's OWN route test
+//   src/routes/__tests__/assist.v1.scenario-versions.test.ts
+//     :370 list schema · :795-822 restore, incl. the EXACT top-level key set
+//          ["analysis_state","receipt","request_id","restored","scenario_id","schema"].
+// These are the producer's bytes, not this lane's model of the producer.
+// ─────────────────────────────────────────────────────────────────────────────
+
+const ANALYSIS_HASH = 'd'.repeat(64)
+const SNAPSHOT_VERSION = 'cccccccc-3333-4333-8333-cccccccccccc'
+const RESTORED_VERSION = 'dddddddd-4444-4444-8444-dddddddddddd'
+const MUTATION_ID = 'eeeeeeee-5555-4555-8555-eeeeeeeeeeee'
+
+/** A v2 summary exactly as CEE's `summaryV2()` emits it for its default row
+ *  (actor_kind null + authored_by null ⇒ {kind:'unknown'}; creation_kind null
+ *  with no source pointer ⇒ {kind:'unknown', …}; root_version_id null ⇒
+ *  lineage {kind:'unknown'}). */
+function wireSummaryV2(overrides: Record<string, unknown> = {}) {
   return {
-    id: VERSION_A,
+    version_id: VERSION_A,
     scenario_id: SCENARIO,
-    owner_user_id: USER,
-    version_number: 4,
-    graph_identity_hash: HASH_A,
-    hash_algorithm: 'sha256',
-    identity_projection_version: 'identity.v1',
-    identity_normaliser_version: 'normaliser.v1',
-    graph_schema_version: 'graph.v1',
+    sequence: 4,
     label: 'Before pivot',
-    provenance: 'user_save',
-    restored_from_version_id: null,
     created_at: '2026-08-17T10:00:00.000Z',
+    actor: { kind: 'unknown' },
+    creation: { kind: 'unknown', mutation_id: null, source_turn_id: null },
+    lineage: { kind: 'unknown' },
+    full_hash: HASH_A,
+    analysis_affecting_hash: ANALYSIS_HASH,
     ...overrides,
+  }
+}
+
+function listBodyV2(overrides: Record<string, unknown> = {}) {
+  return {
+    schema: 'model_versions_list.v2',
+    request_id: 'req-1',
+    scenario_id: SCENARIO,
+    current_version_id: VERSION_A,
+    versions: [wireSummaryV2()],
+    next_cursor: null,
+    ...overrides,
+  }
+}
+
+const RESTORED_GRAPH_V2 = {
+  nodes: [
+    { id: 'n1', label: 'Take the job', kind: 'option' },
+    { id: 'n2', label: 'Commute time', kind: 'factor' },
+  ],
+  edges: [
+    {
+      from: 'n1',
+      to: 'n2',
+      strength: { mean: 0.4, std: 0.1 },
+      exists_probability: 0.9,
+      effect_direction: 'positive',
+    },
+  ],
+}
+
+function restoreBodyV2(receiptOverrides: Record<string, unknown> = {}) {
+  return {
+    schema: 'model_version_restore.v2',
+    scenario_id: SCENARIO,
+    restored: true,
+    receipt: {
+      schema: 'model_version_mutation_receipt.v1',
+      scenario_id: SCENARIO,
+      mutation_id: MUTATION_ID,
+      version_id: RESTORED_VERSION,
+      sequence: 4,
+      graph: RESTORED_GRAPH_V2,
+      full_hash: HASH_A,
+      hash_algorithm: 'sha256',
+      identity_projection_version: 'identity.v1',
+      identity_normaliser_version: 'normaliser.v1',
+      graph_schema_version: 'graph.v1',
+      analysis_affecting_hash: ANALYSIS_HASH,
+      actor: { kind: 'known', authored_by: 'owner' },
+      creation: { kind: 'restore', source_version_id: VERSION_A },
+      source_turn_id: null,
+      lineage: {
+        kind: 'known',
+        parent_version_id: SNAPSHOT_VERSION,
+        root_version_id: SNAPSHOT_VERSION,
+      },
+      undo_version_id: SNAPSHOT_VERSION,
+      event_id: `model_version_restored_mutation_${MUTATION_ID}`,
+      ...receiptOverrides,
+    },
+    analysis_state: null,
+    request_id: 'req-5',
   }
 }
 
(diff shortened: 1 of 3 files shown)


Talchain:
# Independent adversarial review — #890

**VERDICT: APPROVE.** I tried to refute the fix at the producer's bytes and with my own mutation kit and could not. Every load-bearing claim reproduced, several at the exact numbers claimed. Five findings below; **none blocks the merge** — the largest is a reporting overclaim in the PR body, not a code defect.

Instrument: fresh blobless clone at a unique `/private/tmp` path, `HEAD` asserted `464a5c00`. Mutation worktree **outside** the repo root, isolation proved by **writing** a sentinel and confirming the source was unchanged (inodes 745062246 vs 744980282, so no APFS hard-link). Leading and trailing controls `applied=0`; trailing pristine 23/23. I did not run the full suite locally.

---

## Premises re-derived (two corrections, neither changes the outcome)

**CEE has moved since the author derived.** Deployed is now `/healthz build: 6712de1` == CEE `staging` `6712de1b`, not `3c3d3d53`. I checked rather than inherited: `3c3d3d53` **is an ancestor**, and `git diff 3c3d3d53..HEAD` over `assist.v1.scenario-versions.ts`, `history-v2.ts` and `mutation-receipt.ts` is **empty**. The derivation basis is still current — but the PR body's SHA is now stale and should say so.

**The orchestrator's "`.v1`: 7" is the spec's count, not the adapter's.** The old *adapter* carries **6** v1 literals (2 restore + 2 save + 2 list) and **0** v2; the old *spec* carries 7. Direction unaffected.

`CanvasMVP.tsx` is at `src/routes/` — confirmed, and I re-derived the whole chain rather than taking either version: `src/routes/CanvasMVP.tsx:295` (Suspense, no flag) → `VersionsPanelHost` → `WhatChangedPanel:325` → `ServerVersionsSection`. **Unflagged at every hop.**

---

## Claims, attacked

**1. The v2 shape — CONFIRMED at the deployed producer, field by field.** Mapping checks out 1:1 against `summaryV2()` and the two schema modules. The one I most expected to be wrong is right: `full_hash ← row.graph_identity_hash`, so the UI's `graphIdentityHash` — the CAS expectation — is **the same underlying value it was under v1**, not a fork. `analysis_affecting_hash` is the other hash and is correctly left alone. Restore envelope `AtomicRestoreRouteResponseSchema` is `.strict()` with exactly `schema · scenario_id · restored · receipt · analysis_state · request_id`.

**2. Zero v1 producers — CONFIRMED by my own contrast sweep** at `6712de1b`, whole repo, one invocation: `model_versions_list.v1` **0**, `model_version_restore.v1` **0**; contrasts in the same run `model_versions_list.v2` **5**, `model_version_restore.v2` **2**, `model_version_save.v1` **2**, and a broad `model_version` control at **209**. Route literals confirmed at `:115` / `:117` / `:118`. No content negotiation. The v2-only decision is sound and the two refusal tests are the right way to hold it.

**3. Save untouched — CONFIRMED byte-identical.** I extracted `saveModelVersion` and `parseWriteOutcome` from both revisions and compared: identical. No leak.

**4. Mutants — REPRODUCED, at the exact claimed failure counts.**

| mutant | claimed | I measured |
|---|---|---|
| list guard v2→v1 | RED (4) | RED (4) |
| restore guard v2→v1 | RED (2) | RED (2) |
| graph read receipt→top-level | RED (2) | RED (2) |
| `receipt.sequence`→`version_number` | RED (2) | RED (2) |
| list `version_id`→`id` | RED (3) | RED (3) |
| list `full_hash`→`graph_identity_hash` | RED (3) | RED (3) |

**Discriminating pair reproduced:** save-guard mutant → v2 describes only **11 passed / 12 skipped, GREEN**; same mutant, full spec → **1 failed / 22 passed**. I added five mutants of my own on properties the author's kit did not cover — `undo_version_id` read position RED(2), `creation.source_version_id` RED(1), `created_at` RED(3), `current_version_id` RED(2), `label` RED(3). Counts vary across all twelve; this is real discrimination.

**5. Collected 13 → 23 — CONFIRMED by name**, with `.env.local` present. `origin/staging`: 13 passed. PR head: 23 passed.

---

## ⭐ The three "wrong reason" tests — settled by execution, and it is **two**

I did not read this one; I ran it. Original spec at `origin/staging` + **new** adapter → **2 failed / 11 passed**.

- **Genuine REDs** (property broke loudly): `POSTs the user id … parses the list envelope`, `sends the CHOSEN version id and expected hash …`.
- **Genuinely vacuous** (passed with the property unexercised): `fails CLOSED on a malformed version row`, `a restore answer with restored:true but NO graph object is unusable`. **Exactly the two the file's own comments name (L250, L445).**
- The other nine passers are **legitimately** unaffected — HTTP-status mapping, transport, and request-shape assertions that never reach the schema guard.

So the third is `never sends the guest sentinel as a user id`, and it is **not** vacuous: it asserts the request body, which the guard never touched. The file is honest about this — its comment says *"this case asserts the REQUEST, not the response"*. **It is the PR body that rolls all five conversions up into "three pre-existing tests would have kept passing via the schema guard short-circuiting before the property under test ever ran."** That is the trap-20 shape: the in-file record is correctly scoped and the summary generalises it, in the direction that flatters the work.

**The replacements are genuinely stronger, and nothing was traded away.** The malformed-row property comes back re-pointed at v2 *and* gains a twin (wrong-type `sequence` + missing `sequence`). The re-aim is the good kind: the new case carries a **perfectly good graph at the old v1 top-level position**, so a naive fallback passes the old test and fails this one — and the original property is separately preserved by a new `receipt carries NO graph` case. Old text is preserved verbatim with its reason at all five conversion sites (L178, 209, 248, 365, 443).

**I looked for a fifth and did not find one** — the execution above enumerates every passer and each one's property is intact.

---

## ⭐⭐ `deduped` — accept the narrowing; the restraint was right; the copy is not merely "vaguer"

Confirmed at the bytes: `replayed` is computed at `assist.v1.scenario-versions.ts:901` and reaches **only** `log.info` at `:928`; the response schema is `.strict()` with no slot for it. **Not inventing a client-side substitute is the correct call and I endorse it** — a fabricated replay signal would be the exact defect class this estate keeps paying for.

But "vaguer, still true" understates what changed. `changedNothing` previously answered one question and now answers two, because `deduped` used to separate them:

- **(a) genuine idempotent replay** — the canvas is already correct, nothing to do;
- **(b) reconcile refused or mismatched** — the canvas may be wrong, a reload may help.

The copy was written for (b) and now also serves (a): *"If the canvas looks unchanged, reload the page to see the restored model."* For population (a) the canvas looks unchanged **because it is right**, and the prescribed remedy is a **no-op**. That is not vagueness — it is a true sentence attached to a futile action, which is the near neighbour of the Research-CTA pattern. Two knock-ons: `logger.warn('server_versions.restore_applied_no_canvas_change')` now fires on a **healthy, expected** outcome, which is a broken alarm; and `ServerVersionsSection.tsx:279` `if (result.deduped)` is now **structurally dead**.

**None of this blocks #890** — the remedy lives in `ServerVersionsSection.tsx`, which this PR does not touch, and demanding it here is "while we're here". Row it.

---

## Findings and smallest remedies

1. **PR body: "three" is two.** *Measured:* original spec + new adapter = 2 failed / 11 passed; only two tests pass with their property unexercised. **Remedy:** edit the PR body to say two, and name them. Zero code risk. This is the one I would not leave — it is a claim about the lane's own verification, and it is what a later session inherits.
2. **PR body cites a stale CEE SHA.** *Measured:* deployed is `6712de1`; `3c3d3d53` is an ancestor with the route bytes unchanged. **Remedy:** one sentence — "derived at `3c3d3d53`; re-confirmed unchanged at deployed `6712de1`".
3. **`next_cursor` is dropped silently.** *Measured:* v1 had **no cursor at all**; v2 adds `next_cursor`, and `MODEL_VERSION_LIST_DEFAULT_LIMIT = 50` with the caller (`ServerVersionsSection.tsx:130`) passing no limit. `next_cursor` occurrences in UI `src` = **2**, both in this PR's own fixtures (contrasts in the same sweep: `current_version_id` 7, `full_hash` 7). So beyond 50 versions the user sees a truncated history and **the server now says so on the wire while the adapter ignores it**. Behaviour is identical to v1, so this is **not a regression** — but the PR enumerates the v2 shape field by field and discloses `deduped` while omitting this, and the adapter's own header says a *"silently shortened history would misrepresent what exists to be restored"*. **Remedy:** one comment naming `next_cursor` as knowingly unread, matching the `deduped` treatment, plus a row.
4. **`analysis_state` is new on the v2 restore wire and unread.** Same class as (3); `analysis_state` is already a live concept in the UI (472 occurrences in `src`), so this is a capability question, not a correctness one. **Remedy:** row it; do not expand scope here.
5. **The `restored !== true` guard is unpinned.** *Measured:* removing it leaves **23/23 GREEN** — the only survivor in twelve mutants. **Pre-existing** (`origin/staging:411`, never pinned in either version), so not a regression; the asymmetry is that this PR pinned every *sibling* refusal branch (no receipt, no graph, v1 envelope) and left this one bare, and an uncovered branch is what a later tidy-up deletes. **Remedy:** one test asserting a `restored: false` payload is `unusable`.
6. **Deploy-order note (non-blocking).** v2-only means a CEE rollback past `4c29c5b5` re-breaks version history. It fails closed to the same honest retry state, so it is a broken capability and not a trust defect — worth one sentence in the body naming the window.

---

## CI attribution

Read run-level first: `Contract Validation` **success**, `Staging Tests` **success**. Then per-job at the exact head `464a5c00`: `total_count=14`, no `in_progress`, nothing `cancelled`. **`Staging Gate` — the only required context — SUCCESS.** All four `Full Test Suite` shards SUCCESS; `TypeScript + Lint` SUCCESS. `mergeable: MERGEABLE` (not DIRTY), so the checks genuinely attach to this SHA; `UNSTABLE` is the advisory alone.

**`Visual Regression (advisory)` discharged by control, not by label:** the identical failure is present on staging's own head `98c4dcaa` (run 33027331558). #879 is the open fix.

**The local shard 3/4 failures need no local reproduction from me:** all four shards are **green in CI at this exact SHA**, which is stronger evidence than a local run, and re-running the suite on a contended machine would be could-not-measure in both directions.

**Scope:** measured, not assumed. Base is `staging` `98c4dcaa`; merge-base is `16c55158`, so the branch is behind. Files changed on staging since the merge-base are entirely `pre-analysis-v3` / `conversation` / `v5/blocks` — **intersection with this PR's three files = 0.** Consistent with the author's zero-overlap report for #886/#887.

---

## A correction of my own instrument, disclosed

One of my mutation runs used `perl -0pi -e "s/\Q…/…/"` on an anchor containing `/`. **The perl invocation failed with a syntax error and mutated nothing — and the harness printed `Tests 23 passed (23)`, which is indistinguishable from a surviving mutant.** It was caught only because `applied=0` sat on the same line. A probe that fails must not share a display path with a probe that returns nothing; mine did, for one mutant. I re-ran every mutant under a Python replacer with a per-mutant anchor-count assertion and a `HARD ERROR` path that cannot be mistaken for a result. Every number in this review comes from that re-run.

PR #893, opened 2026-08-27
chore(dead-authority): delete graphPreflight and the selector's write-only leader trio

Talchain:
Deletes two dead authorities. Both were treated as hypotheses and revalidated at the current tip; every number below is reproducible by an outside reviewer without my context.

## Heads measured at

| | SHA |
|---|---|
| **Source tip swept** (`staging`) | `429519be257ce412b655c6a576263cfeb2e0c424` |
| **Deployed bundle swept** (`staging--olumi.netlify.app/version.json`) | `429519be257ce412b655c6a576263cfeb2e0c424` — **identical**, so source and served artefact were the same thing during this measurement |
| This branch head | `b8655b8244d84c951ba51951f4330eb32bea1d7e` |

Re-derive the deploy before relying on the bundle numbers — it will not stay equal.

## How to re-run every sweep

```bash
git clone --filter=blob:none --branch staging https://github.com/Talchain/DecisionGuideAI.git ui
cd ui && git rev-parse HEAD   # must print 429519be...
S=$PWD
# absolute grep: `rg` is a shell function in some environments and is invisible to /bin/sh
/usr/bin/grep -rna --exclude-dir=.git --exclude-dir=node_modules -- "<SYMBOL>" "$S"
```
No `|| true` on any sweep (it turns "tool missing" into "string absent"). No `| head` on any sweep. `-a` throughout, because this repo has NUL-bearing source files that plain `grep` silently skips.

**Bundle sweep** — crawl from `index.html`, BFS over `/assets/*.js`: **81 chunks / 6,000,146 bytes**, then `/usr/bin/grep -laF -- "<literal>" *.js`.

---

# Symbol 1 — `graphPreflight.ts` (+ both its specs)

`src/canvas/validation/graphPreflight.ts`, exporting `validateGraph`, `isWithinLimits`, `getCurrentLimits`, `ensureHydrated`.

**What it claimed to do vs what enforces it.** It is a client-side graph validator on the `/v1/limits` path. Nothing enforces anything through it: its only importers were its own two test files, so it validated nothing a user ever submitted. The live limits path is `limitsStore`, which is untouched.

### Six-way revalidation

| # | Gate | Result | Contrast control, same invocation |
|---|---|---|---|
| 1 | Static imports | **2** — both its own specs (`__tests__/graphPreflight.test.ts`, `__tests__/graphPreflight.spec.ts`). Raw-name sweep = 9 lines total. | `hasLeadingOption` → **226 lines** in the same sweep |
| 2 | Dynamic imports | `import\([^)]*graphPreflight` → **0** | same probe: **722** `import(` occurrences, **28** `lazy(()=>import(`, sample printed |
| 3 | String / template consumers | Raw text sweep over the **whole tree** (not just `src/`): only CI baselines + one archived doc | `AUTHORITY` fires in `analyticalNodeFields.ts:152,158` |
| 4 | Route registration | String-path sweep `validation/graphPreflight` → **3**, all CI artefacts, no registry | `routes/CanvasMVP` → **19** |
| 5 | Flag posture | 0 flag references in the file | `isEnabled(` → **33** repo-wide |
| 6 | Open PR dependencies | **0 of 18** open PRs touch it | verified via `gh pr diff --name-only` on every non-dependabot PR (#892, #879, #622) |

### ⭐ Bundle-presence — the strongest fact here

**Absent from the served artefact.** Four independent distinctive literals, including the loosest containment pattern:

| Literal | chunks | occ |
|---|---|---|
| `is missing a source node` | **0** | 0 |
| `is missing a target node` | **0** | 0 |
| `is missing an ID` | **0** | 0 |
| `Edge at position` ← loosest containment | **0** | 0 |
| *control* `decision_brief` | 4 | 23 |
| *control* `Explore more options` | 2 | 3 |

⚠ **A near-miss worth recording.** `STRENGTH_OUT_OF_RANGE` reads **2 chunks / 4 occurrences** and looks like a PRESENT signal. It is **not distinctive to this module** — it also lives in six live modules (`autoFix.ts:434`, `PreAnalysisGuidance.tsx:37`, `useValidationFeedback.ts:93`, `SummaryTab.tsx:40`, `useDebugData.ts:2330,2350`, `DebugPanelV2.spec.tsx:544`). A shared enum is not a distinctive literal. Had I stopped at that one string I would have wrongly concluded the module was live.

---

# Symbol 2 — the selector's write-only leader trio

`leaderClaimPermitted`, `leaderWithheldReason`, `leaderSeparation` on `ComposedAnalysisState` (`src/canvas/state/analysisStateSelector.ts`).

**What it claimed to do vs what enforces it.** The selector applies the contract's conjunction — `wire.leader_claim.permitted && wire.run_state.kind === 'complete_current'` — and its own comment says it is *"applied HERE, once"*. **Nothing reads the answer.** It is an unenforced mirror: it RECORDS a verdict rather than MAKING one. What actually decides whether a leader renders is `ResultsBody.tsx:369`, from `hasLeadingOption`.

### ⛔ Two different things share this name — the live one is untouched

| | Selector member (**deleted**) | Prop (**left alone**) |
|---|---|---|
| Source | `analysisStateSelector.ts` :266 type, :667 compute, :683 emit | `ResultsBody.tsx:369` ← `verdict?.hasLeadingOption === true` |
| Consumers | none in production | `DecisionBriefSection.tsx` :66,:70,:95,:142,:173,:174,:177,:184 |
| Specs | 3 in the `LEADER CLAIM` describe + 3 stray assertions | 8 refs across two spec files, **all still green** |

### ⭐ Containment argument (this is what makes the zero evidence rather than an absence)

A property can travel into a component via spread without its name ever appearing, so a name sweep alone would not be sound. I enumerated **every** production site that binds the whole composed object and listed **every** line each one uses it on:

- `OutputsDock.tsx:988` → only `.displayedFreshness` (989), `.requiresRerun` (3435), `.semantic` (3436)
- `AnalysisFreshnessNotice.tsx:165` → only `.authority` (174), `.displayedFreshness` (175)
- `useAnalysisRunState.ts:224` → only `.runStateKind` (247), `.displayedFreshness` (256)

**None spreads the object; all access members by explicit name.** Therefore the raw-name sweep *is* containment, and zero production readers is established rather than assumed.

### Six-way revalidation

| # | Gate | Result | Contrast control |
|---|---|---|---|
| 1 | Static imports | Trio never imported; only reached as members | `useModelEditAuthority` → 38 lines |
| 2 | Dynamic imports | `import\([^)]*analysisStateSelector` → **0** | 722 `import(` / 28 `lazy` |
| 3 | String / template | Whole-tree raw sweep: `leaderClaimPermitted` 27 lines (10 selector+spec, 17 live prop chain), `leaderWithheldReason` 5, `leaderSeparation` 3 | `hasLeadingOption` → 226 |
| 4 | Route registration | n/a — object members, no registry | `routes/CanvasMVP` → 19 |
| 5 | Flag posture | No flag gates the selector members | `isEnabled(` → 33 |
| 6 | Open PR dependencies | **0 of 18** open PRs touch the selector | `gh pr diff --name-only`, all non-dependabot PRs |

### ⭐ Bundle-presence — the write is visible and no read exists

The two namesakes **separate cleanly across chunks**, which is the decisive evidence:

`useAnalysisTrust-qQxden7H.js` — **1 occurrence, the selector write, nothing reads it:**
```
source:i??"none",leaderClaimPermitted:null===b?null:b.leader_claim.permitted&&"complete_current"===b.run_state.k
```
`ReactFlowGraph-DigJRBzG.js` — **4 occurrences, all the LIVE prop chain:**
```
function Lce({brief:e,leaderClaimPermitted:a})                                    <- DecisionBriefSection
function Fce({leaderClaimPermitted:e})                                            <- Container
n.jsx(Fce,{leaderClaimPermitted:!0===e.recommendation.verdict?.hasLeadingOption})  <- ResultsBody:369
```
`leaderWithheldReason` → **1 occ (write only)**. `leaderSeparation` → **1 occ (write only)**.
Control — the sibling field from the *same object literal*: `displayedFreshness` → **1 write, 6 reads** (2 chunks / 7 occ). Same file, same sweep, one field reads and the other does not.

---

## Proof the deletion is inert

- **Named gate `pnpm run typecheck` (= `scripts/ci/typecheck-gate.sh`, what the required `Staging Gate` job runs): PASSED.** Coverage phase 3738/3778 files; ratchet 2251 → **2250**.
- The gate itself printed `::notice::1 diagnostic(s) fixed since the identity baseline, none added — regenerate to bank it.` I regenerated with `--update-baseline` **in this PR**, as its own escape-hatch documentation instructs. **The baseline diff is exactly the deleted file's single `TS2322` plus the count decrement — no other churn.**
- **Targeted suites, with `VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` exported** (without them ~24 spec files fail at COLLECT while the run still prints a healthy total): **5 files collected, 5 passed, 81 tests passed, 0 failed.** Each file asserted collected non-zero **by name** — `analysisStateSelector.spec.ts`, `analysisState.threeTruths.spec.tsx`, `useAnalysisRunState.unionWiring.spec.ts`, `DecisionBriefSection.spec.tsx`, `decisionBriefRobustnessCaveat.spec.tsx`.
- **Test-count delta is exactly the intent**: `analysisStateSelector.spec.ts` 27 → 24 `it(` blocks; the three removed are precisely the `LEADER CLAIM` describe. The other two spec files kept **all** their tests (9 → 9, 8 → 8) — only assertions inside surviving tests were removed. The live prop-chain specs (`the caveat is gated on the leader claim`) are **still green**, which is the discriminating check that I did not touch the live path.

## ⚠ Scope limits, stated rather than left to be discovered

- **Newly orphaned by this PR, deliberately NOT included**: `src/adapters/plot/v1/limitsManager.ts`. `graphPreflight.ts:16` was its **sole** importer. It self-hydrates at module load (`:161`), so it is inert only while nothing imports it. Left for a follow-up rather than widening this PR.
- **UNKNOWN, exact scope**: the deleted selector member added a **staleness** conjunct (`run_state.kind === 'complete_current'`) on top of the producer's `permitted`. Whether CEE also refuses a leader on a stale run is a **cross-repo** question I did not measure from this repo. `ResultsBody.tsx:369` applies **no** staleness check. If a stale-leader guarantee is wanted client-side, it must be built on the live prop path — it was never in force through the deleted member, because nothing read it.
- Sweeps covered the **whole working tree** including `e2e/`, `tests/`, `scripts/`, `docs/`, `netlify/`, `public/`, `supabase/` — not only `src/`. They did **not** reach other repos, and a filesystem sweep cannot see an unmerged branch, which is why gate 6 was done through the GitHub API instead.
- The graphPreflight absence rests on **bundle literals plus a direct import-graph read**, not on bundle counts alone. The leader-trio absence rests on **a direct enumeration of every whole-object binder** (above) plus bundle counts.
- `scripts/isolation-results/external-files.txt` still names the deleted files. It is a dated evidence artefact from `scripts/verify-isolation.sh`, not a live gate input, and this estate treats historic capture records as append-only — so I did not rewrite it.


Talchain:
## CI verdict at the exact head `b8655b8244d84c951ba51951f4330eb32bea1d7e`

**13 of 14 checks green, including the required `Staging Gate`.** `mergeable=true`, `mergeable_state=unstable` — never `DIRTY`, so these checks belong to the commit being proposed, not to a stale head.

```
success  Staging Gate                 <- REQUIRED
success  TypeScript + Lint
success  Typecheck Gate Self-Test
success  Full Test Suite (shard 1/4 .. 4/4) + Summary
success  Production Build
success  CEE Contract Validation
success  Install & Cache
success  Security Audit (pnpm)
success  Flag Deployment Drift (report only)
failure  Visual Regression (advisory)  <- discharged below, by NAME
```

### `Visual Regression (advisory)` — discharged by name, not by colour

It **fails on the base tip itself** (`429519be`), so its colour carries no information. Diffing the failing set instead:

| | base `429519be` | this PR `b8655b82` |
|---|---|---|
| totals | **11 failed / 23 passed** | **11 failed / 23 passed** |
| failing names | 10 × `states.visual.spec.ts` (5 states × 2 viewports) + `selftest.visual.spec.ts:160` | **identical, same 11** |

Same count, same names, same specs. The self-test failure is the harness reporting its own noise floor (`0.0569` against a `0.0005` tolerance) — the instrument is declaring itself unable to discriminate, which is why PR #879 exists to regenerate the linux references. Nothing in this PR renders: one deleted module was never imported outside its own tests, and the other change removes object members that no component reads.

### Baseline regeneration absorbed no unrelated drift

Regenerating a baseline can silently bless pre-existing drift as yours, so here is the check rather than the assurance. **Every changed line in `scripts/ci/`, verbatim — four lines:**

```
-# count=2251
+# count=2250
-1  src/canvas/validation/graphPreflight.ts  TS2322  Type '"STRENGTH_OUT_OF_RANGE"' is not assignable to ...
-# count=2251
+# count=2250
-1  src/canvas/validation/graphPreflight.ts
```

Cross-checked by **two independent decompositions**, neither of which restates the gate's own number:

| | rows (`grep -cE '^[0-9]+\t'`) | sum of counts (`awk`) |
|---|---|---|
| `typecheck-baseline.txt` | 556 → 555 (**−1**) | 2251 → 2250 (**−1**) |
| `typecheck-baseline-identities.txt` | 1350 → 1349 (**−1**) | 2251 → 2250 (**−1**) |

`git diff --numstat` is `1 insertion / 2 deletions` per file. A regeneration that had absorbed unrelated diagnostics would show up in all four cells and in numstat; it shows up in none.

### TS6133 discovery pass — no cascade

Ran `tsc -p tsconfig.app.json` after the deletion and read TS6133/TS6196/TS6192 as a work list. **350 total, and the gate independently reports `1 diagnostic(s) fixed … none added`.** The only two hits near the touched files (`graphGuardrails.spec.ts` `assessEdgeDeletion`, `linkifyCoachingText.tsx` `end`) are **pre-existing baseline rows**, not consequences of this change. No new unreferenced chain opened.

⚠ **Stated limit of that instrument**: TS6133 fires on unused *local* declarations. It structurally **cannot** see a module that became orphaned, because an exported const in an unimported file is still "used" as far as `tsc` is concerned. That is why `limitsManager.ts` — orphaned by this PR, `graphPreflight.ts:16` having been its sole importer — was found by hand and is called out in the PR body rather than by the compiler.

Talchain:
## ⭐ Independent review requested — **second Codex session**, bounded scope

Per the ownership split agreed with Paul: **this is a bounded reachability review, not a planning or implementation remit.** Verify only, at the exact head, and return a verdict:

1. **Current-tip reachability** of each deleted symbol
2. **Dynamic and string consumers** — ⚠ `lazy(() => import(...))` defeats a static-import regex, and a **runtime string literal** hid a live reference here this morning that the type system could not see
3. **Route registration** — a route is reached by registration, not import
4. **Build flags** — flag values are **inlined at build time**, so grepping a flag NAME returns a false zero
5. **Open-PR dependencies**
6. **Exact-head verdict** — name the SHA you measured at

**No implementation, no scope expansion, no new candidates.** If a deletion is unsafe, say which and why; if the evidence below does not reproduce, say that.

---

## What this PR deletes — **2 of 6 briefed candidates survived revalidation**

**`graphPreflight.ts` + both its specs.** Only importers were its own tests. **Absent from the deployed bundle** — four independent literals, including the loosest containment, all read 0 while controls fired.

**The selector's leader trio** (`leaderClaimPermitted` / `leaderWithheldReason` / `leaderSeparation`). ⚠ **There is an identically-named live PROP fed from `hasLeadingOption` — the two have never been the same thing, and that is why no sweep ever found the divergence.** The bundle separates them cleanly: the selector write sits alone with **no reader**, while all four live-prop occurrences sit in a different chunk. Control — the sibling field from the **same object literal** — `displayedFreshness`: **1 write, 6 reads**.

## Four candidates dropped on measurement, each with its reason

- **`EDITABLE_FIELD_TABLE`'s pin spec** — premise **refuted**: it is a **live pin-skew tripwire** with its own positive control and a union assertion. Not an orphan.
- **`ModelEditAuthority` / `contracts.ts`** — zero importers **confirmed**, but it is **types-only (0 bundle bytes) and a forward design contract**; a sibling module states *"It implements what `contracts.ts` §2 declared"*. **Unenforced policy is not dead policy.**
- **`AddRow`** — **not superseded.** The else-branch substitutes the *intent* (ask the AI to suggest) but **not the capability** (a human authoring a named node). Fail-closed on a receipt-bearing carrier, deliberately left.
- **`stabilityReadiness.ts`** — reassigned to the lane holding the coupled readiness cluster.

## Three instrument findings that bear on how you verify this

⭐ **A shared enum nearly produced a false "present".** `STRENGTH_OUT_OF_RANGE` reads 2 chunks and makes `graphPreflight` look live — **it actually lives in six other modules.** Bundle presence only means anything for a literal that is **distinctive**, and distinctiveness has to be checked against source. Please apply that to any counter-evidence you find.

⭐⭐ **TS6133 cannot see an ORPHANED MODULE.** An exported const in an unimported file is still "used" as far as `tsc` is concerned. `limitsManager.ts` is orphaned **by this PR** (graphPreflight was its sole importer) and was found **by hand, not by the compiler** — reported here, deliberately **not** bundled in.

⚠ **A probe failure nearly became a measured zero**: an unquoted `--include` glob under zsh printed *"no matches found"* and the count read 0. Re-run quoted with a positive control (722 `import(`, 28 `lazy`).

Baselines absorbed nothing — four changed lines, cross-checked by **two independent decompositions**. `Visual Regression` discharged **by name**: identical 11 failed / 23 passed against base, not by colour.

Talchain:
# Independent adversarial review — `b8655b82`

**Verdict: FIX-FIRST.** One blocker, and it is *not* about the deletion — the deletion itself survives every gate I could point at it. The blocker is a **baseline-count collision with `staging`, which has moved 4 commits since this PR's base**. It REDs the sole required check. Remedy is one command, verified by execution below.

Clone `/private/tmp/rev893-adv-b8655b82-…/ui`, `HEAD` asserted `b8655b8244d84c951ba51951f4330eb32bea1d7e`, clean tree. All sweeps `/usr/bin/grep` absolute (`rg` **and** `grep` are both zsh functions on this box). No `|| true`, no `| head`.

---

## ⛔ GATE 0 (BLOCKER) — the typecheck baseline silently mis-merges with current `staging`

`origin/staging` is now `56059e18`, **4 commits ahead** of this PR's base `429519be`: #895, #894, #896, #892.

Both sides independently rewrote the **same header line** `# count=2251` → `# count=2250` in `scripts/ci/typecheck-baseline.txt` *and* `-identities.txt`, for **different reasons** (staging removed #894/#896's errors; this PR removed graphPreflight's single `TS2322`). Git sees the identical literal `2250` on both sides, **auto-merges it with no conflict**, and applies **both** row deletions.

**Measured by trial merge** (`git merge origin/staging` into `pr893` — clean, 0 conflicts):

```
::error::Baseline is internally inconsistent: '# count=2250' but rows sum to 2249.
::error::Regenerate with --update-baseline instead of editing by hand.
pnpm run typecheck EXIT=1
```

**Required-check impact.** `staging` protection requires exactly `["Staging Gate"]`. That aggregator `needs: [tsc, typecheck-selftest, vitest, vitest-summary, build]`, and the `tsc` job runs `pnpm run typecheck` (`staging-full-tests.yml:114`). So this reds the one check that gates the merge.

**Why this PR's 13 green checks do not cover it** (trap 24). The workflow triggers on `pull_request: branches: [staging]` with `actions/checkout@v4` and **no `ref:` pin**, so CI tests the **merge ref**. The green `Staging Gate` completed **10:18:42Z**; all four staging commits landed *after* it (10:37:54Z, 10:49:32Z, 11:29:13Z, 11:29:17Z UTC). **The green is about a merge ref that no longer exists.**

**Smallest remedy — verified by execution in the trial merge:**
```bash
git merge origin/staging          # or rebase; merges clean, 0 conflicts
bash scripts/ci/typecheck-gate.sh --update-baseline
```
→ `Baseline regenerated: 554 file(s) / 2249 error(s)`; re-run `pnpm run typecheck` → **PASSED**.

Reassurance on the rest of that delta: the 4 new staging commits introduce **zero** references to any deleted symbol (all 10 targets = 0; contrast `ROBUSTNESS_LEVEL_COLOURS`=2, `lucide`=17 — **fired**). `OutputsDock.tsx` is touched by both, but in unrelated regions. **No semantic conflict — this is purely the count header.**

---

## The six-way deletion gate — each measured separately

| # | Gate | graphPreflight | Selector trio | Contrast control (same run) |
|---|---|---|---|---|
| 1 | Static imports | **0 code importers.** 3 hits total: 1 archived doc, 2 in a *generated* artefact | **0 readers** of the members | `graphguardrails` → 12 files ✔ / `robustnessAggregateLevel` → 4, `factorsThatFlipLeader` → 3 ✔ |
| 2 | Dynamic / string | 0 — all "preflight" hits are Tailwind `corePlugins`, archived `AskPreflightError`, docs prose | n/a (object members) | — |
| 3 | Registry / route | 0; no barrel exists in `src/canvas/validation/` (`git ls-tree` at base) | n/a | — |
| 4 | Build flags | 0 in `netlify.toml` + `src/flags.ts` | 0 | 41 `VITE_` entries in `netlify.toml` ✔ |
| 5 | Bundle | design verified offline (below) | weak by construction — superseded by containment proof | — |
| 6 | Open PRs | **0 of 19** | **0 of 19** | see note ⚠ |
| 7 | Age | **ADDED 2025-11-11**, 4 commits ever, cold since 2026-01-19 → safe-dead ✔ | **added 2026-08-17 (#737) — TEN DAYS** ⚠ half-wired shape, see below | — |

**Sweep-form note.** My repo-wide sweeps are *not* line-anchored to `^import`, so **multi-line imports are covered by construction** — the `from '…graphPreflight'` line is itself matched.

**Alias/barrel escape hatches closed.** `tsconfig.app.json` paths = only `{"@/*":["src/*"]}`; vite alias = `@/`→`src/` plus Supabase/react-query stubs. Every alias form still contains the literal `graphPreflight`.

⚠ **My own instrument failure, disclosed.** My *first* gate-6 contrast control (`hasLeadingOption`) **did not fire** — 0 across all 19 diffs — which made the absence claim unsupported. I re-ran with proven-sighted controls: `optionCoverage` = 8 hits (#898), `analysisNewCopy` fires (#897), `diff --git` present in 19/19. Probe proven sighted, targets re-asserted at 0. (`hasLeadingOption` is genuinely absent from the *changed hunks*, though present in the tree.)

⚠ **Symbol-name greps are NOT evidence for graphPreflight** — recorded so nobody re-derives from them. 2 of its 6 exports are **name-twins that survive elsewhere**: `validateGraph` (2 other declarations — `graphValidator.ts:14`, `graphToRunRequest.ts:76`) and `ValidationResult` (4 others). Only **path-based** evidence counts here. The other four (`isWithinLimits`, `getCurrentLimits`, `ensureHydrated`, `ValidationViolation`) are now zero-declaration **and** zero-reference — the 5 `CommandPalette.tsx` hits are substring false positives from a local `setValidationViolations`, typed `ValidationError[]`.

---

## ⭐ "Write-only" — PROVEN by execution, not by reading

This is the claim that mattered most, and it holds. I re-derived the containment argument independently rather than inheriting it.

The trio is reachable only as a member of `ComposedAnalysisState`. The pure core `composeAnalysisState` has **no production caller outside its own module**. The public accessor is `useAnalysisState()`. I enumerated **all six** production call sites — the PR body lists the 3 *binders*; there are **3 more** that access a member inline (worth recording, though harmless — an inline access cannot bind, therefore cannot spread):

- **Binders:** `OutputsDock.tsx:988` → `.displayedFreshness`, `.requiresRerun`, `.semantic` · `useAnalysisRunState.ts:224` → `.runStateKind`, `.displayedFreshness` · `AnalysisFreshnessNotice.tsx:165` → `.authority`, `.displayedFreshness`
- **Inline (not in the PR body):** `useEditConfirmation.ts:21` `.displayedFreshness` · `useAnalysisDisplayState.ts:32` `.displayState` · `useAnalysisTrust.ts:51` `.trust`

**Explicit spread check:** `...composed | ...composedAnalysisState | ...useAnalysisState` over `src/` → **0 occurrences**. **Contrast:** spread-form `\.\.\.[a-zA-Z]` in `src/` → **5558**. Probe sighted. Every use is an explicit named member access.

**The name-twin is real and correctly left alone.** `ResultsBody.tsx:369` feeds a prop of the identical name from `verdict?.hasLeadingOption === true` into `DecisionBriefSection`. Different source, different question, untouched by this PR — and its specs (`decisionBriefRobustnessCaveat`, `DecisionBriefSection`) are **green in my run**, which is the discriminating check that the live path was not disturbed.

---

## The staleness conjunct — I confirm the lane's own disclosure

The deleted `LEADER CLAIM` describe pinned three real properties, incl. *"WITHHOLDS a permitted leader on a stale run"*, plus a `RUNTIME FLOOR` assertion that an unknown kind permits no leader.

**My independent measurement:** `src/lib/decisionVerdict.ts` — the surviving authority behind `hasLeadingOption` — contains **zero** occurrences of `run_state|complete_stale|stale|freshness|Freshness`. Contrast `separation` = 13 in the same file, **fired**. So the surviving authority genuinely does **not** carry the staleness conjunct.

**This is not concealment** (trap 8): the PR body states it explicitly under *Scope limits → "UNKNOWN, exact scope"*, including *"`ResultsBody.tsx:369` applies no staleness check"* and *"it was never in force through the deleted member, because nothing read it."* My measurement **confirms** the disclosure. Both are true: nothing regresses, **and** a designed-and-tested staleness guarantee leaves the tree with no successor.

**Ask, not a blocker:** given gate 7 flags this trio as **10 days old** (#737, *"consumers migrated"*) — a half-wired shape, not a dead one — the loss should land in `ROADMAP.md` as a row rather than only in this PR body, or the next session re-derives it from scratch. That is the estate's chronic failure #1, and a deletion is the moment it goes invisible.

---

## Gate runs (my own, this clone)

- `pnpm run lint` → **PASS**, 0 errors / 1172 warnings (pre-existing); hooks-ratchet 229 across 13 files, exactly matching baseline. *(CI runs Lint **before** typecheck — trap 22e; both run.)*
- `pnpm run typecheck` **at this head** → **PASS**. Coverage 3738/3778; ratchet measured **2250** vs baseline **2250** — exact, which independently validates the 2251→2250 edit *in isolation*.
- Targeted specs, env vars exported, 5 files named and quoted → **5 collected / 5 passed / 81 tests / 0 failed**. No shrunk collect: each named file shows non-zero tests in verbose output.
- **Not run locally** (per instruction): full suite, e2e, build. CI owns those.
- CI on the exact head: `total_count=14`, success 13, **failed 1**, **running 0**. The failure is `Visual Regression (advisory)` — **not** in the required contexts, with #879 already open against it. Not caused by this PR.

## Residual — confirmed, and slightly *understated* by the PR body

`src/adapters/plot/v1/limitsManager.ts`: at the base its only importers were `graphPreflight.ts:16` **and** `graphPreflight.test.ts:7`. After this PR it has **zero importers repo-wide** — fully orphaned, still carrying 3 baseline TS errors, still self-hydrating at module load (`:161`), which now never executes. Live path is `limitsStore` (6 files, contrast fired). Correctly scoped out; worth a row.

## What my evidence EXCLUDES / could not reach

1. **The deployed bundle.** `version.json` has moved to `56059e18` (11:32:26Z), so the PR's bundle numbers are scoped to `429519be` and I could not re-measure them. I verified the probe *design* offline instead: all four distinctive literals are **unique to `graphPreflight.ts`** at the base, and the disclosed near-miss is accurate — `STRENGTH_OUT_OF_RANGE` lives in 8 files.
2. **Deployed dashboard env** (trap 18) — I derived flags from `netlify.toml`/`src/flags.ts` only. The narrower sufficient claim: **no consumer exists in source**, and a flag switches between existing code paths, it cannot create a reader.
3. **Other repos.** The staleness question is cross-repo — whether CEE refuses a leader on a stale run is unmeasured from here.
4. **No meaningful mutant.** This is a pure deletion; the discriminating evidence is the surviving live-path specs staying green while the deleted-field specs go with the field. I did not invent a mutant for it.

**Bottom line:** the deletion is sound and unusually well evidenced — every gate I pointed at it holds, and the two things I found that the body had not stated (the 3 extra inline call sites, the *full* orphaning of `limitsManager`) both point the same way. **Fix the baseline against current `staging` and this is good to merge.**

Talchain:
## ⚠ Correction to my review above — one claim withdrawn, the blocker stands

I asserted that *"a fresh CI run right now would fail"* because CI tests the merge ref. **I then measured it, and that claim is not proven. Withdrawing it.**

I fetched GitHub's own merge ref instead of relying on my hand-merge:

```
refs/pull/893/merge = 6f2fb14e  "Merge b8655b82 into 429519be"
parents: 429519be (the OLD base) + b8655b82
```

**GitHub has not recomputed the merge ref against current `staging` (`56059e18`).** Running the gate on that ref: header `2250`, measured `555 files / 2250 errors` → **PASSED, exit 0**. So the PR's checks are green about a *genuine* merge — just one computed against the old base.

**What still stands, and is still measured:** merging against **current** `staging` produces header `2250` vs rows summing to `2249`, and `pnpm run typecheck` exits **1** with `Baseline is internally inconsistent`.

**So the corrected mechanism is:** the failure materialises when the merge is computed against current `staging` — i.e. **at merge time**, where the `push: branches: [staging]` trigger runs the same `tsc` job on `staging` itself — or earlier, as soon as GitHub refreshes the merge ref. Either way **`staging` goes red on the required gate after this lands**, which is the outcome that matters.

The remedy is unchanged and still verified by execution:
```bash
git merge origin/staging          # clean, 0 conflicts
bash scripts/ci/typecheck-gate.sh --update-baseline   # -> 554 files / 2249 errors
pnpm run typecheck                # -> PASSED
```

Nothing else in my review depends on the withdrawn sentence: the six-way gate, the containment proof, the staleness measurement and the `limitsManager` residual were all derived independently of it. **Verdict is unchanged: FIX-FIRST on the baseline, then good to merge.**

PR #894, opened 2026-08-27
A5 — delete the two dead confidence→readiness mappers and the five-link chain behind them

Talchain:
Deletes two same-named `mapConfidenceToReadiness` functions, the never-mounted component carrying one of them, an orphan module, and the five-link chain that fed the other. **10 files, 2 commits, +59 / −978.** No behaviour change — nothing deleted here reached a surface.

Measured at **`6a11402c`** (deletion train `bae9f01b` + orphan-residue removal `6a11402c`), based on staging `429519be257ce412b655c6a576263cfeb2e0c424`. Every command below is re-runnable at that head.

---

## The load-bearing claim, and how to re-derive it

The one thing a reviewer cannot re-derive quickly: **`decisionReadiness` never reached a surface.** At the pristine base it had exactly two occurrences in `OutputsDock.tsx` — its own definition (`:1619`) and `hasDecisionReadiness: !!decisionReadiness` (`:1715`), inside a `verboseDebug('[TrustSignals] OutputsDock', …)` call. The boolean was coerced for a debug line and discarded.

```bash
git checkout 429519be
/usr/bin/grep -n 'decisionReadiness' src/canvas/components/OutputsDock.tsx
# => exactly 2 lines: 1619 (definition), 1715 (inside verboseDebug)

# CONTRAST CONTROL — the probe must discriminate, not return 2 for everything:
/usr/bin/grep -c 'ceeDegraded' src/canvas/components/OutputsDock.tsx    # => 4
```

That is what makes this a dead computation rather than a lightly-used readiness authority.

---

## ⭐ The technique that found the rest: the compiler, not a grep

**My own word-grep got this wrong and a gate caught it.** Checking whether `verdict` was still used, `/usr/bin/grep -c '\bverdict\b'` returned **38** and read as *"still used"* — they were comments and `verdict?.verdict`. **`tsc` disagreed:**

```
OutputsDock.tsx(1591,9): error TS6133: 'verdict' is declared but its value is never read.
```

A text match binds by a **value predicate anything can satisfy**; the compiler binds by **identity**. Removing one link surfaced the next, one `TS6133` at a time, until the whole chain was out:

```
canonicalBands → mostLikelyValue → verdict → readinessFromConfidence
  → decisionReadiness → a debug boolean
```

Five derivations terminating in a coerced boolean. **Nobody had scoped this chain** — it came out only because the compiler was allowed to name each link. Recommended technique for any dead-code removal: delete the leaf, re-run `pnpm run typecheck`, repeat until clean. Do not trust a symbol grep for a reachability claim.

### ⚠ The bound on this technique — TS6133 cannot see an ORPHANED MODULE

Stated because the next reader will otherwise trust it further than it goes. `tsc` reports unreferenced symbols **within the reachable module graph**. An **exported** const in a file nothing imports is still "used" as far as the compiler is concerned — it cannot see a whole module that has fallen out of the graph.

**This fired inside this very PR.** Deleting `stabilityReadiness.ts` orphaned `MIN_STABLE_RECOMMENDATION_STABILITY` in `components/results/constants.ts`. The typecheck gate was **green at exactly baseline with that constant already dead**. It was found by a hand sweep, not by the gate (commit 2).

So the pairing is:
- **TS6133** finds dead **symbols inside live modules**.
- **A reference / bundle sweep** finds **dead modules and dead exports**.

**Neither alone is complete.** Running only the first would have left residue behind.

> ⚠ And a bundle-presence sweep only means something for a **distinctive** literal — a shared string can make a dead module look live. Check distinctiveness against source before concluding anything from a hit count.

`interpretOutcome.ts` was checked for exactly this after its `deriveVerdict` import was removed: **3 production importers remain**, so it is not orphaned. `Tooltip` (57), `IdentifiabilityBadge` (1), `EvidenceCoverage` (1) likewise all retain production importers.

**Related stale-prose finding:** `verdict`'s own comment said *"Compute verdict for VerdictCard"*. **No VerdictCard consumed it** — it had been orphaned by an earlier change and kept alive solely by the dead mapper. `VerdictCard.spec.tsx` (7 tests) is green after removal. A comment asserting a consumer that had already gone.

---

## What was deleted, and the evidence for each

| File | Lines | Evidence |
|---|---|---|
| `canvas/components/DecisionReadinessBadge.tsx` | 338 | **Zero production importers repo-wide.** Every reference was its own spec, CI baselines, or docs. Carried the second `mapConfidenceToReadiness` — **different signature, different logic** (`level.toUpperCase()` vs lowercase + an extra `supportsObjective` parameter). |
| `canvas/components/__tests__/DecisionReadinessBadge.spec.tsx` | 353 | Spec for the above. |
| `canvas/utils/mapConfidenceToReadiness.ts` | 79 | One importer (`OutputsDock.tsx:133`), traced above to a debug line. |
| `canvas/utils/__tests__/mapConfidenceToReadiness.test.ts` | 64 | Spec for the above. |
| `components/results/utils/stabilityReadiness.ts` | 80 | Zero references repo-wide, case-insensitive, all file types. |

Two same-named functions, **neither mounted**. This was never a convergence problem.

---

## Six-way revalidation — re-runnable, with the controls and what they returned

Run each at `429519be`. `rg` is a shell function in some environments and invisible inside `/bin/sh`, so these use `/usr/bin/grep` absolutely. No `|| true`, no piping a sweep through `head`.

**(1–3) Static imports, dynamic imports, string/template consumers** — one case-insensitive, all-filetype, whole-repo sweep covers all three:

```bash
for sym in mapConfidenceToReadiness DecisionReadinessBadgeCompact DecisionReadinessBadge \
           assessStabilityReadiness StabilityReadiness readinessFromConfidence; do
  echo "=== $sym"
  /usr/bin/grep -rni "$sym" . --exclude-dir=.git --exclude-dir=node_modules --exclude-dir=dist -l | sort -u
done
```

Returned: every hit confined to the deletion set plus docs/CI baselines. `assessStabilityReadiness` and `StabilityReadiness` appear **only in their own file**.

> ⚠ Case matters here. An earlier lowercase-only sweep for `stabilityReadiness` would have missed `StabilityReadiness` and `assessStabilityReadiness`. Use `-i`.

**Contrast control, same invocation** (an absence claim needs a probe shown capable of returning presence):

```bash
/usr/bin/grep -rni "readinessObjectsToRun" . --exclude-dir=.git --exclude-dir=node_modules --exclude-dir=dist -l
# => 10 files. Non-empty, so the sweep is not blind.
```

**(4) Dynamic import / route registration** — a static-import regex is defeated by `lazy(() => import(...))`, so probe that form directly:

```bash
/usr/bin/grep -rniE "import\([^)]*(DecisionReadinessBadge|mapConfidenceToReadiness|stabilityReadiness)" \
  . --exclude-dir=.git --exclude-dir=node_modules --exclude-dir=dist
# => empty

# CONTROL — the same pattern must find real lazy routes:
/usr/bin/grep -rnoE "lazy\(\(\) => import\([^)]*\)" src --include='*.tsx' --include='*.ts' | head -5
# => 5 real lazy routes in src/poc/AppPoC.tsx. The probe works.
```

**(5) Flag posture** — moot by containment: a component with **zero importers** cannot be mounted under any flag posture. No flag can reach it.

**(6) Open PR dependencies** — open UI PRs at time of measurement: #892 (`analyticalNodeFields.ts` only), #879 (visual-regression PNGs + playwright config), #622 (dependabot config), remainder dependabot.

```bash
for p in 892 879 622; do
  gh pr diff -R Talchain/DecisionGuideAI $p --name-only \
    | /usr/bin/grep -iE 'DecisionReadinessBadge|mapConfidenceToReadiness|stabilityReadiness|OutputsDock|ds-compliance-baseline|typecheck-baseline'
done
# => empty for all three. No overlap.
```

**Importer counting** used module-specifier form (static + dynamic, any relative depth), **counting importers, not references** — a reference count is not a consumer count. Validated by a containment argument: the strict specifier set must be a subset of a looser bare-name set. **0 containment violations**, and the loose-minus-strict remainder was exactly the definition file plus comment mentions.

---

## Not touched, deliberately

**The one gate authority is `readinessObjectsToRun`** (`canvas/utils/canRunAnalysis.ts:440`), where the CEE producer verdict **supersedes** the side-car rather than being OR'd with it — converged 19 Aug 2026, four call sites, pinned **in both directions** by `canonicalReadinessAuthority.spec.ts` (35 tests, green here). **Nothing deleted in this PR fed it.**

`CANONICAL_EDIT_AUTHORITY` and the inspector field tables are untouched. They answer different questions from each other and are not duplicates; that analysis is a separate follow-up.

---

## Baselines — handled in the same train, so no baseline lists a deleted file

**`tools/ci-guards/ds-compliance-baseline.json`.** `ci:guard:ds:update` was tried **and reverted**: it rewrote **1429 lines across files this PR never touched** (`legacy-tailwind-token` total 2165→1638). That is pre-existing repo drift, and blessing it here would have signed this PR's name to it. Hand-edited instead, after proving safety:

```bash
# ROUND-TRIP CONTROL — parse and re-stringify with NO changes; diff must be empty
node -e 'const fs=require("fs"),p="tools/ci-guards/ds-compliance-baseline.json";
         fs.writeFileSync(p,JSON.stringify(JSON.parse(fs.readFileSync(p,"utf8")),null,2)+"\n")'
git diff --stat tools/ci-guards/ds-compliance-baseline.json   # => empty. Formatting cannot churn.
```

Removed the `byFile` entry (**12**) and the three signature blocks (**5 + 5 + 2 = 12**). Those two numbers are derived independently and **agree** — a genuine cross-check, not a restated number. Then corrected `legacy-tailwind-token.total` 2165→2153 to restore the `total === sum(byFile)` invariant that **every other class in the file already satisfies** (verified: `production-hex`, `uppercase-text`, `panel-typography-scoped`, `emoji-icon` all match exactly).

**`scripts/ci/typecheck-baseline{,-identities}.txt`.** Removed the row and corrected the `# count=` header 2251→2250. The gate is fail-loud on this and caught the header when I first missed it (*"Baseline is internally inconsistent: '# count=2251' but rows sum to 2250"*). Regenerating with `--update-baseline` would have absorbed **~221 unrelated pre-existing diagnostics**.

**✅ `scripts/isolation-results/external-files.txt` left alone on purpose.** It still names two deleted files. It is a **generated run record** — written by `scripts/verify-isolation.sh` (`mkdir -p "$ISOLATION_DIR"`), regenerated on the next run. **Hand-editing it would falsify evidence of a run that actually happened.** A dated capture is append-only. Please do not "tidy" it in a later pass.

---

## Gates — all foreground, all at this head

| Gate | Result |
|---|---|
| `pnpm run typecheck` | **PASSED** — 555 files / 2250 errors, **exactly** baseline |
| `pnpm run lint` | **PASSED** — 0 errors; hooks ratchet 229 violations / 13 files, exactly baseline |
| `node tools/ci-guards/check-ds-compliance.mjs --enforce` | **PASSED** — no net-new ratcheted violations |
| `vitest run src/canvas/components/__tests__/ …canonicalReadinessAuthority.spec.ts` | **167 files / 1802 passed / 15 skipped / 0 failed** |

The lint gate is `npx eslint . && npm run lint:hooks-ratchet` — **both steps run**, because a green local subset that skips step 1 is a different check, not a partial one.

**Collected counts asserted BY NAME** (a suite total, a green exit code and a zero-failure line are all consistent with a new spec collecting zero):

```
canonicalReadinessAuthority.spec.ts             35 tests
analyticalNodeFields.registry.spec.ts           26 tests
useInspectorMutations.writtenFields.spec.tsx     6 tests
mutationAuthority.spec.ts                        2 tests
OutputsDock.bandslessReport.spec.tsx             2 tests
OutputsDock.staleVerdictCopy.spec.tsx            4 tests
VerdictCard.spec.tsx                             7 tests
```

UI suites were run with `VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` set — confirmed live by `SUPABASE_BINDING: real` in the run output. Without them ~24 spec files fail at COLLECT while the run still prints a healthy total and zero failures.

---

## Reviewer's shortest path to refuting this

1. `git checkout 429519be` and run the two-occurrence grep on `decisionReadiness` with the `ceeDegraded` control. If it renders anywhere, this PR is wrong.
2. Run the case-insensitive whole-repo sweep with the `readinessObjectsToRun` contrast control. If any deleted symbol has a production consumer, this PR is wrong.
3. `git checkout bae9f01b` and run `pnpm run typecheck` + `pnpm run lint`. Both must be exactly at baseline.

---

## Deliberately left alone (scope discipline)

- `ROBUSTNESS_LEVEL_COLOURS` and `SWITCH_PROBABILITY_THRESHOLD` in `components/results/constants.ts` also read **zero consumers** — but they were dead **before** this PR. Not residue from this deletion; folding them in would put pre-existing drift under this train's evidence. Named in a file comment for whoever prunes next.
- `scripts/isolation-results/external-files.txt` — generated run record, see above.
- `CANONICAL_EDIT_AUTHORITY`, the inspector field tables, and `readinessObjectsToRun` — untouched.

---

## CI status at `6a11402c` — and the one red, pre-derived

**`Staging Gate` (the ONLY required check on `staging`) = SUCCESS.** All four `Full Test Suite` shards green, `TypeScript + Lint` green, `Production Build` green. 13 of 14 checks green.

```bash
gh api repos/Talchain/DecisionGuideAI/branches/staging/protection --jq '.required_status_checks.contexts'
# => ["Staging Gate"]
```

**The one red is `Visual Regression (advisory)` — pre-existing, not caused by this PR.** Derived rather than assumed:

```bash
# The SAME check on the pristine base commit, with none of this PR's changes:
gh api "repos/Talchain/DecisionGuideAI/commits/429519be257ce412b655c6a576263cfeb2e0c424/check-runs" \
  --paginate | jq -rs '[.[].check_runs[]]|.[]|select(.status=="completed")|"\(.conclusion)\t\(.name)"' | sort
# => failure  Visual Regression (advisory)
# => success  (all 13 others)
```

Identical pattern at base and at this head. Two further tells from the job output, read directly rather than inferred from a registry:

1. **The harness's own self-test fails** — `selftest.visual.spec.ts › visual harness self-test — proves the instrument can fail › tolerance separates antialiasing noise from a real regression`. When the instrument's own control fails, its verdicts are not evidence about the product.
2. **All 10 founder-state captures fail uniformly** at both viewports. A per-item probe returning the same answer for every item is evidence about the probe, not the data.

Open PR #879 (*"retire the stale linux references so CI can regenerate them"*) addresses exactly this. Independently, this PR deletes a component with **zero importers** and a chain terminating in a debug call — it cannot alter rendering.


Talchain:
## Merging. Delta verified at the bytes.

Ten files. **Every non-spec source file reads ZERO non-comment added lines** — `OutputsDock.tsx` (+43/−34), `constants.ts` (+12/−5) and the three removed modules all come back 0, with the control firing at 1 on each typecheck-baseline count line. **Pure deletion plus explanatory comment.** `Staging Gate` **completed/success**, `not_completed=0`, nothing based on this branch (control 4).

⭐⭐ **The TS6133 correction fired inside your own train, which is the best possible demonstration of it.** Sweeping before pushing found `MIN_STABLE_RECOMMENDATION_STABILITY` orphaned by the `stabilityReadiness.ts` deletion — **while the typecheck gate was green at exactly baseline with that constant already dead.** The compiler could not name it; a hand sweep could. The bound now stated in the body is the right one and I want it kept:

> **TS6133 finds dead symbols inside live modules · a reference/bundle sweep finds dead modules and dead exports · neither alone is complete.**

⭐ **And checking every module the deleted files imported — `interpretOutcome` (3), `Tooltip` (57), `IdentifiabilityBadge` (1), `EvidenceCoverage` (1), none orphaned — is the half most deletions skip.** A deletion that silently strands its neighbours is how the next dead module is created.

✅ **Refusing to absorb `ROBUSTNESS_LEVEL_COLOURS` and `SWITCH_PROBABILITY_THRESHOLD` is exactly right.** They read zero consumers but were dead **before** this PR; folding them in would put pre-existing drift under this train's evidence and make a clean deletion arguable. Named in a comment and left for the delete-first lane.

✅ **`Visual Regression` discharged by derivation, not by pointing at an open PR**: the same check fails on the pristine base `429519be` with all 13 others green. And two tells from the job output itself — **the harness's own self-test (*"proves the instrument can fail"*) is among the failures**, and all 10 captures fail uniformly at both viewports. *An instrument whose own control fails is not producing evidence about the product.*

PR #897, opened 2026-08-27
Print each coaching sentence once, and stop cutting words in half

Talchain:
Two render defects on the **Analysis (New)** A/B surface, both measured at the DOM by a separate lane on a pinned quartet, both fixed at the **consumer** rather than the producer.

## 1. Each bias card printed its sentence twice

`strengthen/buildRecommendations.ts:259-260` puts the producer's body on **both** display fields by design — `signal: item.signal ?? item.body`, `whyNow: item.body` — and a producer `signal` is carried today on only one deterministic nudge. **So for every other item, bias cards included, the two fields hold the same string by construction.**

That file's own comment records the contract: *"The PANEL dedupes display: an open row renders the body once, in full."* The old panel does. This surface concatenated unconditionally.

**Measured:** 413 characters for a ~205-character sentence, while the same sentence appeared **exactly once on the old tab, in the same DOM, at the same moment.** That contrast is what proves it is the render path and not the data.

`strengthenWhyLine` dedupes at the render site on **exact equality**. A fuzzy or prefix match would be this surface judging whether two producer strings *mean* the same thing, which it cannot do honestly.

**`buildRecommendations` is deliberately unchanged** — it is correct as written for a consumer that dedupes. A producer that is right for its existing consumer should not be bent to suit a new one that skipped the contract.

## 2. Uncertainty headlines were cut mid-word

`buildAnalysisNewViewModel.ts:413` used a bare `text.slice(0, 80)`.

**Measured:** two distinct items cut at exactly 80 characters, with every CSS cause excluded (`line-clamp: none`, `overflow: visible`, `scrollWidth === clientWidth`). Natural sentences do not land on 80 twice.

`truncateAtWordBoundary` keeps whole words and appends an ellipsis, **so a reader can tell a cut string from a finished one.**

### Severity — withdrawn, then reinstated by measurement

I first argued this was the product emitting broken reasoning. I then **withdrew that**, because the same finding carries `implication: u.suggestion || text` on the same row, so with no suggestion the full sentence would survive.

**It has since been measured, and the stronger claim holds.** `u.suggestion` is present on every row and is the constant string `"Review this assumption"` — so the implication carries a **remedy, never the sentence**, and the truncated headline is the only place the original reasoning ever appears. 3/3 rows, not recoverable by expanding the disclosure, and the sentence continues nowhere else on the page (contrast control fired).

⚠ **Bounded:** all three rows are one finding type (`uncertainty:SENSITIVE_ASSUMPTION`). Measured **for that type**. A finding type carrying no suggestion could still render its full text.

## Tests

Bind by **identity, not by the measured numbers** — the doubling test counts occurrences of the sentence itself, so a two-character reword cannot flip it. Truncation asserts *the surviving last word is a whole word of the original*, not a length.

It also asserts **two genuinely different sentences are still joined**, each appearing once: a dedupe that swallowed real prose would be worse than the defect it fixes, and the fix's own success criterion is silent about that.

- **65 tests green** across the surface, including the 4 pre-existing spec files
- Both fixes **proven to bite by revert**, controls green either side
- `shell-conformance` guard green (52/52)

## Not done here

Rowed as **2.1329**: three truncation helpers now exist, one duplicated verbatim across two node files with its own heuristic and its own callers. Converging them would quietly change what canvas nodes display, which is not a render fix's business.

**No DOM-level test.** The measurements above were taken by a separate lane's battery, which now pins both defects continuously.

*Verified by: the author. Independent review required before merge — I do not merge my own work.*

### DIFF
--- a/docs/CC2-EVIDENCE-2.1326-2.1328.md
+++ b/docs/CC2-EVIDENCE-2.1326-2.1328.md
@@ -1,0 +1,82 @@
+# 2.1326 build evidence — `optionCoverage`
+
+**Clone:** `/private/tmp/cc2-ui-sa` @ `f287c012` · **26 Aug 2026** · CC2
+
+## Targeted spec
+`npx vitest run src/components/results/utils/__tests__/optionCoverage.spec.ts`
+→ **1 file, 10 tests, 10 passed, exit 0.** Asserted BY NAME and by expected count — never inferred
+from a suite total (a new spec collecting `(0 test)` is invisible to every aggregate).
+Env: `VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` set.
+
+## Mutation kit — 8/8 as expected, exit 0
+Isolation **proven by WRITING a sentinel**, not by locating a path: inodes compared, sentinel
+confirmed not to propagate into the source. Restores read a **pristine archive outside the repo**,
+never the mutated tree. Every mutant asserts its anchor is **unique** (an unapplied mutation is
+indistinguishable from an equivalent one) and that it **applied**. Control reads exactly `0`;
+trailing control GREEN.
+
+| id | mutation | expect | got |
+|---|---|---|---|
+| M1 | truthiness instead of key presence | RED | RED |
+| M2 | drop the denominator dedupe | RED | RED |
+| M3 | `kind` always `even` | RED | RED |
+| M4 | accept a single option | RED | RED |
+| M5 | provisional on every reading | RED | RED |
+| M6 | set/unset swapped | RED | RED |
+| **M7a** | **reverse `perOption` ORDER** | **GREEN** | **GREEN** |
+| **M7b** | **assign a different option id** | **RED** | **RED** |
+
+**M7a/M7b are a discriminating PAIR, and neither alone proves anything.** M7a green shows the test
+does not bind by position; M7b red shows it still discriminates. A single biting mutant proves
+sensitivity to *something*; the pair proves sensitivity to *the named object*.
+
+## ⚠ M1 SURVIVED ON THE FIRST RUN, AND THAT IS THE MOST USEFUL THING HERE
+The first fixture wrote each intervention as `{ normalised_value: 0 }`. **The wire sends a BARE
+NUMBER** — `{"6886a726": 0}` — with `intervention_details` as the separate object map (re-read at the
+capture, not guessed). An object is always truthy, so the truthiness mutant survived: the
+*"counts ZERO as SET"* test asserted exactly the right thing and **could not fail**.
+
+**The assertion was right, the fixture made it vacuous, and only the mutant could see that.** The spec
+header had itself quoted *"a fixture you wrote yourself is not evidence about the wire"* — while
+getting the shape wrong. Corrected to the wire shape; M1 now bites. The note is kept in the spec so
+the next reader knows why the shape is spelled that way.
+
+## Process, stated accurately
+**This was NOT written RED-first** — the module came before the spec. Rather than claim an ordering
+that did not happen, the evidence is the mutation kit above: every assertion is shown to fail when
+the behaviour it names is removed, and the identity binding is proven with a discriminating pair.
+
+## Not yet done
+- The render half (`V5AnalysisResultBlock`) and its wiring to the readiness slice.
+- The repo's named typecheck gate (heavy; announced to CC before running, per the throttling note).
+
+---
+## Update — render half wired, kit extended to 11/11
+
+Rebased onto `842f5267` (was 15 commits behind at `f287c012`, and did not know it). Insertion point had
+moved — cosmetically. Spec **15/15**, kit **11/11**, typecheck gate **PASSED at baseline 2250 exactly**.
+
+**Gate green proven, not assumed:** phase 1 derives coverage from `git ls-files` and my files were
+UNTRACKED, so a green could have meant "never looked". Injected a deliberate type error → the compiler
+named the file (`optionCoverage.ts(147,14) TS2322`); `--listFilesOnly` confirms all three files in the set.
+
+**The pure module had ZERO importers while it was being built.** Merging it alone would have shipped a
+module with no readers — the precise thing 2.1326 exists to expose. The render half is in the same change.
+
+### M8–M10: do the NEGATIVE copy guards bite?
+A `not.toMatch` passes trivially when the phrase was never going to appear. Each banned phrase is forced
+INTO the rendered copy; all three RED. **M8** "unreliable" (the `ready`-and-uneven contradiction) ·
+**M9** a direction · **M10** dropping the provisional register.
+
+⚠ **M9 v1 WAS A FALSE SURVIVOR OF MY OWN MAKING.** It added a DEAD constant carrying the banned phrase,
+which never reached the rendered copy — so the mutation applied to the FILE and not to the BEHAVIOUR, and
+a working guard read as decoration. **An unapplied-in-effect mutant is worse than no mutant: it accuses a
+healthy guard.** Repaired to inject into the copy itself; it now bites.
+
+⚠ Also caught: a stray double comma from a scripted edit put an **elision** in the mutant array, so the kit
+**crashed after M7b** while printing eight plausible result lines. `node --check` is now run before the kit.
+**Eight healthy-looking lines and a non-zero exit is not a result.**
+
+### Still not done
+No render-level spec on `V5AnalysisResultBlock`; **no mutants on the render half**. Wire-derived, not
+journey-witnessed.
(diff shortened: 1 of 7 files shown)


Talchain:
## Independent adversarial review — **FIX-FIRST** (4 gates, all small)

Reviewed at `ec4b33ff`, fresh blobless clone, `HEAD` asserted before any read. Base `842f5267`. Mutants ran in a detached worktree outside the repo root; isolation proven by **writing** a sentinel (distinct inodes, clone unchanged), restores `HEAD`-relative, `applied=1` per mutant and `applied=0` on both controls.

**The change is a real improvement and I could not refute either defect.** I tried and failed to break it four ways, listed at the bottom. What follows is about the **kit**, one **shipped comment**, and one **latent divergence** — not about the diagnosis.

⚠ My first mutant run produced no output because `$SPECS` was passed unquoted under zsh (no word-split → vitest found no files → exit 0). Everything below was re-run under `bash` with a hard error on zero collected. Spec collection asserted **by name**: `analysisNewRenderDefects.spec.ts` = **6 tests**, `StrengthenTheReasoning.spec.tsx` = **8**.

---

### G1 ⛔ The word-boundary property is unpinned — the fixture cannot discriminate its own mechanism

**Measurement.** Delete the word-boundary logic outright in `src/utils/text.ts` — `const body = lastSpace > keep * 0.6 ? cut.slice(0, lastSpace) : cut` → `const body = cut`, i.e. revert to a hard cut with an ellipsis stapled on:

```
applied=1
Test Files  3 passed (3)
      Tests  40 passed (40)     ← SURVIVOR
```

**Contrast control that fired** — same harness, same run set, break `strengthenWhyLine` instead (`if (false) return signal`):

```
applied=1
Tests  2 failed | 38 passed (40)   ← the harness discriminates
```

So the survival is real, not a blind instrument.

**Mechanism, exactly.** The fixture is 95 chars; `keep = 79`; `TEXT[79]` is a **space**. The hard cut therefore lands on a word boundary *by coincidence*:

| | output | last word whole? |
|---|---|---|
| fixed | `…then the ordering of the two…` | ✅ |
| mutant | `…then the ordering of the two options…` | ✅ |

`expect(TEXT.split(' ')).toContain(lastWord)` passes for both. All three truncation tests survive: test 2 takes the early return, test 3 asserts only `length > 10` and the suffix.

I could not construct **any** revert of the word-boundary logic that reds. The only truncation revert that bites is deleting the ellipsis — which pins the suffix, not the property the PR is named for. **"Both fixes proven to bite by revert" does not hold for the truncation fix.**

**Smallest remedy — delete one word from the fixture.** Drop `two`:

```ts
const TEXT = 'If the adoption rate changes significantly then the ordering of the options changes with it'
```

Measured: mutant now yields `…of the options cha…`, `wholeWord = false` → **RED**. Fixed yields `…of the options…` → **GREEN**. A discriminating pair, one word.

Also: the `describe` is titled *"never cuts a word in half"*, and test 3 exercises the branch where it **does** (`lastWord` is not a whole word for a single long token). That is a deliberate, documented trade-off — but the block name over-claims it. Rename to *"cuts at a word boundary where one is available"*.

---

### G2 ⛔ Neither fix binds at its call site — and the obvious remedy is vacuous

Disclosed as "no render-level spec on the strip". Sharpening it, because the shape matters:

- **M1** — revert `StrengthenTheReasoning.tsx:110-112` to the original concatenation: `applied=1`, **40/40 green**. Survivor.
- **M2** — revert `buildAnalysisNewViewModel.ts:420` to `text.slice(0, 80)`: `applied=1`, **40/40 green**. Survivor.

A render spec **does** exist — `analysisNew/__tests__/StrengthenTheReasoning.spec.tsx:42` asserts `analysis-new-strengthen-why`. Its fixture uses `signal !== whyNow`, so it pins the **join** and is structurally blind to the dedupe. Adding a case there is the right move, **but not in the file's existing style**:

```
UNDER M1:
  ✓ A: toHaveTextContent(BODY)            ← PASSES. jest-dom matches on SUBSTRING.
  × B: occurrence count === 1             ← REDs.
```

`toHaveTextContent` cannot see `"BODY BODY"` as different from `"BODY"`. A case added with the established matcher would be a guard agreeing with itself.

**Smallest remedy** — one `it` in the existing spec, bound by identity, proven GREEN at pristine / RED under M1:

```ts
it('prints the sentence once when signal and whyNow are the same string', () => {
  render(<StrengthenTheReasoning interventions={[rec({ id: 'strengthen:phase3:g1', signal: BODY, whyNow: BODY })]} />)
  const t = screen.getByTestId('analysis-new-strengthen-why').textContent ?? ''
  expect(t.split(BODY).length - 1).toBe(1)
})
```

---

### G3 ⛔ A shipped comment asserts what this PR's own evidence refutes — and it is the justification for truncating

`buildAnalysisNewViewModel.ts:419`:

> `// The full sentence still rides `implication` below.`

The PR body says this was **measured false**: `u.suggestion` is present on every row and is the constant `"Review this assumption"`, so `implication` carries a remedy, never the sentence.

I confirmed it independently **from source** — a complete field manifest for a non-threshold uncertainty with `suggestion` present:

| field | value |
|---|---|
| `headline` | truncated text |
| `implication` | `u.suggestion` (the remedy) |
| `detail` | `undefined` (threshold-only) |
| `inspect` | Severity / E-value / Factor confidence |

**No field on that row carries the full text.** The comment is the stated reason the cut is acceptable, and it is false exactly where the cut happens.

This is worth more than tidiness: the PR makes the truncation **honest** (word boundary + ellipsis) but does not recover the lost reasoning. The user still cannot read the sentence anywhere on the page.

**Smallest remedy:** replace the sentence with what was measured — *"⚠ `implication` carries `u.suggestion` when present (measured: the constant "Review this assumption" on 3/3 `SENSITIVE_ASSUMPTION` rows), so the full sentence is NOT recoverable on the row. The residual loss is rowed separately."* Then row the residual loss. Bound it to the one finding type measured — do not generalise.

---

### G4 ⚠ Exact equality diverges from the reference implementation the PR cites — **latent**

`strengthenWhyLine` uses bare `===`. `StrengthenPanel.tsx:123` — the panel this PR holds up as the correct behaviour — uses `rec.signal.trim() === rec.whyNow.trim()`. Two consumers of one contract, two predicates, diverging on a whitespace-only difference. The semantics differ too: the panel keeps `whyNow` and drops `signal`; this keeps `signal` and drops `whyNow`.

**I could not show this is reachable today** — when `signal` is absent both fields are the same `item.body` reference, and all nine hardcoded pairs in `buildRecommendations` are genuinely distinct sentences. So: latent, not live.

**Smallest remedy:** `whyNow.trim() === signal.trim()`, matching the cited reference. The PR's argument against fuzzy matching is right and untouched by this — `.trim()` is not a semantic judgement.

---

### Refutations I attempted and **failed** — these hold

1. **`rec.signal` undefined → `"undefined …"` rendered.** The old JSX was undefined-safe; a template literal is not. **Refuted:** `Recommendation.signal` is `string` (required) at `strengthenTypes.ts:46` and always coalesced at `buildRecommendations.ts:259`. (`signal?: string` at :128 is the intermediate `StrengthenPhase3Item`.)
2. **Dedupe fix incomplete.** **Refuted:** the only render sites of `rec.whyNow` are `StrengthenPanel:205` (old panel) and `StrengthenTheReasoning:112` (fixed); the rest are `context:` for the Ask-Olumi drawer.
3. **Other mid-word cuts left on the surface.** **Refuted:** remaining `slice(0, N)` in `analysisNew` are array caps and id construction, never display text.
4. **`buildRecommendations` quietly touched / node twins converged.** **Refuted:** the diff is exactly 5 files, none under `strengthen/` or `canvas/nodes/`.

Minor: the twins are **not** verbatim — `OptionNode` appends `'...'`, `DecisionNode` appends `'…'`. They already render differently. Doesn't change the decision to leave them alone; the word "verbatim" is just wrong.

`EVIDENCE.md` is **absent from this branch** (`git ls-files`, contrast control fired: 21 `strengthen` paths). The M1/M9 self-refutations are not in the reviewable artefact — please land them or link them.

---

### CI — `Visual Regression (advisory)` discharged

At `ec4b33ff`: `total_count = 14`, **0 in-progress**, 13 success, 1 failure. The failure is `Visual Regression (advisory)`, which is `completed/failure` on **five consecutive staging commits including this PR's merge-base `842f5267`**. Not introduced here. `mergeable: MERGEABLE` (not DIRTY), so the checks belong to this head.

### What my evidence EXCLUDES

- **Not journey-witnessed and not DOM-witnessed by me.** Everything above is source-derived plus jsdom. I have not driven the deployed surface.
- **I did not run the full suite locally** — CI ran it (4 shards green at this SHA). My runs were 3 spec files / 40 tests.
- **G4's unreachability is a claim about today's producer**, derived from `buildRecommendations` only. A wire `item.signal` differing from `item.body` by whitespace would make it live; I have no wire capture.
- **G3's field manifest is for one finding shape** — non-threshold uncertainty with `suggestion` present. It does not certify other finding types, matching the PR's own bounded claim.
- I did not exercise the `analysisNew` surface under the deployed flag posture; jsdom green says nothing about what the flags mount.

PR #898, opened 2026-08-27
Show what an analysis was based on, beside the result

Talchain:
The **results panel** now states **which option×factor effects the model actually specifies**, beside the ranking it qualifies.

## ⭐ The strip moved surfaces — this is the largest change in the PR

It was first built into the conversational `analysis_result` block. That could not be made honest: coverage derives from **live store state** and a transcript entry is a **historic record**, so every past block showed today's coverage, and the strip vanished silently on the many paths that invalidate readiness. **Two successive gates were built for that and both were wrong** — the second was measured failing on a real capture where the graph hash was identical across two turns while coverage went from every cell set to none.

The answer was not a third gate. **A guard is the wrong instrument when the derivation is in the wrong kind of surface.** On a live results panel, reading live state is not a hazard to be fenced — it is the correct authority, because the panel is *about* the current model. The strip now renders in `ResultsBody`; `computed_against_hash` and all the gating machinery are reverted out.

## Why

On a captured run (CEE `58fdb11`, UI `f287c012`) the product reported:

> *keep what we have came out ahead in **83%** of runs of this model.*

One option had **1 of 3** effects set; the other had **2 of 3**. Two of the empty cells were licence cost and adoption — the substance of the brief. `confidence_tier: "fair"`. **Nothing on screen said any of that.**

Paul's ruling (26 Aug): show the result, show the unequal coverage, use provisional language — *"currently scores higher"*, not *"winner"*. **Never auto-suppress the comparison.**

## This is a plug-in, not an invention

CEE already publishes all of it. `analysis_ready.options[].interventions` and `repair_proposal.unresolved_inputs[]` ride every analyse turn, and **no production file read either one**:

| symbol | production readers |
|---|---|
| `repair_proposal` | **0** |
| `unresolved_inputs` | **0** |
| `confidence_tier` | 7 ← contrast control |
| `win_probabilities` | 6 ← contrast control |

Re-derived at `f287c012` and again at `842f5267`. The controls fire, so the zeros are absence, not blindness.

## ⚠ Prior art, and why this is a different question

`src/components/results/utils/notAnalysedOptions.ts` lives in the same directory, derives from intervention absence, and has shipped copy and live consumers. **Its header explicitly invokes trap 21 against a second authority on its question — so this PR owes an answer.**

It answers **"was this option in the analysis at all?"** — an option with *no* interventions is excluded from the PLoT submission, gets no rank and no probability, and is shown as a proposed alternative.

This answers **"among the options that WERE analysed, which effects were set?"** In the captured run both options were `status: "ready"` with real win probabilities (0.168 / 0.832), so `notAnalysedOptions` says nothing about either — and one of them still had two empty cells.

**Inclusion versus completeness. Two questions, kept apart.**

Also noted, not extended: `ModelSnapshot.tsx:215-219` already renders *"{mapped} of {total} options have mapped effects"* but is **dark** — an optional prop no caller passes. `OptionPanel.tsx:163-166` derives `interventionIds`, the same input. Neither is on this surface.

## Design decisions worth review

- **The denominator is a required parameter with no fallback.** Deriving it from the union of intervened factors would silently hide a factor that *no* option addresses — in the captured run, adoption, which is exactly the case that matters. A denominator that changes with its input is a mirror.
- **⚠ `rankingIsProvisional` HAS BEEN DELETED, and the hedge is adjacency-only.** It was exported to gate leader emphasis and acquired **zero production consumers** — one occurrence repo-wide, its own definition, against contrast controls that fired in the same sweep. Its header called it *"the load-bearing call"* while nothing called it: **a false label on an unreferenced function, inside the module written to stop the product claiming things that are not true.** Wiring it means gating the leader treatment on a live panel, which is a product-surface decision that does not belong in a fabrication fix. So: **nothing gates the ordering, `data-leader`, or the heavier border, and the producer's summary still renders above the strip unchanged.** The provisional register is carried by the words beside the ranking and by nothing else. That is the accepted scope, now stated in the module header. *(The reasoning for never keying it on `robustness.level` — which measures sensitivity to values that were sampled, and an unset effect is never sampled — survives in the header.)*
- **A factor with no honest label is dropped, never printed as its id.**
- **`option.id` was checked, not assumed.** The wire sends `option_id`; `normaliseV5AnalysisReady` maps it, preferring a non-empty `opt.id`. The existing consumer that reads `o.id` is a dark path, so its usage is evidence about the *type*, not the runtime.

## Copy, pinned in both directions

**⚠ CORRECTED — this section described a fabrication that has since been removed.** It said the even case reads *"Every option has all its effects set."* **Two options at 2 of 3 AGREE and neither is COMPLETE**, and that copy shipped on exactly that state.

There are now three states. **Complete** (every option sets every factor): *"Every option has all its effects set."* **Even-but-incomplete**: *"The same effects are unset on every option."* **Uneven**: *"Not every option is equally specified"* + the counts + *"treat the ranking as which option currently scores higher, not as a settled result."*

**Honest at zero is said, not encoded as an absence** — otherwise "no disclosure" and "nothing to disclose" are indistinguishable. **And its mirror: "the counts match" must never be spoken as "nothing is missing."**

Tests assert it **must** carry the provisional register, and **must not**: read as a verdict on whether the run should have happened (a payload can be `ready` *and* unevenly covered — 5 such runs in a 122-run corpus, and contradicting the producer on its own screen is the harm the `may_run` waiver prevents), assert a direction, or assert a cause.

## Verification

- **18 module tests + 5 surface tests**, asserted by name and expected count (the two dark-predicate tests went with the deletion; the surface spec gained the id-guard pin below)
- **11-mutant kit**, all as expected: a **discriminating pair** for identity binding (reverse order → GREEN, different id → RED), and **three mutants forcing each banned phrase into the copy** to prove the negative guards are not vacuous
- Isolation **proven by writing a sentinel**, not by locating a path; restores read a pristine archive outside the repo
- **Typecheck gate passed at baseline 2250**, with a **positive control** confirming the gate loads these files — phase 1 derives coverage from `git ls-files` and these were untracked, so a green could otherwise have meant "never looked"

⚠ **Two of my own instruments failed first and are documented in [`docs/CC2-EVIDENCE-2.1326.md`](docs/CC2-EVIDENCE-2.1326.md):** an assertion that was correct and structurally unable to fail, and a mutant that applied to the file but not to the behaviour. Both were caught and repaired; a reviewer should read that file, not only the diff.

## Review findings closed since first submission

- **The `even` branch claimed completeness on an incomplete model.** `kind` was `Math.min(counts) === Math.max(counts)` — equality over **counts** — selecting *"Every option has all its effects set."* Two options at 2 of 3 agreed and neither was complete, and **the feature's own prescribed remedy produced it**. Now three states: `complete` (every count === denominator) · `even-incomplete` · `uneven`. Verified by compiling the module at base and head and running the fixture matrix against both, with a control that reads `uneven` on both sides so the harness demonstrably discriminates.
- **A raw option id could reach the screen** (*"5f6f5e36 has 1 of 3 set."*). The guard now lives in `deriveOptionCoverage`, not in the caller, because there is more than one caller. An option that cannot be named **aborts the whole reading**.
- **An id-less option was silently dropped** (`continue`). Dropping a participant shrinks the denominator, so an incomplete comparison can read COMPLETE. It now returns `null`, **and that guard is pinned** — a mutant reverting it to `continue` REDs, with a different-object mutant staying GREEN so the pair proves binding by identity rather than sensitivity to any edit.

## Not done, stated precisely

**The upstream drop is NOT closed, and cannot be closed from here — rowed as ROADMAP 2.1335.** `normaliseV5AnalysisReady` (`applyV5State.ts:229`, sole call site `:1228`) drops options carrying neither `id` nor `option_id` **before any consumer sees them**, so on the wire path the denominator can be short with nothing downstream able to detect it: `[A 1of3, B 3of3, C 3of3]` reads **uneven**, and with A dropped `[B 3of3, C 3of3]` reads **complete**. That is this PR's own P0 through a different door. The fix belongs in the normaliser — report what it dropped — not in each consumer. **Latent, not witnessed:** across the 7 in-repo capture fixtures (27 options) 0 are missing an id and 0 are missing a label.

**Wire-derived, not journey-witnessed.** No staging drive.

**The strip has no captured visual state — rowed as ROADMAP 2.1336.** `Visual Regression (advisory)` is green at this head and it genuinely compared (step 9 ran, steps 10-12 skipped, *"Found 10 committed linux reference image(s)"*, 34 passed, completeness guard 10/10 by name). ⛔ **That is not pixel verification of this strip.** The ten captured states are `blocked-provisional`, `fresh-draft`, `graph-default-zoom`, `model-tab`, `olumi-tab`, and **none mounts a results panel with a completed two-option analysis** — the only state that renders the strip. Adding that state belongs in its own commit: a reference blessed in the same PR as the change it is meant to catch proves nothing.

## ⭐ Instrument note: a contrast control from the same naming family can fire while blind

Reviewing this PR I reported *"0 linux visual references committed"* from `grep 'linux\.png$'`, with a contrast control of `darwin.png` = 4 that fired. **The zero was false.** The references live at `e2e/visual/references/linux/*.png` — **the platform is in the DIRECTORY name, not the filename** — and my pattern only knew the Playwright filename-suffix convention. The contrast came from *that same naming family*, so **it proved my pattern worked on the shape I already believed in, while being blind to exactly what I was looking for.**

**The rule: a contrast control must come from a different naming family than the one your pattern assumes.** Otherwise a firing control is evidence about your assumption, not about the world. (Kin to the note above: an occurrence count is not the property either.)

## ⭐ A note for the next reader: the count was not the property

A review swept this body for the fabricated sentence and found *"Every option has all its effects set"* **twice** — which reads as the defect persisting. **Both are legitimate:** one is the correction notice, one is the genuine `complete` copy, which is the only state entitled to say it. **Do not "fix" the correction notice.** An occurrence count can rise while the defect it measures is genuinely removed.

*Independent review required before merge — I do not merge my own work.*

### DIFF
--- a/docs/CC2-EVIDENCE-2.1326.md
+++ b/docs/CC2-EVIDENCE-2.1326.md
@@ -1,0 +1,82 @@
+# 2.1326 build evidence — `optionCoverage`
+
+**Clone:** `/private/tmp/cc2-ui-sa` @ `f287c012` · **26 Aug 2026** · CC2
+
+## Targeted spec
+`npx vitest run src/components/results/utils/__tests__/optionCoverage.spec.ts`
+→ **1 file, 10 tests, 10 passed, exit 0.** Asserted BY NAME and by expected count — never inferred
+from a suite total (a new spec collecting `(0 test)` is invisible to every aggregate).
+Env: `VITE_SUPABASE_URL` / `VITE_SUPABASE_ANON_KEY` set.
+
+## Mutation kit — 8/8 as expected, exit 0
+Isolation **proven by WRITING a sentinel**, not by locating a path: inodes compared, sentinel
+confirmed not to propagate into the source. Restores read a **pristine archive outside the repo**,
+never the mutated tree. Every mutant asserts its anchor is **unique** (an unapplied mutation is
+indistinguishable from an equivalent one) and that it **applied**. Control reads exactly `0`;
+trailing control GREEN.
+
+| id | mutation | expect | got |
+|---|---|---|---|
+| M1 | truthiness instead of key presence | RED | RED |
+| M2 | drop the denominator dedupe | RED | RED |
+| M3 | `kind` always `even` | RED | RED |
+| M4 | accept a single option | RED | RED |
+| M5 | provisional on every reading | RED | RED |
+| M6 | set/unset swapped | RED | RED |
+| **M7a** | **reverse `perOption` ORDER** | **GREEN** | **GREEN** |
+| **M7b** | **assign a different option id** | **RED** | **RED** |
+
+**M7a/M7b are a discriminating PAIR, and neither alone proves anything.** M7a green shows the test
+does not bind by position; M7b red shows it still discriminates. A single biting mutant proves
+sensitivity to *something*; the pair proves sensitivity to *the named object*.
+
+## ⚠ M1 SURVIVED ON THE FIRST RUN, AND THAT IS THE MOST USEFUL THING HERE
+The first fixture wrote each intervention as `{ normalised_value: 0 }`. **The wire sends a BARE
+NUMBER** — `{"6886a726": 0}` — with `intervention_details` as the separate object map (re-read at the
+capture, not guessed). An object is always truthy, so the truthiness mutant survived: the
+*"counts ZERO as SET"* test asserted exactly the right thing and **could not fail**.
+
+**The assertion was right, the fixture made it vacuous, and only the mutant could see that.** The spec
+header had itself quoted *"a fixture you wrote yourself is not evidence about the wire"* — while
+getting the shape wrong. Corrected to the wire shape; M1 now bites. The note is kept in the spec so
+the next reader knows why the shape is spelled that way.
+
+## Process, stated accurately
+**This was NOT written RED-first** — the module came before the spec. Rather than claim an ordering
+that did not happen, the evidence is the mutation kit above: every assertion is shown to fail when
+the behaviour it names is removed, and the identity binding is proven with a discriminating pair.
+
+## Not yet done
+- The render half (`V5AnalysisResultBlock`) and its wiring to the readiness slice.
+- The repo's named typecheck gate (heavy; announced to CC before running, per the throttling note).
+
+---
+## Update — render half wired, kit extended to 11/11
+
+Rebased onto `842f5267` (was 15 commits behind at `f287c012`, and did not know it). Insertion point had
+moved — cosmetically. Spec **15/15**, kit **11/11**, typecheck gate **PASSED at baseline 2250 exactly**.
+
+**Gate green proven, not assumed:** phase 1 derives coverage from `git ls-files` and my files were
+UNTRACKED, so a green could have meant "never looked". Injected a deliberate type error → the compiler
+named the file (`optionCoverage.ts(147,14) TS2322`); `--listFilesOnly` confirms all three files in the set.
+
+**The pure module had ZERO importers while it was being built.** Merging it alone would have shipped a
+module with no readers — the precise thing 2.1326 exists to expose. The render half is in the same change.
+
+### M8–M10: do the NEGATIVE copy guards bite?
+A `not.toMatch` passes trivially when the phrase was never going to appear. Each banned phrase is forced
+INTO the rendered copy; all three RED. **M8** "unreliable" (the `ready`-and-uneven contradiction) ·
+**M9** a direction · **M10** dropping the provisional register.
+
+⚠ **M9 v1 WAS A FALSE SURVIVOR OF MY OWN MAKING.** It added a DEAD constant carrying the banned phrase,
+which never reached the rendered copy — so the mutation applied to the FILE and not to the BEHAVIOUR, and
+a working guard read as decoration. **An unapplied-in-effect mutant is worse than no mutant: it accuses a
+healthy guard.** Repaired to inject into the copy itself; it now bites.
+
+⚠ Also caught: a stray double comma from a scripted edit put an **elision** in the mutant array, so the kit
+**crashed after M7b** while printing eight plausible result lines. `node --check` is now run before the kit.
+**Eight healthy-looking lines and a non-zero exit is not a result.**
+
+### Still not done
+No render-level spec on `V5AnalysisResultBlock`; **no mutants on the render half**. Wire-derived, not
+journey-witnessed.
--- a/src/canvas/conversation/types.ts
+++ b/src/canvas/conversation/types.ts
@@ -219,6 +219,15 @@ export interface V5AnalysisResultBlock {
   leading_option_id: string | null
   win_probabilities?: Record<string, number>
   enrichment?: Record<string, unknown>
+  /**
+   * The graph hash this run computed against, carried from the wire.
+   *
+   * A transcript block is a HISTORIC record, and anything rendered inside it
+   * that reads LIVE store state describes today rather than the run. This lets
+   * a block say which graph it was about, so a live-state derivation can refuse
+   * to speak for a run it does not describe.
+   */
+  computed_against_hash?: string
 }
 
 export interface V5GraphPatchBlock {
--- a/src/canvas/conversation/types.ts
+++ b/src/canvas/conversation/types.ts
@@ -219,15 +219,6 @@ export interface V5AnalysisResultBlock {
   leading_option_id: string | null
   win_probabilities?: Record<string, number>
   enrichment?: Record<string, unknown>
-  /**
-   * The graph hash this run computed against, carried from the wire.
-   *
-   * A transcript block is a HISTORIC record, and anything rendered inside it
-   * that reads LIVE store state describes today rather than the run. This lets
-   * a block say which graph it was about, so a live-state derivation can refuse
-   * to speak for a run it does not describe.
-   */
-  computed_against_hash?: string
 }
 
 export interface V5GraphPatchBlock {
--- a/src/components/results/ResultsBody.tsx
+++ b/src/components/results/ResultsBody.tsx
@@ -9,6 +9,10 @@
  * - Storybook stories (with fixture data)
  */
 
+import { useOptionCoverage } from '../../v5/blocks/useOptionCoverage'
+import { buildCoverageDisclosure } from './utils/optionCoverage'
+import { useCanvasNodeLabels } from '../../v5/blocks/useCanvasLabels'
+import { resolveCanvasLabel } from '../../canvas/domain/canvasLabels'
 import { useRef, useMemo, memo, useState } from 'react'
 import { useCanvasStore } from '../../canvas/store'
 import { typography } from '../../styles/typography'
@@ -123,6 +127,13 @@ export const ResultsBody = memo(function ResultsBody({
 }: ResultsBodyProps) {
   // UI-SEM-065 input: engine blocker/approximate critiques live on
   // graphHealth (ValidationPanel's source), not on confidence.uncertainties.
+  const coverageReading = useOptionCoverage()
+  const coverageLabels = useCanvasNodeLabels()
+  const coverage = useMemo(
+    () => buildCoverageDisclosure(coverageReading, id => resolveCanvasLabel(id, coverageLabels)),
+    [coverageReading, coverageLabels],
+  )
+
   const engineDegradedCritique = useCanvasStore(s =>
     (s.graphHealth?.issues ?? []).some(
       (i: { severity?: string; code?: string }) => i.code === 'GRAPH_TOO_LARGE' || i.severity === 'blocker',
@@ -661,6 +672,30 @@ export const ResultsBody = memo(function ResultsBody({
               isNormalised={resultsSectionData.recommendation.isNormalised}
             />
             {/* TippingPoints removed — superseded by TornadoChart (Brief 5.4 Phase 1) */}
+
+            {/* ROADMAP 2.1326 — what this comparison was actually based on.
+                Adjacent to the ranking it qualifies, on a LIVE panel where
+                reading live store state is the correct authority rather than a
+                hazard to be gated. */}
+            {coverage && (
+              <div
+                className="mt-3"
+                data-testid="results-coverage-disclosure"
+                data-coverage-kind={coverage.kind}
+              >
+                <p className={typography.panelMeta}><strong>{coverage.headline}</strong></p>
+                <p className={`${typography.panelMeta} text-text-light`}>{coverage.detail}</p>
+                {coverage.unsetByOption.length > 0 && (
+                  <ul className="mt-1" data-testid="results-coverage-unset">
+                    {coverage.unsetByOption.map(o => (
+                      <li key={o.label} className={`${typography.panelMeta} text-text-light`}>
+                        {o.label}: no effect set on {o.unsetLabels.join(', ')}
+                      </li>
+                    ))}
+                  </ul>
+                )}
+              </div>
+            )}
           </div>
         </SectionErrorBoundary>
       )}
--- a/src/components/results/__tests__/ResultsBody.coverageDisclosureLiveMount.spec.tsx
+++ b/src/components/results/__tests__/ResultsBody.coverageDisclosureLiveMount.spec.tsx
@@ -1,0 +1,216 @@
+/**
+ * ResultsBody — the coverage disclosure renders ON THE SURFACE, and says the
+ * complete case out loud.
+ *
+ * WHY THIS FILE EXISTS. The disclosure was reviewed with ZERO surface-level
+ * assertions — 0 against same-card siblings at 5, 5, 6 and 7, with the contrast
+ * controls firing. That was tolerable while it sat beside tested siblings and is
+ * not tolerable across a surface move, which would ship it unpinned by
+ * construction. Its honest-at-zero property was pinned only at the MODULE, so
+ * the one behaviour the feature exists for was untested exactly where it renders.
+ *
+ * These bind to the MOUNT PATH, not to a predicate: the assertions run against
+ * `ResultsBody`'s own rendered output, so if the strip stops being mounted here
+ * they fail loudly rather than passing against a component nobody loads.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { render, screen } from '@testing-library/react'
+import { ResultsBody } from '../ResultsBody'
+import type { ResultsSectionDataReturn } from '../useResultsSectionData'
+import type {
+  ConfidenceSectionData,
+  DecisionResultData,
+  DriversSectionData,
+  ImprovementsSectionData,
+  OptionResult,
+} from '../types'
+
+vi.mock('../../../canvas/utils/focusHelpers', () => ({
+  focusNodeById: vi.fn(),
+  focusByTarget: vi.fn(),
+  focusExistingTarget: vi.fn(),
+  focusModelTarget: vi.fn(() => true),
+}))
+
+vi.mock('@/flags', async () => {
+  const actual = await vi.importActual<typeof import('@/flags')>('@/flags')
+  return {
+    ...actual,
+    isFocusNowPanelEnabled: vi.fn(() => true),
+    isStrengthenPanelEnabled: vi.fn(() => false),
+    isAiPanelV2Enabled: vi.fn(() => true),
+  }
+})
+
+import { useCanvasStore } from '@/canvas/store'
+import { useUIStore } from '@/stores/uiStore'
+
+const SWITCHING = 'fac_switching'
+const LICENCE = 'fac_licence'
+const ADOPTION = 'fac_adoption'
+
+const FACTOR_NODES = [
+  { id: SWITCHING, type: 'factor', data: { label: 'One-Off Switching Cost' } },
+  { id: LICENCE, type: 'factor', data: { label: 'CRM Annual Licence Cost' } },
+  { id: ADOPTION, type: 'factor', data: { label: 'CRM Adoption and Usability' } },
+]
+
+function seed(options: Array<Record<string, unknown>>) {
+  useCanvasStore.setState({
+    analysisFreshness: null,
+    analysisFreshnessDirty: false,
+    nodes: FACTOR_NODES,
+    ceeAnalysisReady: { options },
+  } as never)
+}
+
+function makeData(): ResultsSectionDataReturn {
+  const winner = {
+    id: 'opt_a',
+    label: 'Option A',
+    expected: 0.8,
+    outcome: { mean: 0.8, p10: 0.6, p50: 0.78, p90: 0.95 },
+    p10: 0.6,
+    p50: 0.78,
+    p90: 0.95,
+    isRecommended: true,
+    winProbability: 0.7,
+    goalProbability: 0.7,
+  } as unknown as OptionResult
+  const runnerUp = {
+    id: 'opt_b',
+    label: 'Option B',
+    expected: 0.4,
+    outcome: { mean: 0.4, p10: 0.2, p50: 0.38, p90: 0.6 },
+    p10: 0.2,
+    p50: 0.38,
+    p90: 0.6,
+    isRecommended: false,
+    winProbability: 0.3,
+    goalProbability: 0.3,
+  } as unknown as OptionResult
+  const recommendation = {
+    recommendedOption: winner,
+    allOptions: [winner, runnerUp],
+    goalLabel: 'Maximise success',
+    goalThreshold: 0.6,
+    isSingleOption: false,
+    analysisStatus: 'computed',
+    recommendationStability: 0.92,
+    robustnessLevel: 'high',
+    isNormalised: false,
+    coachingReadiness: 'ready',
+    coachingReadinessDimensions: { evidence: 0.8, robustness: 0.75, clarity: 0.85 },
+  } as DecisionResultData
+  const drivers: DriversSectionData = {
+    drivers: [],
+    topDrivers: [],
+    driversStatus: 'computed',
+    totalCount: 0,
+    hasMagnitudeData: false,
+  }
+  const confidence = {
+    tier: { tier: 'strong', icon: 'Check', label: 'Tier', description: 'd' },
+    qualityScore: 80,
+    uncertainties: [],
+    topUncertainties: [],
+    improvements: [],
+    topImprovements: [],
+    evidenceGaps: [],
+    topEvidenceGaps: [],
+    nextActions: [],
+    topNextActions: [],
+  } as unknown as ConfidenceSectionData
+  const improvements: ImprovementsSectionData = {
+    improvements: [],
+    count: 0,
+    hasHighPriority: false,
+  } as ImprovementsSectionData
+  return {
+    recommendation,
+    drivers,
+    confidence,
+    improvements,
+    isLoading: false,
+    isError: false,
+    goalLabel: 'Maximise success',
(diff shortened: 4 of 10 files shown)


Talchain:
# Adversarial review — **FIX-FIRST**

Independent reviewer, no prior involvement. Fresh blobless clone at a unique `/private/tmp` path, `git fetch origin pull/898/head:pr898 && git checkout pr898`, **HEAD asserted `5ca637e5ab71aba1383c009cf46bed4085fc7511`** = the stated head. All searches `/usr/bin/grep` (never `rg`, which is a shell function here), never `|| true`, every absence claim with a contrast control in the same invocation.

**Four gates block; one is a P0-class honesty defect that the PR's own test suite pins as correct.** The engineering underneath is strong — the mint-check obligation is discharged, the CI guards are clean by execution, and the zero-intervention-as-set decision is right and well-evidenced. The defects are all in **predicate breadth**, not in care.

---

## ⛔ GATE 1 (BLOCKING, P0-class) — the `even` branch tells the user the model is COMPLETE when it is not

`kind` is computed as **`Math.min(counts) === Math.max(counts)`** (`optionCoverage.ts:121`) — an **equality predicate over COUNTS**. The copy it selects asserts **COMPLETENESS**:

```
EVEN_HEADLINE = 'Every option has all its effects set'
EVEN_DETAIL   = 'This comparison rests on a complete model.'
```

Count-equality and completeness are two different questions under one name — inside a module whose header invokes trap 21 against exactly that.

### Measurement

I bundled `optionCoverage.ts` with `esbuild` and ran the **compiled module** (not a re-implementation) against six inputs:

| # | input (denominator = 3 factors) | `kind` | rendered headline |
|---|---|---|---|
| **A** | **CONTROL** — the author's captured run, 1of3 vs 2of3 | `uneven` | "Not every option is equally specified" ✅ |
| **B** | the spec's **own** even fixture: both **2 of 3**, adoption unset on **BOTH** | `even` | **"Every option has all its effects set"** ❌ |
| **C** | disjoint singles — each sets a *different* factor, **4 of 6 cells empty** | `even` | **"…rests on a complete model."** ❌ |
| **D** | both options set **ZERO of 3** | `even` | **"Every option has all its effects set"** ❌ |
| **E** | genuinely complete, 3 of 3 both | `even` | correct ✅ |

**Contrast control:** case A returns the uneven copy in the same invocation, so the probe discriminates. **B, C and D are indistinguishable from E at the user surface.**

### Reachability — one prescribed user action away from your own captured run

This needs no exotic payload. In the run this feature was built from, **adoption is unset on BOTH options** and the challenger is missing licence cost. The strip's own call to action is *"no effect set on CRM Annual Licence Cost, CRM Adoption and Usability"*.

**The user sets the licence cost — the single action the strip prescribes — both options reach 2 of 3, `kind` flips to `even`, and the product tells them "This comparison rests on a complete model" while adoption is still empty on both.** The header calls adoption *"exactly the case that matters"*. **The feature's own remedy produces the fabrication.**

### The suite pins it

`optionCoverage.spec.ts:188-200` — *"SAYS the even case rather than rendering nothing"* — asserts `headline === 'Every option has all its effects set'` on **fixture B**, where neither option has all its effects set. 15/15 green. This is trap 13d exactly: the invariant was written against **the failure mode in hand** (unequal counts) rather than against **the spec** (completeness), and the corpus shares the code's blind spot. The header rejects a union denominator *because* it would "silently HIDE a factor that NO option addresses" — and then `kind` throws that away.

### Smallest remedy

**Three states, not two** (the module already argues for naming questions apart):
- `complete` — `perOption.every(o => o.unsetFactorIds.length === 0)` → keep the current even copy
- `uneven` — counts differ → keep the current uneven copy
- `equally_incomplete` — counts match, something unset → its own honest sentence (e.g. *"Every option has 2 of 3 effects set. No effect is set on CRM Adoption and Usability for either option."*), and **`rankingIsProvisional` must be true here too** — a ranking on an incomplete model is provisional whether or not the incompleteness is symmetric.

One-line variant if the three-state version is out of scope for this PR: `kind = perOption.every(o => o.unsetFactorIds.length === 0) ? 'even' : 'uneven'`. B/C/D then render counts + hedge + the unset list, which is honest.

Either way: **re-point the `SAYS the even case` test at a genuinely complete fixture, and add a test pinning the equally-incomplete case** — otherwise the fix lands under a suite that still asserts the old copy.

---

## ⛔ GATE 2 (BLOCKING) — the option axis prints a raw id, re-opening a defect **this file already fixed and documented**

`useOptionCoverage.ts:63`:
```ts
label: typeof raw.label === 'string' ? raw.label : raw.id,
```

`V5AnalysisResultBlock.tsx:206-245` is **`⭐ R-4 — ONE OPTION-LABEL POLICY FOR src/v5/blocks, AND IT REFUSES RAW IDS`**. It removed `optionLabels.get(id) ?? id` from **this very file**, records that *"showing the raw id is honest"* is **false**, and notes the card had been *"a THIRD policy in the same folder, and the weakest of the three"*. **This is a fourth.**

**Measured** (case F, compiled module): detail renders `5f6f5e36 has 1 of 3 set.  has 2 of 3 set.` and `<li>5f6f5e36: no effect set on …`, plus an empty-label row rendering `: no effect set on …`.

It is also **asymmetric with the module's own rule** at `optionCoverage.ts:212-214` — factors are *dropped* rather than printed as ids; options fall through to the id.

**Bounded honestly:** `CEEOptionV3.label` is typed `string` (`adapters/cee/types.ts:343`), but `normaliseV5AnalysisReady` (`applyV5State.ts:237-258`) spreads `...opt` and overrides only `id`/`interventions`/`status` — it never normalises or validates `label`. So this is a **type** guarantee, not a runtime one — the exact class the hook's own header warns about ("a type declaring `reason` while the wire sends `message`"). And `''` is type-legal and passes the `typeof` check **today**.

**Smallest remedy:** resolve option labels through the same null-refusing `resolveCanvasLabel` the card already has in scope at line 322; drop the row when the option cannot be honestly named; keep the id as a `data-option-id` machine reference, exactly as R-4 prescribes.

---

## ⛔ GATE 3 (BLOCKING) — coverage is **present-tense global state** rendered inside a **historic** transcript card

`V5AnalysisResultBlock` is a `ConversationBlock` (`types.ts:216`) rendered **per block** from `InlineBlocks.tsx:681`. Its payload carries only `summary`, `leading_option_id`, `win_probabilities`, `enrichment` — **no options, no interventions** (confirmed: `resolveUncertaintyInputs` shows `enrichment` carries `robustness` and `option_comparison`, and `option_comparison` has no interventions).

`useOptionCoverage` reads `useCanvasStore(s => s.nodes)` and `useCanvasStore(s => s.ceeAnalysisReady)` — **live, global, present tense**.

1. **Every result block in the transcript is annotated with the LATEST coverage.** A card titled *"what an analysis was based on"* is false for every block but the newest.
2. **The disclosure silently vanishes.** `invalidateAnalysisReady` sets `READINESS_CLEAR_FIELDS.ceeAnalysisReady = null` (`store.ts:1410-1419`) on ~22 edit paths — a single unrelated `add_node` (`store.ts:2139`) empties `options`, `deriveOptionCoverage` returns `null`, and the strip **disappears** from a card that carried it a moment ago. **"No disclosure" and "nothing to disclose" become indistinguishable** — which is precisely what the even-branch comment at `optionCoverage.ts:198-199` exists to prevent, one level up.

**Smallest remedy — pick one and say which:** scope the strip to the latest analysis-result block (`InlineBlocks.tsx:689` already threads `isLatestAssistantTurn` to a sibling, so the plumbing exists), **or** snapshot the reading into the block at ingestion. And state the invalidation behaviour explicitly either way.

---

## ⛔ GATE 4 (BLOCKING, disclosure) — `rankingIsProvisional` ships with **zero production consumers**

**Target:** `rankingIsProvisional` appears only in its own module and its own spec.
**Contrast controls, same invocation:** `buildCoverageDisclosure` → 1 production consumer (`V5AnalysisResultBlock.tsx:64,322`); `calibrateUncertaintyCopy` → 3 in the same file. The probe sees production consumers, so **the zero is absence, not blindness.**

The PR body calls it *"the load-bearing call"* and gives it a 17-line header. **It is load-bearing on nothing.** The leader treatment (`data-leader`, the heavier border, first position via `sortedProbs`) is entirely unchanged by coverage, and `block.summary` renders verbatim **above** the strip. The hedge is delivered by **adjacency only** — a sentence below an unhedged producer claim and above an unchanged leader treatment.

In a PR whose stated purpose is plugging in a dark producer field, shipping a new dark export is chronic failure #1.

**Smallest remedy:** wire it (gate the leader emphasis or prefix the summary), **or** delete it and say in the PR body that the hedge is adjacency-only and that is the accepted scope. Do not leave a headline design decision unreachable. *(Your reasoning for not keying it on `robustness.level` is correct and should survive either choice.)*

---

## ⚠ GATE 5 (evidence) — the strip has **zero render-level coverage**, and the pixel instrument is dead

**Target:** `v5-analysis-result-coverage` → **0** test references (source only).
**Contrast:** `v5-analysis-result-uncertainty-copy` → 2 spec files. The probe sees test references; the zero is absence.

So the honest-at-zero path, `data-coverage-kind`, the `<ul>`, and every line of copy are **unasserted at the render**. Your disclosure is accurate and I am not reading 11/11 as covering the strip — but note that **Gates 1 and 2 are both defects a single render-level spec would have surfaced.**

⛔ **And I make no pixel-level claim in either direction.** `Visual Regression (advisory)` is `failure` at this head **and identically `failure` at base `842f5267`** — standing-red for unrelated reasons, not caused by this PR, not required. But this PR **adds a visible element to a rendered card**, so a visual diff *should* have changed, and the one instrument that could say whether it changed correctly has stopped carrying information. **This PR's rendered output is UNVERIFIED at the pixel level, and nothing in this verdict rests on pixel-level safety.** I am not claiming a visual regression and I cannot rule one out.

**Smallest remedy:** one render-level spec asserting, by testid — the strip renders for an uneven store state; it renders the complete-model sentence rather than vanishing when nothing is missing; and no rendered string is a raw id.

---

## ✅ WHAT PASSES

**⭐ MINT-CHECK (Priority 1) — DISCHARGED.** Verified at the bytes: `notAnalysedOptions` answers *"was this option in the analysis at all?"* from `option_probabilities` presence behind `runAnalysedAnyOption`; this module answers *"which effects were set?"* from `interventions` keys against canvas factor nodes. **Different inputs, different questions.** The PR body states the distinction explicitly and evidences it from the captured run (both options `ready` with real win probabilities, so `notAnalysedOptions` is silent about either while one had two empty cells). This is a genuine second question, not a second authority, and the obligation is answered. Two additions the body does not name: **`graphDisplayCalculations.ts:424-436`** builds `directlyControlled` from the same `interventions` keys — but as a **union** ("is this factor controllable at all"), not per-option, so not a second authority; worth naming for completeness. `ModelSnapshot.tsx:215-219` is correctly identified as dark.

**⭐ Priority 2 CI guards — discharged BY EXECUTION, not inspection.**
- Raw typography: both new files measure **0**; `V5AnalysisResultBlock.tsx` holds at **4** (four pre-existing `font-medium`). `text-text-light` is not matched by `RAW_TYPO`. Contrast: `src/styles/typography.ts` = 38.
- `shell-conformance.spec.ts` + `panel-list-conformance.spec.ts`: **56/56 PASS**. The `<ul>` takes `PANEL_LIST_STACK` with no stray marker/indent/rhythm, and STACK is the correct category per `panelLists.ts` (rows carry their own label).

**Zero-as-set is right, and the mutant story that caught the fixture shape is the best thing in this PR.** The `hasOwnProperty` decision, the real-capture fixture, the dedupe of a repeated denominator id, and the refusal to derive the denominator from a union are all correct and correctly argued.

**No regression:** `src/v5/blocks/__tests__` + `InlineBlocks.narrativeChannel` → **21 files, 417/417 PASS** at this head.

**Merge state:** `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` (the advisory only). Check-runs at the **exact head SHA**, filtered `status == "completed"` first: **total 14 · running 0 · success 13 · failed 1** (the advisory). Base `842f5267` reads **identically**. Staging has moved to `56059e18` (`analyticalNodeFields.ts`, `results/constants.ts`) — **no file overlap** with this PR.

---

## Minor

- **The PR body tells reviewers to read `EVIDENCE.md`** — that file is not in the repo at this SHA (`git ls-files | grep -i EVIDENCE` returns only unrelated `docs/evidence/**` and workflow files). Commit it or drop the instruction; a reviewer cannot read it.
- `key={o.label}` on the `<li>` keys React by label. `unsetByOption` drops `optionId` at the disclosure boundary, so identity binding — which the module is careful about everywhere else — is lost exactly where the render needs it. Carry `optionId` through and key on it.

---

## What this evidence EXCLUDES

- **Not journey-witnessed. Not driven on staging.** No wire capture of my own; I inherited yours.
- **No pixel-level verification of any kind** (Gate 5).
- **I did not run the full suite.** I ran: the new spec **by name** (`Test Files 1 passed (1)`, `Tests 15 passed (15)` — your count is accurate), both named CI guards (56/56), and the 21 files under `src/v5/blocks/__tests__` plus `InlineBlocks.narrativeChannel` (417/417). `VITE_SUPABASE_URL`/`VITE_SUPABASE_ANON_KEY` exported and asserted non-zero (lengths 40/208); no collect failures in any run.
- **I did not sample a corpus of real `analysis_ready` payloads**, so I cannot give a **rate** for the Gate 1 fabrication — only that it is structurally reachable and one prescribed user action away from your own captured run.
- I did not audit whether `notAnalysedOptions`' rendered surfaces and this strip can contradict each other on one screen.

**Gates 1–4 fix-first. Gate 5 strongly recommended in the same PR — it is the instrument that would have caught 1 and 2.**

Talchain:
## FIX-FIRST — focused re-review, interval `5ca637e5..a3971bf0`

**Interval measured:** `5ca637e5..a3971bf0` — one commit, "Close review gates: say complete only when it is complete". Not base→head.
**HEAD asserted `a3971bf02460beaed6fb152244d7289c35c260c0`** at checkout and again immediately before posting; `gh api` and `git ls-remote` agree. **The head has NOT moved** — the ruled move of the strip to the live panel, and the revert of the block-gating machinery, are **not in this tree**. Everything below is measured against `a3971bf0`.

Gate 3 not reviewed — withdrawn by the author, ruled out by CC. Two incidental facts from my sweep, offered as data and not as a review of it: `computed_against_hash` **is** contracted (`blocks.d.ts:36`, `z.ZodOptional<z.ZodString>` on a `"strict"` `AnalysisResultBlockSchema`), and `current_graph_hash` is **not in `@talchain/schemas` 0.48.0 at all** — 0 hits against contrast controls `win_probabilities` 57 and `leading_option_id` 93 in the same sweep.

---

### ✅ GATE 1 — CLOSED. Verified by execution, with a discriminating base/head pair.

I compiled `optionCoverage.ts` at **both** `5ca637e5` and `a3971bf0` with esbuild and ran the fixtures against each. Same harness, same inputs, two modules.

| fixture | base `5ca637e5` | head `a3971bf0` |
|---|---|---|
| **A** 1of3 vs 2of3 *(control)* | `uneven` | `uneven` |
| **B** both 2of3, same factor unset | `even` → **claims complete** | `even-incomplete` → does not |
| **C** disjoint singles, 4 of 6 empty | `even` → **claims complete** | `even-incomplete` → does not |
| **D** both 0 of 3 | `even` → **claims complete** | `even-incomplete` → does not |
| **E** genuinely 3of3 | `even` → claims complete | `complete` → claims complete |
| **F** remedy path from A | `even` → **claims complete** | `even-incomplete` → does not |

**The remedy path is walked and closed.** From A, setting the missing licence value so both options reach 2 of 3 no longer flips the copy: head reads `even-incomplete`, headline *"The same effects are unset on every option"*, and it still **names** the missing factor on both options. Control A reads `uneven` on both sides; E is the only state that may claim completeness and does.

`rankingIsProvisional` moved from `kind === 'uneven'` to `kind !== 'complete'`, so B/C/D/F now correctly stay provisional.

**No surviving test pins a false claim.** The only two assertions of the completeness copy (`optionCoverage.spec.ts:234-235`) are bound to the genuinely-3of3 fixture. The spec that previously asserted the false headline on B now asserts its negation, including `expect(d.detail).not.toMatch(/complete model/i)`.

I also checked the empty-input corners: `deriveOptionCoverage` guards `options.length < 2` and `modelFactorIds.length === 0` before the `every()`, so the vacuous-truth path (`[].every() === true` → `complete`) is unreachable.

---

### ⚠ GATE 2 — the leak is closed; the reasoning holds; **one door is still open.**

**The leak is closed on the live path.** The `label: raw.id` fallback is gone.

**The literal answer to the question asked: no.** An unnameable option does **not** still count toward the denominator or the kind — nothing does. `useOptionCoverage.ts:96` `return null` aborts the entire reading, so the whole strip goes silent. That is fail-closed and honest-by-silence (the user sees no strip, not a false one), and it does avoid the harm the author names.

**But the harm it fail-closes against is already open one line above, and I reproduced it.** The two guards are asymmetric:

```
useOptionCoverage.ts:87   if (typeof raw.id !== 'string' || raw.id.length === 0) continue      // SILENT DROP
useOptionCoverage.ts:96   if (typeof raw.label !== 'string' || raw.label.length === 0) return null  // fail-closed
```

The label guard's own comment gives the reason for not dropping: *"a comparison that silently omits a participant can turn uneven into even."* That is exactly what the `id` guard does — and `normaliseV5AnalysisReady` (`applyV5State.ts:239-259`) already drops any option lacking both `id` and `option_id` **before the hook ever sees it**, so the derivation cannot know a participant is missing.

Executed against the head module:

```
[A 1of3, B 3of3, C 3of3]  -> uneven          "Not every option is equally specified"
[B 3of3, C 3of3]          -> complete        "This comparison rests on a complete model."
```

**That is Gate 1's fabrication, reopened through the id door.** A milder form flips `uneven` → `even-incomplete`, where *"The same effects are unset on every option"* is simply false.

**Reachability, stated honestly:** across the 7 live-capture fixtures in this repo — 27 options — **0 are missing a label and 0 are missing an id** (options carry `option_id` + `label`; the normaliser maps the former to `id`). So this is **latent, not witnessed**. I did not measure staging live, and the capture that reportedly produced `"5f6f5e36 has 1 of 3 set."` is not among these fixtures. It still costs one line to close, and it is the exact defect class this PR exists to close.

---

### ⛔ GATE A — the strip has **zero** surface-level tests, and the move is the change no existing test can observe.

Sibling testids on the **same card**, asserted in specs:

```
v5-analysis-result-uncertainty-copy      7
v5-analysis-result-probabilities         6
v5-analysis-result-summary               5
v5-analysis-result-decision-review       5
v5-analysis-result-coverage              0     <-- the strip
data-coverage-kind                       0
```

The contrast controls fire, so the zero is absence, not blindness. **Credit where due: the body discloses this** (*"No render-level spec and no mutants on the strip itself… Please do not read it as covering the render"*). That was tolerable while the strip sat among already-tested siblings. **It is not tolerable across a surface move**, because the module suite stays green whether the new panel renders the strip, renders it wrongly, or renders nothing at all.

**On CC's question — is honest-at-zero pinned on the new surface?** It cannot be, and it is not pinned on the old one either. The completeness copy appears in exactly two files repo-wide: `optionCoverage.ts` and its own unit spec. **Every scrap of evidence for this feature is at the pure-module level.** So the move would ship honest-at-zero unpinned by construction — and *"no disclosure"* versus *"nothing to disclose"* is precisely the distinction that would go unobserved.

**Required with the move:** a spec on the new panel asserting, bound by testid and not by a value predicate, that (i) the strip mounts, (ii) the `complete` case **renders** the honest-at-zero copy — assert the element present with `data-coverage-kind="complete"`, not merely that the module returns it, and (iii) `even-incomplete` does not carry the completeness copy. Prove it bites by deleting the render and watching it RED.

I confirmed CC's placement facts independently: `ResultsBody` and `AnalysisNewTabBody` are both imported and mounted in `OutputsDock.tsx` (`:3372`, `:3469`) — live panels, not transcript entries.

---

### ⛔ GATE B — the option-label guard must travel with the module, not with the caller.

The guard added in this interval lives in the **hook**. The module is unguarded, and I executed it:

```
head module, given an id-shaped option label -> "5f6f5e36 has 1 of 3 set."   LEAKS: YES
head module, given an unnameable FACTOR      -> factor dropped               LEAKS: no
```

Factor labels are guarded **inside** `buildCoverageDisclosure`; option labels are guarded only in `useOptionCoverage`. Today that is safe, because the hook is the module's only production caller. **The move creates a second caller.** If the new panel wires up `buildCoverageDisclosure` without replicating the hook's guard, the leak returns with no red anywhere — and per Gate A there is no surface test to catch it.

---

### ⛔ GATE C — the PR body still over-claims, in three places, all stale against the head.

- **The "Copy, pinned in both directions" section still states the fabrication G1 removed.** Body: *`Even: "Every option has all its effects set."`* At head, `even-incomplete` must **not** say that. The body presents the corrected design's predecessor as the design.
- **"15 tests"** — the head spec has **19** (`5ca637e5` had 15).
- **"provisional stands while coverage is uneven"** — the shipped predicate is `kind !== 'complete'`, so it also holds for `even-incomplete`. And the bullet still reads as a description of live behaviour, when `rankingIsProvisional` has **zero production consumers**: 1 definition + 6 test references, against contrast controls `buildCoverageDisclosure` and `deriveOptionCoverage`, which each do have one.

**G4 is otherwise discharged:** *"load-bearing"* is struck — 0 occurrences in the body, contrast control `coverage` = 3 in the same sweep.

---

### The merge question, plainly

**Yes — shipping G1+G2 on the transcript surface without any G3 fix would introduce a NEW falsehood, and "pre-existing" is not available.** The strip is new in this PR. G1 corrected the **predicate**; it did not touch **which data the predicate reads**. On a historic block the strip still derives from live store state, and CC's own capture `20260812T202414Z` shows the options slice going fully-set → nothing-set with the graph hash unchanged, so a T1 block would report T2's coverage — telling the user a run that had every cell set had none.

**The ruled direction dissolves this rather than routing around it, and I agree with it on the evidence I hold:** a live panel is *about* the current model, so deriving from live store state there is correct and needs no gate. But I have **not reviewed that code — it does not exist at this SHA.** Gates A and B are the conditions I would want on it.

---

### Checks and non-claims

- **`Visual Regression (advisory)`: DISCHARGED, no information.** `failure` at head `a3971bf0` **and** `failure` at base `842f5267` — identical delta. **I make no pixel-level claim in either direction.** The DOM contract I can support instead: the strip renders `data-testid="v5-analysis-result-coverage"` carrying `data-coverage-kind` ∈ {`complete`, `even-incomplete`, `uneven`}, plus `v5-analysis-result-coverage-unset` when anything is missing.
- **Typography pin intact.** The interval's only change to `V5AnalysisResultBlock.tsx` is one line (the hook call); the strip's copy goes through `typography.panelMeta`. Zero raw typography tokens added, so the bidirectional 4-token pin is untouched. The new modules add none.
- **CI at the exact head SHA:** 12 checks, **9 completed, 3 `Full Test Suite` shards still `in_progress`**. The only completed non-success is the advisory visual check. **I make no verdict on the suite while shards are running.** `mergeable_state: blocked`.

### What my evidence EXCLUDES

- **I did not run the UI suite** (instructed not to; a truncated local run is could-not-measure). No claim about test outcomes at head.
- **I executed the pure module only** — `optionCoverage.ts`, esbuild-compiled from head and base. **I executed no React hook and no component.** There is no render evidence, no browser, no journey witness anywhere in this review.
- **The new live-panel placement is not in this tree.** Gates A and B are conditions on work I have not read, derived from the surrounding code, not a review of it.
- Guard reachability is measured against **7 in-repo capture fixtures (27 options)** only — not staging live.
- Gate 3 not reviewed.

PR #899, opened 2026-08-27
Hero badge states RANK, not the ordinal frozen on the first run

Talchain:
## The mechanism

After a re-run that flips the leader, the hero option badge showed the ordinal **frozen at the first run**, while everything around it showed current rank.

```
Producer   src/canvas/store/stableOptionNumbers.ts:14   append-only, ordinals NEVER recycled
Registered store.ts:4113  <- useResultsSectionData.ts:3841-3849, seeded via sortOptionsForDisplay (run 1's ranking)
Built      buildHeroModel.ts:403-404   index = i+1 (RE-RANKS every run) · stableNumber = frozen identity
Rendered   HeroOptionRow.tsx:243       {row.stableNumber ?? row.index}   <- stableNumber WON whenever all rows were registered
```

Consequence, reproduced deterministically: **the new leader's badge is filled leader-blue and carries the numeral 2.** The fill and the numeral disagree inside one 24×24 element.

Reachable only in **one page session with no reload** — `optionNumbering` has no `persist()`, so a reload clears it. That is why runs without a flip agreed.

## Three authorities, one visual element

| element | question it answered | source |
|---|---|---|
| badge numeral | *who led on the **first** run?* | `stableNumber` |
| list order + "Highest on this view" | *who leads **now**?* | `row.index` |
| badge fill + "Leading option" chip | *who leads **now**?* | `isLeader` |

The badge is a **rank affordance** — it is filled for the leader, and `showOrdinal`'s own doc says it "ranks ALL of them". It was being fed identity. **The false premise was written in the render file** (`HeroOptionRow.tsx:45-46`): *"on a first run both are seeded from the probability order — so '1' IS `rank == 1`."* True of a first run, false of every flipping re-run — written from the case in hand rather than from the spec. It is corrected in place, with the reason visible.

## The fix, and the opposite-direction harm

The badge now renders `row.index`. **The one-liner alone would trade a wrong badge for a cross-surface disagreement** — the same option reading "1" in the cockpit and "Option 2" on the card below it and on its canvas node. Two harms cannot share one window, so they were given **two elements**:

| element | question it answers | source |
|---|---|---|
| hero badge (filled for leader) | who leads **now**? | `row.index` |
| hero `Option N` **text** | which option **is** this? | `stableNumber` |
| card `Option N` text | which option **is** this? | `stableNumber` |
| card colour swatch (no numeral) | who leads **now**? | `rank` |

**No third convention was invented.** `OptionCards.tsx:722-738` has drawn rank as a numeral-free colour swatch and identity as the text `Option {stableNumber}` since D17, and `canvas/nodes/OptionNode.tsx` mirrors the same chip. The hero converged on it.

One derived detail: the identity chip is gated **only** on `stableNumber != null`, never on `showOrdinal`. Withholding suppresses **designations**, and identity is not one — OptionCards gates its rank swatch on `!neutralised` and its identity chip on nothing but the number. The hero now matches that precedent exactly.

`analysisDisplaySelector.ts:50`'s `displayIndex` models the same concept correctly and has no live call site. It was **not** consumed: `buildHeroModel` already computes `i + 1` over the same `sortOptionsForDisplay` output, so `row.index` **is** `displayIndex`. Nothing was minted.

## Which specs changed, and why

**One** existing spec pinned the behaviour that intentionally changed: `analysis-hero/__tests__/content.spec.tsx > Wave 2: stable number badges`. Its two cases became **five**:

- `row badges show the identity-anchored ordinal, surviving a rank flip` — was a correct description of the code and a correct pin of the **defect**. Split into a BADGE case (states rank) and an IDENTITY case (still asserts the number survives the flip — on the element that now carries it).
- `without a numbering map the badge falls back to the display rank` — renamed; there is no fallback any more.
- **New**: rank and identity are on separate elements; and an absent number is **never fabricated** (no chip when `stableNumber` is null) — an honest-absence assertion the block never had, with a positive control.

**Nothing was weakened to go green.** Net +3 assertions in that block, and every one of them bites (see M1–M4 below).

### ⚠ A correction to the brief's premise

The brief named `selectors/__tests__/analysisDisplaySelector.spec.ts:79-88` and `:90-110` as specs that "must change". **They must not, and they did not.** They pin the *selector*, which emits `displayIndex` and `stableNumber` correctly and is untouched by this change. Far from contradicting the fix, `:79-88` (`displayIndex 1` with `stableNumber 2`) is the **selector-level statement of exactly the distinction this PR renders**. Changing them would have destroyed the evidence the fix rests on.

Symbol sweep at my tip (`/usr/bin/grep -rl -a`, scope `src/`, `*.ts`+`*.tsx`, contrast control `displayIndex` → 2 files, non-zero): **14** files touch `stableNumber`, not the 17 the brief estimated. Of those, one spec REDed and is the one rewritten above; the whole `src/components/results` suite (274 files / 4705 tests) is green.

## Evidence

**RED-first at pristine `56059e18`**, by name, 11 tests collected across two new specs, 9 RED (the 2 green are deliberate standing guards: the mount-path precondition, and "the card carries no rank numeral"). Both specs drive the **real hook and the real PLoT V2 mapper** across two runs with **only `win_probability` moved**.

**An instrument defect caught at pristine and disclosed:** my first fill/numeral assertion said "the filled badge reads 1" and **passed at pristine** — the default lens is Goal fit, whose leader is `opt_a` on **row 2**, and `opt_a`'s frozen ordinal happens to be 1. A test written for row 1 passed on row 2 with its positive control firing happily (trap 19). Rewritten to derive the filled row's *position* and assert the numeral equals it; it then REDs.

**Mutation battery** — throwaway worktree outside the repo root, `--detach`, **isolation proven by writing a sentinel** (distinct inodes; source SHA-256 unchanged), restores from a pristine archive outside both trees, applied-check scoped to `src/`:

| mutant | applied | result |
|---|---|---|
| M0 control (cannot match) | 0 | 74/74 green |
| M1 defect on **all** rows | 1 | 9 RED |
| M2 defect on every row **except display rank 1** | 1 | 4 RED |
| M3 identity chip deleted | 1 | 7 RED |
| M4 identity chip **fed rank** (the "harmonisation") | 1 | 7 RED |
| M5 trailing control | 0 | 74/74 green |

**The discriminating pair.** `the badge on the new leader row carries its CURRENT RANK` is **RED under M1** and **GREEN under M2**, while `every badge in the list reads its display rank` REDs under both. A single biting mutant proves sensitivity to *something*; the pair proves this assertion binds to **row 1 by identity**.

**The cross-surface pin** (`ResultsBody.crossSurfaceOptionNumbering.spec.tsx`) is REDed by M3 and M4 from **opposite directions** — chip removed, and chip relabelled with rank — which is why its central assertion is an **inequality between two named quantities** rather than two equalities a single well-meaning edit could satisfy at once. It renders `ResultsBody`, the only production parent of `OptionCards` and also the host of `AnalysisHeroContainer`, so both halves of the contradiction are mounted on one screen (CLAUDE.md trap 3b — a green component suite is not evidence about a screen).

**Gates at my tip, all foreground to conclusion:** `pnpm typecheck` PASSED (3736/3776 files, ratchet 2250/2250) · `pnpm lint` PASSED (0 errors; hooks ratchet 229 = baseline) · `src/components/results` 274 files / **4705 tests** all pass · `src/components/results/analysis-hero` 38 files / **616 tests** all pass. Supabase env exported for every run, asserted non-empty, and every collected count checked **by name**.

## What this evidence EXCLUDES

- **jsdom proves presence and text, never layout or visibility** (trap 3). Nothing here says the chip is legible, unclipped, or well-placed at any width. The label's two-line clamp measurement is deliberately untouched — `labelId`/`labelRef` stay on the label span alone — but that is a structural argument, not a measurement.
- **No live witness.** This is CODE EXISTS → TESTED. Not deployed, not mounted-on-staging, not wire- or journey-witnessed. The staging witness is the flip repro: one session, no reload, run then re-run.
- **The canvas `OptionNode` was not rendered.** It reads the same `optionNumbering` map from the store and is unchanged, but no assertion in this PR mounts it.
- **The identity chip's effect on the row button's accessible name is asserted only as text presence** — no screen-reader verification.
- **`DecisionRecordModal.tsx:365`** renders a third spelling of identity, `"2. Label"`, in its chosen-option select. It is identity-framed and carries no rank affordance beside it, so it does not reproduce this defect. **Deliberately not touched** — flagged, not fixed.

## ⛔ Step 0 (asked before building): does the frozen ordinal reach the DURABLE decision record?

**No.** Settled at the bytes, three ways:

1. **The durable POST body carries no ordinal at all.** `decisionRecordCommitService.ts:127-138` sends exactly `scenario_id · chosen_option_id · chosen_option_label · confidence_0_100 · expectation_statement · [revisit_trigger_or_date] · client_commit_id`. Durable identity is the node id plus a label snapshot — and `chosen_option_label` is the **raw** label, not the ordinal-prefixed select string.
2. **`DecisionRecord.optionNumber` is `sessionStorage`-only with ZERO product readers.** Written once (`DecisionRecordModal.tsx:257`), persisted under `decisionRecord.v1` in `sessionStorage`. Complete reader manifest of the field (scope: all of `src/`, `*.ts`+`*.tsx`, `grep -a`): the type declaration, that one write, and two test references. Contrast control `optionLabel` in the same sweep returns 38 files — the instrument sees.
3. **`optionNumbering` is in no persisted or serialised payload.** No `persist()`; `store.ts:6138` explicitly resets it to `{}` on scenario hydration. Sweep of non-test `src/` intersected with `persist|serial|payload|graph_state|save|snapshot|toJSON` → zero hits.

**Severity is therefore unchanged: a display defect, in-session — not a false durable record. Scope was not changed.**

## Collision

Measured against the two live PRs on `src/components/results/` before writing: **#897** (`analysisNew/`, `src/utils/text.ts`) and **#898** (`results/utils/optionCoverage`, `v5/blocks/`). **Intersection with both is empty.** `buildRecommendations.ts` untouched. This PR changes 1 source file and 3 spec files.

---

## CI at head `2dbf906d` (polled to conclusion in-session, queried at the exact SHA)

`mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE`, `head=2dbf906d…` — matching `git ls-remote` and `gh api` independently.

**13 of 14 checks green**, including all four `Full Test Suite` shards, `Full Test Suite Summary`, `TypeScript + Lint`, `Production Build`, `Security Audit (pnpm)`, `Typecheck Gate Self-Test` and — the **only required context** on `staging`, derived via `gh api …/branches/staging/protection` — **`Staging Gate`: success**.

### The one red: `Visual Regression (advisory)` — and what it actually says

I read the job output rather than a label (CLAUDE.md trap 7b). It is not a diff:

> `No linux visual references were committed, so this run compared NOTHING.`

**Contrast control:** the same check is **already failing on pristine `staging` `56059e18`**, where every other check is green. So its red carries no information about this branch.

⚠ **But that cuts both ways, and it belongs in the exclusions:** the estate's visual instrument is in a COULD-NOT-MEASURE state, so **this PR has had no visual verification at all** — and it *does* change hero row layout by adding an element into the label column. jsdom proved the chip's presence and text; nothing has looked at it. Committing linux references is a separate, reviewable piece of work (the job says so itself) and is deliberately not bundled here.

### DIFF
--- a/src/components/results/__tests__/ResultsBody.crossSurfaceOptionNumbering.spec.tsx
+++ b/src/components/results/__tests__/ResultsBody.crossSurfaceOptionNumbering.spec.tsx
@@ -1,0 +1,253 @@
+/**
+ * CROSS-SURFACE OPTION NUMBERING — hero RANK beside card IDENTITY, on ONE
+ * screen, after a re-run that flips the leader.
+ *
+ * ## THIS IS THE LOAD-BEARING HALF OF THE BADGE FIX, NOT A SIDE-ASSERTION
+ *
+ * The hero badge used to render `stableNumber ?? index` — the ordinal frozen
+ * on the FIRST run — while the list order, the leader fill and the "Highest
+ * on this view" cue all stated the CURRENT rank. Pointing the badge at
+ * `row.index` closes that. But it opens the opposite-direction harm
+ * immediately, and this estate's signature defect is exactly that trade: the
+ * same option would then read "1" in the cockpit and "Option 2" on the card
+ * below it and on its canvas node — a cross-surface disagreement bought with
+ * a fixed badge.
+ *
+ * Two harms cannot share one window, so they were given two elements:
+ *
+ *   | element                          | question it answers        | source        |
+ *   |----------------------------------|----------------------------|---------------|
+ *   | hero badge (filled for leader)   | who leads NOW?             | `row.index`   |
+ *   | hero `Option N` text             | which option IS this?      | `stableNumber`|
+ *   | card `Option N` text             | which option IS this?      | `stableNumber`|
+ *   | card colour swatch (no numeral)  | who leads NOW?             | `rank`        |
+ *
+ * The hero did not invent a convention: `OptionCards.tsx:722-738` has drawn
+ * rank as a NUMERAL-FREE colour swatch and identity as the TEXT
+ * `Option {stableNumber}` since D17, and `canvas/nodes/OptionNode.tsx` mirrors
+ * the same chip. The hero converged on it.
+ *
+ * ## WHY THIS FILE, RATHER THAN MORE CASES IN THE HERO SUITE
+ *
+ * A hero-only suite cannot see a cross-surface disagreement — both halves of
+ * the contradiction have to be mounted at once. `ResultsBody` is the only
+ * production parent of `OptionCards` and it also mounts
+ * `AnalysisHeroContainer` from the SAME `recommendation.allOptions`, so this
+ * is the surface a user loads with both numbers on it. (Same reasoning as
+ * `ResultsBody.notAnalysedMountPath.spec.tsx`, and the same trap it cites:
+ * CLAUDE.md 3b — a green component suite is not evidence about a screen.)
+ *
+ * ## WHAT IT IS WRITTEN TO CATCH
+ *
+ * A future "harmonisation" that makes the two numbers agree — from EITHER
+ * side. Pointing the badge back at `stableNumber` REDs it, and so does
+ * relabelling the card's identity chip with the rank. That is why the central
+ * assertion is an INEQUALITY between two NAMED quantities rather than two
+ * separate equalities that a single well-meaning edit could satisfy at once.
+ *
+ * ## STATE CLASS
+ *
+ * SEEDED-THEN-RERUN, one page session, no reload. `optionNumbering` has no
+ * `persist()`, so a reload clears it and the divergence disappears — which is
+ * why runs without a flip, and every reload, agreed.
+ *
+ * ## WHAT THIS FILE EXCLUDES
+ *
+ * jsdom: presence and text, never layout or visibility (CLAUDE.md trap 3). It
+ * says nothing about the canvas `OptionNode`, which reads the same
+ * `optionNumbering` map from the store but mounts on a different surface.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { render, screen, cleanup, within, renderHook } from '@testing-library/react'
+import { ResultsBody } from '../ResultsBody'
+import { useResultsSectionData } from '../useResultsSectionData'
+import { useCanvasStore } from '../../../canvas/store'
+import { mapV2ResponseToReportV1 } from '../../../adapters/plot/v2/responseMapper'
+import type { V2RunResponse } from '../../../adapters/plot/v2/types'
+
+/** First-appearance order A, B, C — the order run 1 freezes ordinals in. */
+const OPTIONS = [
+  { id: 'opt_a', label: 'Hire two developers', mean: 30 },
+  { id: 'opt_b', label: 'Partner with a consultancy', mean: 24 },
+  { id: 'opt_c', label: 'Continue solo', mean: 12 },
+]
+
+/** ONLY `win_probability` moves between the runs. */
+const RUN1_WIN: Record<string, number> = { opt_a: 0.7, opt_b: 0.2, opt_c: 0.1 }
+const RUN2_WIN: Record<string, number> = { opt_a: 0.2, opt_b: 0.7, opt_c: 0.1 }
+
+function makeV2Response(win: Record<string, number>, leaderId: string): V2RunResponse {
+  return {
+    analysis_status: 'computed',
+    option_comparison_status: 'computed',
+    robustness_status: 'computed',
+    drivers_status: 'computed',
+    option_comparison: OPTIONS.map((o) => ({
+      option_id: o.id,
+      option_label: o.label,
+      confidence_interval: [o.mean - 10, o.mean + 10],
+      win_probability: win[o.id],
+      outcome: {
+        mean: o.mean, std: 5, p10: o.mean - 10, p50: o.mean, p90: o.mean + 10,
+        n_samples: 1000, n_valid_samples: 1000, validity_ratio: 1,
+      },
+    })),
+    critiques: [],
+    drivers: [{ node_id: 'd', label: 'D', contribution: 0.5, direction: 'positive' }],
+    edge_sensitivity: [],
+    factor_sensitivity: [{ factor_id: 'f1', elasticity: 0.4, importance_rank: 1 }],
+    // A PERMITTED run: without a producer leader claim the verdict is
+    // `unknown`, designations are withheld, and BOTH surfaces suppress their
+    // ranked chrome — every assertion here would pass against a screen with
+    // no numbers on it at all.
+    robustness: {
+      // A DETERMINATE run. Without a stability number `buildResultsVM` returns
+      // `decisionState: indeterminate`, which NEUTRALISES the cards' ranked
+      // chrome — the screen would then carry no rank affordance to compare.
+      recommendation_stability: 0.9,
+      fragile_edges: [],
+      robust_edges: ['e1'],
+      near_tie: { is_tie: false, top_option_id: leaderId },
+    },
+    response_hash: 'h',
+    meta: { seed_used: '42', n_samples: 1000, detail_level: 'standard', latency_ms: 100 },
+  } as unknown as V2RunResponse
+}
+
+function seedRun(win: Record<string, number>, leaderId: string): void {
+  const report = mapV2ResponseToReportV1(makeV2Response(win, leaderId), { seed: 42 })
+  useCanvasStore.setState({
+    results: { status: 'complete', progress: 100, report } as never,
+    runMeta: {} as never,
+    nodes: OPTIONS.map((o) => ({
+      id: o.id,
+      type: 'option',
+      position: { x: 0, y: 0 },
+      data: { kind: 'option', label: o.label },
+    })) as never,
+    edges: [],
+    hasCompletedFirstRun: true,
+    rawV2Response: null,
+  } as never)
+}
+
+function driveHook(): ReturnType<typeof useResultsSectionData> {
+  const { result, unmount } = renderHook(() => useResultsSectionData())
+  const value = result.current
+  unmount()
+  return value
+}
+
+/**
+ * Run 1 (mints the ordinals through the REAL hook and the REAL PLoT V2
+ * mapper), then run 2 (flips the leader), then render the real mount path on
+ * run 2's data.
+ */
+function renderFlippedBody(): void {
+  useCanvasStore.setState({ optionNumbering: {} } as never)
+
+  seedRun(RUN1_WIN, 'opt_a')
+  driveHook()
+  const frozen = { ...useCanvasStore.getState().optionNumbering }
+  // PRECONDITION pinned in-test: these are the ordinals a real first run
+  // mints. Fabricating them as a fixture would make every assertion below a
+  // statement about my own arithmetic (CLAUDE.md trap 16-inverse).
+  expect(frozen).toEqual({ opt_a: 1, opt_b: 2, opt_c: 3 })
+
+  seedRun(RUN2_WIN, 'opt_b')
+  const data = driveHook()
+  // PRECONDITION: the re-run renumbered nothing. Append-only is the mechanism
+  // the whole defect rests on.
+  expect(useCanvasStore.getState().optionNumbering).toEqual(frozen)
+  // PRECONDITION: the leader actually flipped.
+  expect(data.recommendation.allOptions.map((o) => o.id)).toContain('opt_b')
+
+  render(
+    <ResultsBody
+      resultsSectionData={data}
+      tornadoData={{ rows: [], expectedOutcome: null }}
+      onSendMessage={() => {}}
+      expertMode={false}
+    />,
+  )
+}
+
+beforeEach(() => {
+  useCanvasStore.setState({ optionNumbering: {} } as never)
+})
+afterEach(() => cleanup())
+
+describe('ResultsBody — hero rank and card identity, on one screen, after a flip', () => {
+  it('PRECONDITION: both ranked surfaces are mounted with their chrome on screen', () => {
+    renderFlippedBody()
+    // Without this every assertion below could pass against a screen that
+    // stopped rendering one of the two surfaces (CLAUDE.md trap 13).
+    expect(screen.getByTestId('analysis-hero-panel')).toBeInTheDocument()
+    expect(screen.getByTestId('option-cards')).toBeInTheDocument()
+    expect(screen.getByTestId('hero-option-row-1')).toBeInTheDocument()
+    expect(screen.getByTestId('option-card-opt_b')).toBeInTheDocument()
+    expect(screen.getByTestId('rank-marker-opt_b')).toBeInTheDocument()
+  })
+
+  it('the hero badge states RANK: opt_b flipped into the lead, so its badge reads 1', () => {
+    renderFlippedBody()
+    // Bound to opt_b by identity — the card carries `data-option-id`, and the
+    // hero row is the one containing that option's label. Never "the badge
+    // that reads 1".
+    const heroRow = screen.getByTestId('hero-option-row-1')
+    expect(within(heroRow).getByTestId('hero-row-label')).toHaveTextContent(
+      'Partner with a consultancy',
+    )
+    expect(within(heroRow).getByTestId('hero-row-number')).toHaveTextContent('1')
+  })
+
+  it('BOTH surfaces state the same IDENTITY for opt_b — Option 2 — in the same words', () => {
+    renderFlippedBody()
+    const heroRow = screen.getByTestId('hero-option-row-1')
+    expect(within(heroRow).getByTestId('hero-row-identity')).toHaveTextContent('Option 2')
+    // The card's identity chip is `OptionCards.tsx:732-738`, unchanged by this
+    // slice — the hero adopted ITS wording, not the other way round.
+    expect(screen.getByTestId('stable-number-opt_b')).toHaveTextContent('Option 2')
+  })
+
+  it('⭐ THE PIN: rank and identity are DIFFERENT QUANTITIES for opt_b, and stay different', () => {
+    renderFlippedBody()
+    const heroRow = screen.getByTestId('hero-option-row-1')
+    const heroRank = within(heroRow).getByTestId('hero-row-number').textContent?.trim()
+    const heroIdentity = within(heroRow).getByTestId('hero-row-identity').textContent?.trim()
+    const cardIdentity = screen.getByTestId('stable-number-opt_b').textContent?.trim()
+
+    // Identity agrees ACROSS surfaces...
+    expect(heroIdentity).toBe(cardIdentity)
+    // ...and differs from rank WITHIN the hero, deliberately, on this run.
+    expect(heroRank).toBe('1')
+    expect(heroIdentity).toBe('Option 2')
+    expect(heroIdentity).not.toBe(`Option ${heroRank}`)
+  })
+
+  it('the card states rank WITHOUT a numeral, so no numeral on the card can contradict the hero', () => {
+    renderFlippedBody()
+    // D17: `#N of M` was removed from the card; rank is a colour swatch. The
+    // ONLY numeral in the card header is the identity chip. If a rank numeral
+    // ever comes back to the card, this REDs — because it would immediately
+    // be a second number about a different question sitting next to identity.
+    const marker = screen.getByTestId('rank-marker-opt_b')
+    expect(marker.textContent?.trim()).toBe('')
+  })
+
+  it('CONTROL: the option that LOST the lead shows the mirror image (rank 2, identity Option 1)', () => {
+    renderFlippedBody()
+    // A single row could satisfy every assertion above by coincidence. opt_a
+    // moves the OTHER way — rank 1 -> 2 while identity stays Option 1 — so
+    // the two quantities are shown to move independently, not merely to
+    // differ once.
+    const secondRow = screen.getByTestId('hero-option-row-2')
+    expect(within(secondRow).getByTestId('hero-row-label')).toHaveTextContent(
+      'Hire two developers',
+    )
+    expect(within(secondRow).getByTestId('hero-row-number')).toHaveTextContent('2')
+    expect(within(secondRow).getByTestId('hero-row-identity')).toHaveTextContent('Option 1')
+    expect(screen.getByTestId('stable-number-opt_a')).toHaveTextContent('Option 1')
+  })
+})
--- a/src/components/results/__tests__/ResultsBody.notAnalysedMountPath.spec.tsx
+++ b/src/components/results/__tests__/ResultsBody.notAnalysedMountPath.spec.tsx
@@ -50,6 +50,7 @@ import { describe, it, expect, afterEach } from 'vitest'
 import { render, screen, cleanup, within } from '@testing-library/react'
 import { ResultsBody } from '../ResultsBody'
 import type { ResultsSectionDataReturn } from '../useResultsSectionData'
+import { useCanvasStore } from '../../../canvas/store'
 import type {
   ConfidenceSectionData,
   DecisionResultData,
@@ -150,7 +151,14 @@ const MIXED = [
   EXCLUDED_OPTION,
 ]
 
-afterEach(() => cleanup())
+afterEach(() => {
+  cleanup()
+  // `registerOptionNumbering` is append-only and the store is a module
+  // singleton, so a seeded test would otherwise leak its numbering into every
+  // sibling — and the leak is SILENT (siblings assert absence of ordinals,
+  // which a stale numbering map would quietly start contradicting).
+  useCanvasStore.setState({ optionNumbering: {} })
+})
 
 describe('ResultsBody — the not-analysed card on the mount path', () => {
   it('PRECONDITION: the options block is mounted and the ranked chrome is on screen', () => {
@@ -199,6 +207,46 @@ describe('ResultsBody — the not-analysed card on the mount path', () => {
     const analysedRow = document.querySelector(`[data-option-id="${ANALYSED_A}"]`)
     expect(analysedRow).not.toBeNull()
     expect(within(analysedRow as HTMLElement).getByTestId('hero-row-number')).toHaveTextContent('1')
+  })
+
+  it('NO-RANK covers the hero IDENTITY CHIP, not just the badge', () => {
+    // ⭐ THE BADGE HALF ABOVE IS NOT THE WHOLE SURFACE. `hero-row-number` is
+    // gated on `showOrdinal` (= `!designationsWithheld && row.isRanked`); the
+    // `hero-row-identity` chip beside it is a SECOND place an ordinal can
+    // appear in the same row, and it was gated on `stableNumber != null`
+    // alone. Registration covers `allOptions` — not-analysed options included
+    // — so `stableNumber` is non-null for every row and the chip rendered
+    // `Option 3` next to a card that deliberately renders no ordinal at all.
+    //
+    // This is the seventh-guard failure `NotAnalysedOptionCard`'s header
+    // predicted in prose ("every future stat row added to that card would have
+    // to remember the eighth guard") arriving on the OTHER surface, where the
+    // fork does not protect it. Pinning the chip by its own testid so a future
+    // ordinal-bearing element in this row cannot inherit the badge's pin.
+    // ⚠ THE SEED IS LOAD-BEARING AND ITS ABSENCE IS SILENT. `stableNumberFor`
+    // is all-or-nothing: with no `optionNumbering` in the store EVERY row's
+    // `stableNumber` is null, no chip renders anywhere, and an absence
+    // assertion here would pass while testing nothing. Registration in
+    // production covers `allOptions` — the excluded option included — so this
+    // seeds all three, which is what makes `Option 3` reachable on the
+    // excluded row at all. The positive control below is what proves the seed
+    // took; the first draft of this test failed on that control, not on the
+    // claim, which is how the vacuity was caught.
+    useCanvasStore.getState().registerOptionNumbering([ANALYSED_A, ANALYSED_B, EXCLUDED])
+
+    renderBody(MIXED)
+    const excludedRow = document.querySelector(`[data-option-id="${EXCLUDED}"]`)
+    expect(excludedRow).not.toBeNull()
+    expect(within(excludedRow as HTMLElement).queryByTestId('hero-row-identity')).toBeNull()
+
+    // POSITIVE CONTROL — the chip must still render for an option that WAS
+    // scored. Without this the absence above passes when the chip is deleted
+    // outright, which would silently undo the identity this PR exists to add.
+    const analysedRow = document.querySelector(`[data-option-id="${ANALYSED_A}"]`)
+    expect(analysedRow).not.toBeNull()
+    expect(
+      within(analysedRow as HTMLElement).getByTestId('hero-row-identity'),
+    ).toHaveTextContent(/^Option \d+$/)
   })
 
   it('the WinGauge draws no segment for it', () => {
--- a/src/components/results/__tests__/ResultsBody.notAnalysedMountPath.spec.tsx
+++ b/src/components/results/__tests__/ResultsBody.notAnalysedMountPath.spec.tsx
@@ -154,9 +154,14 @@ const MIXED = [
 afterEach(() => {
   cleanup()
   // `registerOptionNumbering` is append-only and the store is a module
-  // singleton, so a seeded test would otherwise leak its numbering into every
-  // sibling — and the leak is SILENT (siblings assert absence of ordinals,
-  // which a stale numbering map would quietly start contradicting).
+  // singleton, so a seeded test can leak its numbering into every sibling,
+  // and such a leak is SILENT (siblings assert absence of ordinals, which a
+  // stale numbering map would quietly start contradicting).
+  //
+  // ⚠ PROPHYLACTIC, NOT A LIVE FIX — measured: with this reset removed, NO
+  // sibling in this file currently REDs. An earlier revision of this comment
+  // read as though the leak were live. It is not; it is one seeded test away
+  // from being live, which is why the reset stays.
   useCanvasStore.setState({ optionNumbering: {} })
 })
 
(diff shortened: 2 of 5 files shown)


Talchain:
# Adversarial review — FIX-FIRST

Head **`2dbf906d`**, asserted before every read and again before posting (`git ls-remote` + `gh api` agree). Fresh blobless clone at a unique `/private/tmp` path; mutants in a `--detach` worktree **outside** the repo root with **isolation proven by writing a sentinel** (inodes `750420906` vs `750699311`; source SHA-256 `3692058d…` unchanged across the write), restores from a pristine archive outside both trees, clean tree asserted **before and after** every mutant.

**The fix is right and its evidence is real.** I reproduced the whole battery independently and every headline claim about the *code* holds. One thing it does not hold about is the **not-analysed option**, and that is a regression on the exact axis this PR exists to protect.

---

## What I confirmed (the author is right, the brief was wrong)

**Claim 1 — the two `analysisDisplaySelector.spec.ts` cases must NOT change. CONFIRMED.** They pin `selectDisplayOptions`, untouched here, which has **zero production consumers** (its own doc says "no live call site today"; sweep of `src/` non-test files returns nothing, contrast control `stableNumber` → 67 hits, `displayIndex` → 12). `:79-88` really is the selector-level statement of the distinction this PR renders. Nothing was left pinning old behaviour.

**Claim 2 — `row.index` IS `displayIndex`. CONFIRMED.** `analysisDisplaySelector.ts:50` `displayIndex: i + 1` and `buildHeroModel.ts:403` `index: i + 1`, both over `sortOptionsForDisplay(...)` with the same `designationsWithheld` flag. Same expression, same comparator. No fourth spelling was minted.

**Claim 3 — the chip gate. CONFIRMED at the bytes for the ranked card.** `OptionCards.tsx:724` gates the rank swatch on `!neutralised && rank != null && totalOptions > 1`; `:732` gates the identity chip on `stableNumber != null` and nothing else; `:1382` passes it regardless of `designationsWithheld`. I drove a **withheld run** end to end: badges `0`, identity chips `['Option 1','Option 2','Option 3']`. Consistent with the card on the same screen — so on the *withheld* path the author's reasoning holds.

**Claim 4 — the discriminating pair. REPRODUCED EXACTLY.**

| mutant | applied | result | PR table |
|---|---|---|---|
| M0 control (no match) | 0 | 107/107 green | ✔ |
| M1 badge = `stableNumber ?? index`, all rows | 1 | **9 RED** | ✔ 9 |
| M2 same, `row.index === 1` exempt | 1 | **4 RED** | ✔ 4 |
| M3 identity chip deleted | 1 | **7 RED** | ✔ 7 |
| M4 hero chip fed rank | 1 | **7 RED** | ✔ 7 |
| **M4b (mine) CARD chip fed rank** | 1 | **3 RED incl. ⭐ THE PIN** | claim holds |
| **M6 (mine) model-level `index := stableNumber`** | 1 | **11 RED** | — |
| **M7 (mine) drop `row.isRanked` from `showOrdinal`** | 1 | **1 RED** | — |
| M5 trailing control | 0 | 107/107 green | ✔ |

- *"the badge on the new leader row carries its CURRENT RANK…"* — **RED under M1, GREEN under M2** ✔. *"every badge in the list reads its display rank"* — RED under **both** ✔. The pair binds.
- **The cross-surface pin REDs from both directions**, and I checked the direction the PR's own M4 did **not** cover: **M4b relabels the *card's* chip with rank** and REDs 3 cases including ⭐ THE PIN. The spec header's card-side claim is true, not just asserted.
- **M6** matters because `data-testid={\`hero-option-row-${row.index}\`}` (`HeroOptionRow.tsx:395`) is built from **the same field the badge now renders** — a coupling that could have made these assertions agree with themselves. It does not: `flipAndBuild`'s in-test preconditions (`rows[0].index === 1`, `stableNumber === 2`) and the cross-surface spec's **label-text binding** catch a model-level harmonisation, 11 RED. That is trap 13b done correctly and it is worth saying out loud.

**Integration.** Clean merge of current `origin/staging` (`48b567db`, 3 commits ahead: #879, #897, #893) into `2dbf906d` → **107/107 green**. **CI re-derived at the exact SHA**, filtering on `.status == "completed"` first (trap 24b): `completed=14, running=0, failed=["Visual Regression (advisory)"]`, `mergeable=true`. Required context on `staging` is `["Staging Gate"]` → **success**. Fresh contrast control: the same advisory check is **also failing on staging tip `48b567db`** with `running=0` — so it carries no information about this branch, as claimed.

---

## ⛔ GATE 1 (BLOCKING) — the not-analysed row: this PR opens the exact cross-surface disagreement it exists to close

`HeroOptionRow.tsx:293` gates the identity chip on `row.stableNumber != null` **only** — never on `row.isRanked`. `AnalysisHeroPanel.tsx:485` gates the **badge** on `row.isRanked`, but nothing gates the **chip**.

`NotAnalysedOptionCard.tsx:68`, verbatim:

> `No rank swatch and no "Option N" — the two ordinals the ranked card renders in this slot. The gap is the point.`

and `:20-25`: *"No rank marker · no ordinal · … Not "rendered as zero", not "rendered as '—'": absent."* That is Paul's NO-RANK ruling of 14 Aug, and the estate's existing ruling treats **`Option N` as one of the two ordinals to withhold**.

**Reachability chain, all live, no flag:**
`useResultsSectionData.ts:3841-3847` registers `optionNumbering` over `sortOptionsForDisplay(recommendation.allOptions)` — **`allOptions` includes not-analysed options** → `buildHeroModel.ts:285-289` `stableNumberFor()` returns non-null for **every** row (all-or-nothing) → the hero draws `Option N` beside an option whose own card, on the same screen, deliberately draws none.

**MEASURED**, on the real mount path (`<ResultsBody>`, real `OptionCards` + real `AnalysisHeroContainer`), with `optionNumbering` seeded exactly as production seeds it:

```
PROBE hero identity for excluded option = Option 3
  card: queryByTestId('stable-number-opt_migrate')  -> absent   (existing contract, held)
  hero: queryByTestId('hero-row-number')            -> absent   (existing contract, held)
  hero: queryByTestId('hero-row-identity')          -> "Option 3"  <- NEW
POSITIVE CONTROL: an analysed row in the same list carries 'Option 1'.
```

Pre-PR that row rendered **no number of any kind**. This is a regression, not merely a coverage gap.

`ResultsBody.notAnalysedMountPath.spec.tsx:178` already asserts the **card** half (`stable-number-${EXCLUDED}` absent) and `:182-202` the hero **badge** half — the mirror assertion for the chip was never written, which is why the suite stayed green. (M7 confirms the badge half still bites: 1 RED.)

**Smallest remedy — measured, one line:**

```diff
-        {row.stableNumber != null && (
+        {row.stableNumber != null && row.isRanked && (
```

With it applied: the chip disappears from the excluded row (`identity = undefined`) and **all 107 tests across the five relevant spec files stay green** — it costs nothing. Plus one assertion in `ResultsBody.notAnalysedMountPath.spec.tsx`, mirroring the card's own absence assertion, with the existing analysed-row positive control.

The `showOrdinal` doc's sentence *"An un-analysed option keeps its identity for the same reason"* is the false premise here, and it is the same shape as the premise this PR corrects: **written from the withheld case in hand, then generalised to a per-row case governed by a different ruling.** `designationsWithheld` (per-run) and `isRanked` (per-row) are trap 21 twins, and `heroTypes.ts` says so explicitly — the badge honours the distinction; the chip does not.

---

## GATE 2 (fix-first, cheap) — the committed fixtures do not produce the witnessed artefact, and the self-disclosure names the wrong lens

Measured at `2dbf906d` through the real hook and the real V2 mapper:

```
defaultLens = 'outcome'   lenses = ['outcome']   (no goal lens at all — the fixture sets no goal threshold)
leaders     = {goal: null, outcome: 'opt_a', ...}
rows        = opt_b(index 1, stable 2), opt_a(index 2, stable 1), opt_c(index 3, stable 3)
filledPositions = [2]  -> the FILLED row is opt_a, at position 2; the top row is NOT filled
```

Two consequences:

1. **The disclosure's stated cause is wrong.** It says *"the default lens is Goal fit, whose leader is row 2"*. The default lens is **`outcome`**, and the goal lens is not even available in this fixture. The *substance* is right (the filled row is row 2, `opt_a`, frozen ordinal 1); the reason given was written from memory, not measured — which is the same class of error the PR is correcting elsewhere.
2. **The headline artefact is not reproduced by either committed spec.** The PR says *"reproduced deterministically: the new leader's badge is filled leader-blue and carries the numeral 2."* In these fixtures the pre-fix filled badge reads **1** (at position 2), and the top row's badge reads 2 but is **not filled**. The case titled *"the FILLED badge's numeral is its own row's position (**the witnessed contradiction**)"* is pinning a real contradiction, but **not the witnessed one**.

Because `option_comparison.outcome.mean` is held constant across the two runs while only `win_probability` moves, the outcome leader can never flip. **Remedy, measured:** move `mean` with `win_probability` (`opt_a: 30→24`, `opt_b: 24→30`). Result:

```
leaders.outcome = 'opt_b'   filledPositions = [1]   filledRowId = opt_b
PRE-FIX the filled badge would have read = 2      <- the witnessed artefact, exactly
```

Or keep the fixture and retitle the case. Either is fine; asserting *"the witnessed contradiction"* about a different one is not.

---

## GATE 3 (fix-first, comments + PR body) — the cited precedent does not say what the PR says it says

Both are load-bearing, because *"no third convention was invented; the hero converged on the existing one"* is the argument the whole design rests on.

1. **`canvas/nodes/OptionNode.tsx` does not render the text `Option N`.** `:1206-1213` renders a **bare numeral in a bordered box** — `{stableOptionNumber}`, `aria-label="Option N"`, `h-4 min-w-[16px] rounded border border-panel-border`. The comment at `HeroOptionRow.tsx:271-273` and the PR body both claim it "mirrors the same chip". It does not. **The consequence was never priced:** a small bordered numeral box means **identity** on the canvas and, after this PR, **rank** in the hero. Before this PR the hero badge and the canvas node badge *agreed* (both `stableNumber`); they now disagree at the glyph, and the reconciling information lives in a different element. The PR's exclusions flag that OptionNode was never *mounted* in a test — correct — but the design claim about it is stated as settled.
2. **`OptionCards.tsx:723` is `flex items-center gap-2` — no `flex-wrap`** (`grep -c flex-wrap OptionCards.tsx` = 0; contrast `grep -c "flex "` = 6). The hero's new wrapper is `flex min-w-0 flex-wrap items-baseline gap-x-1.5`. `flex-wrap` and `items-baseline` are **additions to** the precedent, not adoptions of it — see Gate 4.

Remedy: correct the two comments and the PR body. If the hero↔canvas glyph divergence is accepted, say so deliberately rather than by omission.

---

## GATE 4 (decide, do not defer) — a layout change with zero visual evidence, and the visual suite structurally cannot supply it

The exclusion is honest and I am making **no pixel claim**. What I can add is what the instruments would and would not tell you, measured.

**The advisory red genuinely carries no information** — confirmed with a *fresher* contrast control than the PR's: `Visual Regression (advisory)` fails on staging tip `48b567db` with `running=0`. Note the PR's own contrast pointed at `56059e18`, where the 10 linux references still existed; **#879 deleted them** (`git ls-tree` at `48b567db` → **0** linux references) after this branch was cut. Right conclusion, stale evidence.

**But restoring the references would not help this PR.** The suite's five scenarios are `fresh-draft`, `blocked-provisional`, `model-tab`, `olumi-tab`, `graph-default-zoom`. `e2e/visual/harness.ts:131` and `e2e/visual/states.visual.spec.ts:10-24` state that **`completed-analysis` is deliberately not covered**. Hero option rows exist **only** on a completed analysis. So "commit the linux references" is not the path to evidence here, and the PR's exclusion implicitly frames it as though it were.

**What the change actually does to layout, from the compiled CSS — mechanism, not pixels.** I compiled every new class against the repo's real Tailwind config (the `chartClassesCompile.spec.ts` technique; negative control `gap-x-1.7777` correctly absent, so the matcher discriminates):

```
flex          -> display: flex             flex-wrap     -> flex-wrap: wrap
min-w-0       -> min-width: 0px            items-baseline-> align-items: baseline
gap-x-1.5     -> column-gap: 0.375rem      flex-none     -> flex: none
line-clamp-2  -> display:-webkit-box; -webkit-line-clamp: 2
```

No silent-no-emit defect. But the chip is `flex: none` (never shrinks) and the label keeps `flex: 0 1 auto`, so its **hypothetical main size is its max-content width**; with `flex-wrap: wrap`, any label wider than *(column 2 − chip − 0.375rem)* **starts a new flex line**. The row header then becomes *chip on line 1, clamped label on lines 2–3* — up to three text lines where it was two — and the effect is **data-dependent** (short labels do not wrap), which is exactly the kind of thing one screenshot at one width may not reveal either. This is derived from the declarations above; **nothing has measured it**.

**Answer to the gate question, plainly: no, I do not think a layout change to a rendered row should merge on zero visual evidence when a cheaper instrument exists in this repo and is idle.** The minimum substitute is **not** a screenshot suite:

> `playwright.geometry.config.ts` + `e2e/geometry/*.measure.ts` — real Chromium, real layout, run deliberately, in no gate. `e2e/geometry/analysisAnswerFirst.measure.ts` **already replays a real captured CEE analysis turn through `applyV5State`** — the exact setup needed to get hero rows on screen.

One `heroOptionRow.measure.ts` asserting, at 1280 and 1440: (a) `hero-row-identity` has **non-zero** width and height — the fake-control discipline that separates a real 47×21 from a jsdom 0×0; and (b) the chip and the label's first line **share a line box** (or: report that they do not), which is precisely the `flex-wrap` question above. That is a bounded piece of work on an existing harness, and it converts this exclusion from "nothing has looked at it" into a number.

If the call is to merge without it, that is defensible — but it should be recorded as **accepted risk on a named, data-dependent row-height change**, not carried as an exclusion.

---

## GATE 5 (minor) — one assertion is a tautology, and its comment credits it with the discriminating power it does not have

`rerunFlipNumbering.rankCoherence.spec.tsx`, last case:

```ts
expect(rank).not.toEqual(identity)   // '1'  vs  'Option 2'
```

The comment above it: *"If a future 'harmonisation' makes them equal, this REDs — which is the whole reason it is written as an inequality rather than as two separate equalities."* **It cannot RED under any mutation** — a bare numeral can never equal a string beginning `Option `. **Measured:** under M1 that case fails on the **next** line (`AssertionError: expected '2' to be '1'`, i.e. `expect(rank).toBe(String(chart.rows[0].index))`); the inequality passed. The file's real binding is those two `toBe` lines.

The **cross-surface** spec's version is sound — it compares like with like: `expect(heroIdentity).not.toBe(\`Option ${heroRank}\`)`, and M4/M4b both RED it. So the *technique* is right; one of the two instances is written against the wrong comparands.

Also harmlessly vacuous, `content.spec.tsx`: `expect(getByTestId('hero-row-number')).not.toBe(getByTestId('hero-row-identity'))` — two distinct test-ids always return distinct nodes.

Remedy: compare like with like (`expect(identity).not.toBe(\`Option ${rank}\`)`), or delete the line and let the comment describe the `toBe` pair that actually binds.

## GATE 6 (minor) — the withheld path gained a new rendered claim with zero coverage

Measured: on a withheld run the hero now draws `Option 1 / Option 2 / Option 3` top-to-bottom in canonical order where it previously drew nothing. It matches `OptionCards` on the same screen, so it is **not** a disagreement and I am not asking for a code change. But it is a new render on the ROADMAP 1.267 claim-gated path and no assertion in this PR touches it. One case in `withheldDesignations.hero.spec.tsx` — badges 0, chips 3 — beside the existing `renders no ordinal number token`.

---

## What this evidence EXCLUDES

- **No pixel claim of any kind.** Nothing here was rendered in a browser. The `flex-wrap` consequence is derived from compiled CSS declarations; no width, no row height, no screenshot, no reference image.
- **CODE EXISTS → TESTED only.** No deployed, mounted, wire- or journey-witness. Gate 1's reachability is measured in **jsdom on the real mount path with production-shaped seeding** — it is not a staging witness.
- **I did not run the full suite** (CI did, and I re-derived it at the SHA). My runs were **five named spec files, 107 tests, counts asserted by name**, plus three throwaway probes. A green result here is not evidence about any file outside that set.
- **The canvas `OptionNode` was not mounted.** The glyph-divergence finding in Gate 3 is derived from the bytes of both components, not from a rendered screen.
- **M4b, M6 and M7 are my mutants, not the author's**, and a survivor in any battery is a claim about sensitivity, never about whether an expectation is right.
- **Accessibility:** I measured that the row button's accessible name changed (`getByRole('button', { name: 'Partner with a consultancy' })` no longer resolves — it did not before either, because the readout was already in the name). No screen reader was used, and no existing spec broke.

---

## Verdict

**FIX-FIRST.** Gate 1 is the blocking one: a one-line gate plus one mirror assertion, both measured, neither costing a single existing test. Gates 2, 3 and 5 are corrections to claims the PR makes about its own evidence and about its cited precedent — cheap, and this estate pays for stale premises left in comments. Gate 4 is a decision to take deliberately rather than to inherit.

The core change is correct, the discriminating pair genuinely discriminates, and the cross-surface pin genuinely REDs from both directions. Fix Gate 1 and I would approve.

Talchain:
## Fix round pushed — `2dbf906d → 3ea14eff`

**GATE 1 closed.** Taken by the integration lane because this PR's author lane no longer exists; the remedy is the one the review specified, and the delta is verifiable at the bytes.

### The gate, and a sharper root cause than "add a conjunct"

The remedy is `&& row.isRanked`, as ordered. The *reason* is worth recording, because the repo had already written the warning:

`showOrdinal` is `!designationsWithheld && row.isRanked` — **a conjunction of two questions**, and `heroTypes.ts:110-124` states the distinction verbatim:

> *This is PER-ROW and NOT the same question as `designationsWithheld` … (trap 21: two authorities, two questions, kept apart). `designationsWithheld` asks "may this run designate a leader?"; `isRanked` asks "was this option scored?" … **which is exactly the case that shipped an ordinal it had no basis for.***

Declining to gate the chip on `designationsWithheld` was **correct** — withholding suppresses designations and identity is not one. It also silently declined the `isRanked` conjunct riding in the same variable. The old comment's *"an un-analysed option keeps its identity for the same reason"* was the false step; it is **corrected in place with its reason**, not deleted.

Confirmed at the bytes: `NotAnalysedOptionCard.tsx:68` — *"No rank swatch and no 'Option N' … the gap is the point"* under the NO-RANK ruling (14 Aug 2026); `buildHeroModel.ts:411` `isRanked: o.notAnalysed !== true`; `AnalysisHeroPanel.tsx:485` supplies the conjunction to the badge.

### Evidence

| | |
|---|---|
| RED-first | the new pin REDs at pristine **on the claim** (`expected <span> to be null`), while the existing badge pin stays **GREEN** — so the badge half was already covered and the chip half was genuinely the gap |
| **M1** remove `row.isRanked` | REDs **only** the new chip test |
| **M2** ungate the **badge** instead (different object) | REDs **only** the existing badge test |
| | **Disjoint RED sets** — the pin binds to the chip, not to "this row has no numbers". `applied_src=1` each, controls `applied_src=0`, green both ends |
| Sweep **by symbol** | 8 specs, **134 collected, 0 failed**, counts asserted per file |
| Named gate | `pnpm typecheck` exit 0 (baseline 2250, unchanged) · `eslint` exit 0 |

### ⭐ The seed is load-bearing, and its absence was silent

`stableNumberFor` is all-or-nothing: with no `optionNumbering` in the store, **every** row's `stableNumber` is null, no chip renders anywhere, and the absence assertion passes while testing nothing.

**My first draft of this test failed on its POSITIVE CONTROL rather than on the claim.** That is how the vacuity was caught — the control was doing exactly the job the sibling test's own comment says it exists for. The test now seeds registration for all three options as production does, and resets it in `afterEach`, because the store is a module singleton and an append-only leak into sibling *absence* assertions would be silent.

### Prose corrected (no behaviour)

Three claims that `canvas/nodes/OptionNode.tsx` renders the same `Option N` **text**. It does not — `:1206-1213` draws the **bare numeral** in a bordered box and carries `Option N` as its **`aria-label` only**. The surfaces agree on the **accessible name** and deliberately differ in glyph (a canvas node has no room for the word). Corrected rather than deleted, because the aria agreement is the stronger property.

`rerunFlipNumbering.rankCoherence.spec.tsx` is **comment-only in both directions** (0 non-comment lines added and removed, with the same filter returning non-zero on `HeroOptionRow.tsx` as a control).

### ⚠ NOT addressed — stated rather than dropped

1. **The committed fixtures never produce the PR's headline artefact** (outcome `mean` held constant so the fill cannot flip). Real, and a fixture change, not this gate.
2. **The self-disclosure names the wrong lens** (`outcome`, not Goal fit).
3. **"One assertion is a tautology whose comment credits it with binding it doesn't have"** — I could not locate this unambiguously and did not want to guess at which one. If you name the line I will take it.

Items 1–2 are PR-body/fixture work; none of the three is the blocking gate. Flagging them here so they are visible rather than quietly lost.

Talchain:
# Fix-round verification — GATE 1 CLOSED. Approve, with two tidies that are one sentence each.

**Interval: `2dbf906d..3ea14eff`** — one commit, and `2dbf906d` **is an ancestor** of the new head (linear, not rewritten), so the interval is the fix round and nothing else. Everything below is measured on that interval, never on `base..head`. Head re-asserted against `git ls-remote` before posting.

## 1. The delta is what you say it is — plus a third file your description omitted

```
ResultsBody.notAnalysedMountPath.spec.tsx   | 50 +++++++++++-      (new pin)
HeroOptionRow.tsx                           | 48 ++++++-----        (1 code line + comments)
rerunFlipNumbering.rankCoherence.spec.tsx   | 14 ++++--             <- not in your list
```

The third file is **comment-only**, and I proved it mechanically rather than by eye — stripped block/line comments and empty JSX expression containers from both revisions and diffed:

```
rerunFlipNumbering.rankCoherence.spec.tsx : CODE IDENTICAL after comment-strip (5571 / 5571 bytes)
HeroOptionRow.tsx                         : CODE DIFFERS — tokens added: "row.isRanked", "&&"   (8738 / 8754)
```

Non-zero stripped sizes are the positive control on the stripper — an instrument that emitted nothing would report "identical" for everything.

So across all of `src`, **the only code tokens the fix round adds anywhere are `row.isRanked` and `&&`.** The rest is comments, and the comment changes are corrections to the two claims I flagged (the `OptionNode` glyph, and the false "an un-analysed option keeps its identity for the same reason"). Both now say the right thing, and the second one is corrected *in place with its own history visible*, which is the right treatment.

One thing I want to credit rather than pass over: the new comment doesn't just add the guard, it names **why** — `showOrdinal` is a conjunction of two questions and the chip answers only one of them. That is the trap-21 distinction `heroTypes.ts` already states, now enforced at both elements instead of one.

## 2. Disjoint RED sets — reproduced, and they are singletons

Worktree at `3ea14eff`, `--detach`, outside the repo root, isolation **proven by writing a sentinel** (inodes `750420906` vs `751630965`; source SHA-256 unchanged across the write), restores from a pristine archive outside both trees, clean tree asserted before and after every mutant. Baseline **5 files / 108 tests** (was 107 at `2dbf906d`; +1 is the new pin). Collected count asserted by name on every run; no run produced "No test files found".

| mutant | applied | result | RED set |
|---|---|---|---|
| M0 control | 0 | 108/108 green | — |
| **M1** remove `&& row.isRanked` (undo the fix) | 1 | 1 failed / 107 | **only** `NO-RANK covers the hero IDENTITY CHIP, not just the badge` |
| **M2** ungate the badge instead (`showOrdinal={!model.designationsWithheld}`) | 1 | 1 failed / 107 | **only** `NO-RANK on the OTHER surface too: the hero lists it with no ordinal` |
| M3 seed omits `EXCLUDED` | 1 | 1 failed / 107 | the new pin — **on its positive control** |
| M4 delete the `afterEach` reset | 1 | **108/108 green** | — |

**M1 and M2 RED disjoint singletons.** Neither pin inherits the other's coverage: the badge guard cannot see a chip regression and the chip guard cannot see a badge regression. That is the argument, and it now holds in both directions rather than being asserted.

## 3. ⭐ The vacuity you were worried about — the control is not just present, it is *sufficient*

**M3 fails with `Unable to find an element by: [data-testid="hero-row-identity"]`** — i.e. on the positive control, exactly the failure your first draft hit. So the seed is genuinely load-bearing and its absence is caught.

The part worth stating explicitly, because it is *why* the analysed-row control is enough to prove the seed reached the **excluded** row: `stableNumberFor` is all-or-nothing over `sortOptionsForDisplay(recommendation.allOptions)`, and `allOptions` **includes** the not-analysed option. So if `EXCLUDED` were unregistered, `.every()` fails and **every** row's `stableNumber` goes null — the analysed row loses its chip too. The control therefore proves registration covered all three ids transitively; it is not merely "a chip rendered somewhere". M3 is the demonstration of that, not the argument for it.

**The `afterEach` reset is real** — verified directly, since M4 alone could not tell "correct but unneeded" from "silently broken":

```
registerOptionNumbering(['a','b','c'])      -> { a:1, b:2, c:3 }   (seed took)
setState({ optionNumbering: {} })           -> {}                  (reset real)
registerOptionNumbering(['z'])              -> { z:1 }             (append-only history gone)
registerOptionNumbering still a function                            (negative control: partial setState kept the slice)
```

**But be clear about its status: M4 shows no sibling currently REDs without it.** Nothing in the file today depends on it, so it is prophylactic — correct hygiene for the next sibling that asserts an absence, not a pinned property. That is the right call to have made; I am only flagging that its own comment reads as though the leak were currently live, and it is not.

## 4. Your question 2 — the tautology, named

**`src/components/results/analysis-hero/__tests__/rerunFlipNumbering.rankCoherence.spec.tsx`**, in `it('rank and identity DISAGREE on this run, and that is the point — each is labelled')` (line 256):

- **Line 265 — the tautology:** `expect(rank).not.toEqual(identity)`
- **Lines 261-264 — the comment that credits it:** *"If a future 'harmonisation' makes them equal, this REDs — which is the whole reason it is written as an inequality rather than as two separate equalities."*
- **Lines 266-267 — where the binding actually lives:** `expect(rank).toBe(String(chart.rows[0].index))` and `expect(identity).toBe(\`Option ${chart.rows[0].stableNumber}\`)`

`rank` is the badge's `textContent` — a bare numeral, `'1'`. `identity` is the chip's `textContent` — `'Option 2'`. A bare numeral can never equal a string beginning `Option `, under any mutation, including the exact harmonisation the comment names. **Measured at `3ea14eff`** with the badge pointed back at `stableNumber ?? index`:

```
rankCoherence.spec.tsx:229:7   AssertionError: expected [ '2', '1', '3' ] to deeply equal [ '1', '2', '3' ]
rankCoherence.spec.tsx:235:22  AssertionError: expected '2' to be '1'
rankCoherence.spec.tsx:266:18  AssertionError: expected '2' to be '1'      <- the DISAGREE case
```

Line **266**, not 265. Line 265 passed under the very mutant its comment says it exists to catch.

**Remedy** — compare like with like, as its own sibling already does:
```diff
-    expect(rank).not.toEqual(identity)
+    expect(identity).not.toBe(`Option ${rank}`)
```
or delete line 265 and let the comment describe the `toBe` pair on 266-267 that does the work.

⛔ **Do not touch `ResultsBody.crossSurfaceOptionNumbering.spec.tsx:226`** — `expect(heroIdentity).not.toBe(\`Option ${heroRank}\`)` is the **sound** instance of the same idea, and I measured it REDing from both directions (hero chip fed rank, and card chip fed rank). It is the model, not the defect.

Second, minor and harmless, same family — **`content.spec.tsx:642`**: `expect(getByTestId('hero-row-number')).not.toBe(getByTestId('hero-row-identity'))`. Two distinct test-ids always return distinct nodes. It asserts nothing, but it also misleads nobody; take it or leave it.

## 5. Your question on the two still-open items — my call

**Neither is a merge gate.** The mutation battery proves the guard binds regardless of which instance of the contradiction the fixture produces, and both items are accuracy-of-the-record, not behaviour. But the *sentences* should not ship as written, because a false premise left in a comment is precisely what caused the defect this PR fixes — and correcting a sentence is cheaper than rowing it.

**(b) The wrong lens — fix in this round.** One sentence. The disclosure says *"the default lens is Goal fit, whose leader is row 2"*. Measured: `defaultLens = 'outcome'`, `lenses = ['outcome']` — the goal lens is not available at all in that fixture (it sets no goal threshold). The substance is right (the filled row **is** row 2, `opt_a`, frozen ordinal 1); only the stated cause is wrong. Zero risk to change.

**(a) The fixture — retitle now, refixture as a follow-up row.** Drop or soften the parenthetical in `it('the FILLED badge's numeral is its own row's position (the witnessed contradiction)')` — it pins a real fill/numeral contradiction, just not the witnessed one. That is a one-word edit with no risk.

The fixture change itself is a **row**, not this round, and deliberately so: `option_comparison.outcome.mean` is what decides which lenses are available and therefore which row is filled, so moving it can silently change the lens set and hence what the specs are pointed at — it needs its own RED-first and its own battery re-run. Hand the follow-up this, so it is one edit and a re-run rather than a re-derivation:

> Move `mean` with `win_probability` (`opt_a: 30→24`, `opt_b: 24→30`). Measured result: `leaders.outcome = 'opt_b'`, `filledPositions = [1]`, `filledRowId = opt_b`, and **pre-fix the filled badge reads `2`** — the witnessed artefact exactly.

**Still open and unchanged from my review, both your call, neither blocking:** the layout question (Gate 4 — `flex-wrap` on the new wrapper is an addition to the `OptionCards` precedent, not an adoption of it; the visual suite deliberately excludes `completed-analysis` so restoring the linux references would not cover this; `playwright.geometry.config.ts` is the instrument that would). And Gate 6 — on a withheld run the hero now draws `Option 1/2/3` where it drew nothing; it matches `OptionCards` on the same screen so it is not a disagreement, and it remains unpinned by one assertion in `withheldDesignations.hero.spec.tsx`. `row.isRanked` is independent of `designationsWithheld`, so the fix round does not change that behaviour.

## What this round's evidence EXCLUDES

Scope was **the interval `2dbf906d..3ea14eff`**, not a re-review. I re-ran **five named spec files, 108 tests, counts asserted by name** — a green result here says nothing about any file outside that set, and CI remains the authority on the full suite. No pixel claim: nothing was rendered in a browser this round either. Still `CODE EXISTS → TESTED`; no deployed, mounted, wire- or journey-witness. M1-M4 and the reset probe are mine, and a mutant battery measures sensitivity, never whether an expectation is right.

---

## Verdict: APPROVE

Gate 1 is closed properly — gated at the right predicate, for the stated reason, pinned by a test that binds to its own element and whose control is provably sufficient rather than merely present. Take the line-265 fix, the one-sentence lens correction and the retitle in this round; row the fixture. Nothing else stands between this and merge.

PR #900, opened 2026-08-27
chore(dead-code): delete PreAnalysisGuidance, useCanRunAnalysis and limitsManager

Talchain:
Delete-first batch 3. **1,236 deletions, 2 insertions** — both insertions are the `# count=` comment line in a baseline file. Zero non-comment production lines added.

Measured at tip `48b567db`, which is **byte-identical to what staging currently serves** (`version.json` → `48b567db`), so the bundle evidence below is about the build users actually load.

## ⭐ The finding that matters more than the deletions

`PreAnalysisGuidance.gaps.spec.ts` (311 lines) **does not import the module it is named after.** Its own header says it *"mirrors PreAnalysisGuidance.tsx"* — it re-implements the `gapItems` logic inline and tests the copy. It has been green for five months irrespective of whether the component existed, and would have stayed green if the component had been deleted at any point in that window.

**A guard agreeing with itself.**

And a QA record rested on it. `ia-v1-qa-findings.md` cited this spec as PASS evidence for **eight acceptance criteria** (E1, E2, E5–E10) on a component **absent from the deployed bundle**. **Those eight were never covered** — not coverage that rotted, but coverage that never existed: at `48b567db`, `gapItems` and `voiScore` lived only in the two deleted files, and the spec passes with no component on disk (19 passed / 0 failed).

**Remedied in this PR** (second commit): the eight rows are marked **WITHDRAWN in place**, not deleted, so the original claim stays on the record. E3 (CODE VERIFIED), E4 (MANUAL) and J6 also describe the deleted component, and the Section E summary counts are correspondingly overstated — both flagged in the note, deliberately not rewritten. The filename carries no date, so by the estate's naming rule this is a living document; its header does carry an internal `Date: 2026-03-19`, which is why the rows are withdrawn in place rather than edited away.

⛔ **I have deliberately not edited `ia-v1-qa-findings.md`.** It is a dated record of what was claimed on a given day; rewriting it would falsify the record. Flagging it, not fixing it.

This is the second instance today of an instrument reporting PASS about something it never touched — a sibling lane found `e2e/guide/03-trust-features.spec.ts` printing `GATES: PASS — Provenance panel check` with no `expect()` call anywhere in the file. Two independent finds, same shape, same afternoon. A stale QA record that reads as evidence is what teaches the next person to stop checking.

## What was deleted, and the gate each cleared

Seven-way gate per candidate — static / dynamic / registry / build-flag / bundle / open-PR / age — every absence sweep carrying a contrast control in the same invocation.

### `PreAnalysisGuidance.tsx` (728 LOC) + its spec (311 LOC)
- **Static:** 0 external references repo-wide. Contrast `OutputsDock` = 768.
- **Dynamic / registry:** no `lazy()`, `import()`, `require`, no barrel, no route. Single export.
- **Build flags:** moot — no posture can mount a module with zero import sites.
- **Bundle:** proven-distinctive literals `pre-analysis-guidance` and `Configure option interventions` → **0 of 81 chunks**, while controls `key-question-card` / `outputs-dock` / `pre-analysis-panel` each → 1 and a fabricated control → 0. A real absence, not a blind crawl.
- **Age:** added 2025-12-10, last touched 2026-03-19.
- All seven of its imports retain other consumers — nothing downstream is orphaned.

### `useCanRunAnalysis` (4-line wrapper in `usePreRunValidation.ts`)
- **Static:** one hit repo-wide — its own declaration. Contrast `usePreRunValidation` = 68.
- **Age:** introduced 2025-12-29.
- **Bundle: no evidence claimed, in either direction.** A minified function identifier is not a surviving string literal, so that probe cannot answer. Reporting no evidence rather than a zero.

### `limitsManager.ts` (166 LOC)
- Orphaned by **#893**, which removed `graphPreflight.ts` — its only importer. Re-verified *after* that merge landed, not predicted before it.
- **Static:** zero imports of any form — symbol, path, class, side-effect, dynamic. `hydrateLimitsAtBoot` has zero callers.
- **Age:** added 2025-11-17, never modified since.
- **Bundle: no evidence claimed.** Its three distinctive literals live only inside `console` calls, and the production build drops console (`terser drop_console` + esbuild `drop: ['console']`). The bracket-literal positive control returned 0, which proved the probe blind before its output was trusted — so it is not reported as a zero.

## Baselines

`2249 → 2219` (27 + 3 errors removed) in both `typecheck-baseline.txt` and `typecheck-baseline-identities.txt`, edited by hand and then verified against the gate's own three checks:

- header `# count=` equals row sum in **both** files ✅
- identities-derived per-file counts equal `typecheck-baseline.txt` ✅
- all identity rows well-formed ✅

`ds-compliance-baseline.json` is **deliberately untouched**: that ratchet gates only net-new violations and explicitly allows a downward move, and it has no stale-entry check. `scripts/isolation-results/external-files.txt` is likewise untouched — it is regenerated by its script, not a checked-in expectation, and is not a CI gate.

## Spared

- **`usePreRunValidation` — SPARED. Dark is not dead.** It has **four** live production importers (`PreAnalysisPanel.tsx:42`, both `usePreAnalysisData.ts` files, and `PreAnalysisGuidance.tsx:20` — the last removed by this PR, leaving three). Unreachable-under-current-flag-posture is a route-retirement question, not a zero-consumer deletion.

## Lead for batch 4 — not a claim

`'Consider adjusting this weight'` returns **0 of 81 chunks** while `GuidancePanel.tsx` contains it and has five live importers. That could mean the string is dead, the component is unmounted, or the crawl missed a chunk. Those are three different findings and there is evidence for none of them yet. Rowed, not chased.

## Collision

Rebased onto `399b1f2f` (#899). Baselines **re-derived with the gate's own `--update-baseline`**, not hand-merged — the regenerated files are **byte-identical** to the hand edit, and the gate passes at 552 files / 2219 errors. Controls: a +1 perturbation of the identity baseline REDs the cross-file check, and the restored tree passes.

⚠ **#901 will conflict with this PR on the `# count=` line, in either merge order.** Fail-loud rather than silent — but the naive resolutions (2219 or 2247) both **RED the gate**; only the combined value **2217** passes. Sequencing note for whoever lands second.

No other collision: a re-sweep of all open PRs (paginated, with a positive control proving the sweep discriminates) shows none touching a deleted path or either baseline.

**`Visual Regression (advisory)` is a standing broken alarm, not a signal.** On both base and this head the comparison step is *skipped* and the job fails at a step named *“Fail — linux references absent, nothing was compared”*, with 0 committed reference images and its own self-test skipped. It compared nothing on either side. Rowed separately; deliberately not fixed here.


Talchain:
## Adversarial review — **APPROVE**

Independent review at head `b82b71117b56e3589e224d757aa449590722ae4f`, base `48b567db`. Fresh blobless clone at a unique `/private/tmp` path, `HEAD` asserted against the full 40-char SHA before any working-tree read. Every absence claim below carries a contrast control in the same invocation.

Nothing blocks the merge. One non-blocking documentation remedy (gate 2, measured below) and one merge-order coordination note (gate 4).

---

### Gate 1 — Consumers: PASS

Derived independently, not inherited. Multi-line-safe importer walk over `src/` (3,431 files scanned, non-empty asserted), counting **importer files**, not references, and covering `import`/`export … from`, dynamic `import()`, `require()` and named-symbol imports.

| target | importer files | any textual ref |
|---|---|---|
| `PreAnalysisGuidance` | **0** | **0** |
| `useCanRunAnalysis` | **0** | **0** |
| `limitsManager` | **0** | **0** |
| `hydrateLimitsAtBoot` | **0** | **0** |
| `graphPreflight` | **0** | **0** |
| `OutputsDock` *(control)* | 34 | 223 |
| `usePreRunValidation` *(control)* | 8 | 32 |

The uniform zeros would be suspect on their own; the two controls discriminate in the **same** invocation with large, specific numbers, so the zeros are real absence, not a blind probe.

Residual textual hits repo-wide are 21 in `ds-compliance-baseline.json`, 10 in `ia-v1-qa-findings.md`, 2 in `docs/technical/`, 1 in `scripts/isolation-results/external-files.txt`. **Zero in `src/`.** All checked below.

**On "dark ≠ dead":** `PreAnalysisGuidance` had **zero importers at base too** — it is not flag-gated-but-mounted, it is mounted by no path at all. The sparing rule cannot apply to it, so the reasoning *was* applied consistently. Correct call.

### Gate 2 — The mirror spec: CONFIRMED, and demonstrated

The spec's only import is `{ describe, it, expect } from 'vitest'`. Its own line 39 reads `// ----- Extracted/replicated gapItems logic (mirrors PreAnalysisGuidance.tsx) -----`, and it defines a private `computeGapItems`.

Not left as inference — **I ran it with no component on disk**: transpiled the spec standalone against a minimal `vitest` shim.

```
bundled spec: occurrences of "PreAnalysisGuidance" = 0 ; of "computeGapItems" = 20  (contrast control)
SHIM RESULT: passed=19 failed=0
```

19/19 green with the component absent entirely. This is fabricated coverage and deleting it is a net honesty gain.

**Two corrected premises** (reporting rather than routing around them):
1. It is **eight** criteria cited to the spec as ✅ PASS (E1, E2, E5–E10) — not ten. E3 is `✅ CODE VERIFIED` against JSX and E4 is `⚠️ MANUAL`. Control: 74 total PASS rows in that doc, fabricated token 0.
2. The doc's own summary row 250 claims `+10 | E1-E10`, overcounting its own detail table by two. Pre-existing defect, not introduced here.

**Does anything real still cover E1–E10? No — and nothing ever did.** At base, `gapItems` and `voiScore` existed in exactly the two deleted files and nowhere else (control `useState` = 398 files). The capability was never reachable, so nothing user-facing is lost.

**The one remedy.** `ia-v1-qa-findings.md` is undated and in the root — a *living* document by this estate's own naming rule — and after this PR it still asserts ✅ PASS for eight criteria citing a deleted file, plus a `CODE VERIFIED` row (J6) citing the deleted component. That is an honest artefact being removed while the false claim about it survives. Measured, not proposed — **12 lines, one file, no code**:

```
 | E1 | … | ❌ WITHDRAWN (PR #900) | `…gaps.spec.ts:E1` tested a private replica of the
                                    logic, not the component — no import, zero importers, both deleted |
 …E2–E10 likewise (E3/E4 note the JSX and click handler lived in the deleted component)
 | J6 | … | ❌ WITHDRAWN (PR #900) | component deleted; zero importers, never mounted |
 | ~~`PreAnalysisGuidance.gaps.spec.ts`~~ | withdrawn | mirror spec; 8 E-rows cited, not 10; covered nothing |
```
Verified after applying: residual `PASS`/`CODE VERIFIED` citations to the deleted files **0**, withdrawn rows **11**, no encoding damage.

### Gate 3 — Baselines: PASS, all five of the gate's own checks

Ran `typecheck-gate.sh`'s consistency checks directly against the PR's files:

| check (gate line) | result |
|---|---|
| exactly one `# count=` header, each file | 1 / 1 |
| header vs row sum (`:451`) | 2219 = 2219, 552 rows |
| identities header vs sum | 2219 = 2219, 1330 rows |
| per-file sums identities ⇄ baseline (`:475`) | **diff empty** |
| identities row format regex (`:464`) | 0 malformed |

The empty diff is the result that agrees with me, so I controlled it: perturbing one identity count by +1 (`applied=1`) makes check 4 RED. The probe discriminates. Removed rows sum to exactly 30 (27 + 3), 2249 − 30 = 2219 — arithmetic is right, and the two insertions really are just the two `# count=` lines.

Also checked: neither deleted file appears in `scripts/ci/typecheck-uncovered.txt`, whose bidirectional check would FAIL on a listed-but-deleted entry. (That list holds only `archive/` and `.tmp/` paths — my `src/` contrast returned 0, which I verified by reading the file rather than trusting the zero.)

### Gate 4 — Same-file collision with #901: REAL, but fail-loud

**#901 edits both baseline files and rewrites the same `# count=` line** (`2249 → 2247`; this PR `2249 → 2219`). Exactly the scenario that burned a prior merge. I measured both orders rather than reasoning:

- #900→#901 and #901→#900: **both CONFLICT** on the count line in both files. Git stops; it does not merge silently.
- Naive resolution test — union of both PRs' row deletions gives rowsum **2217** across 550 rows:

| resolver picks | outcome |
|---|---|
| `# count=2219` | **gate REDs** (header ≠ 2217) |
| `# count=2247` | **gate REDs** |
| `# count=2217` | passes |

So the damage is bounded and loud at both stages. **Not a blocker on this PR.** Coordination note for whoever merges second: regenerate the baselines rather than hand-resolving the header — the correct combined value is **2217**.

### Gate 5 — Sparing `usePreRunValidation`: correct

3 product importers (`pre-analysis/PreAnalysisPanel.tsx`, `pre-analysis/hooks/usePreAnalysisData.ts`, `canvas/hooks/usePreAnalysisData.ts`) + 5 spec importers. Genuinely live. Nothing else in the batch is in that position — the other three have zero importers of any kind.

### Gate 6 — Advisory red: discharged, and it is worse than "fails on both sides"

Not accepted on label or counts — the failing test **names** were diffed. Both runs: **29 passed, 5 skipped, 0 failed**; the failing set is empty on both sides and the normalised 36-line result sets are **byte-identical**. No swap.

But the mechanism matters more than the discharge: both runs log `Found 0 committed linux reference image(s)` and terminate at a step named *"Fail — linux references absent, nothing was compared"*. **This check compared nothing, on either side** — it is a missing-references guard, and it carries zero information about whether this PR changes the UI. Its own self-test suite (`selftest.visual.spec.ts` — *"proves the instrument can fail"*, *"a missing reference is a hard failure, never a silent bless"*) is **skipped** under `VISREG_BLESS=1` on both sides, so neither run contains evidence the harness can detect a regression at all.

That is a standing broken alarm, pre-existing and out of scope here, but it should be rowed rather than left to teach lanes to ignore a red. Supplementary (not a gate): comparing the two generated reference artifacts, 9 of 10 PNGs are byte-identical and one differs by 18 of 321,152 pixels, max channel delta 2/255, in a 17×18 px corner — consistent with antialiasing. Single sample per commit, so I cannot formally separate that from run-to-run noise.

### Gate 7 — CI at the exact SHA

`check-runs` for the full 40-char head SHA: `total_count=14`, **completed=14, running=0** — filtered on `.status` first, then `.conclusion`. 13 success, 1 failure (`Visual Regression (advisory)`). `mergeable: true`, `mergeable_state: unstable`, `merged: false`. Not silence, not stale.

### Gate 8 — Remaining sweeps

- **DS ratchet**: the claim is correct at the bytes — `--enforce` exits 1 only on **net-new** signatures above baseline; a downward move passes. Leaving the 21 stale entries is right. Corroborated empirically: `Staging Gate`, which depends on `ci:guard:ds:enforce`, is green.
- **`scripts/isolation-results/external-files.txt`**: `verify-isolation.sh` is referenced by **no** workflow and no `package.json` script (control: `typecheck` appears in 3 workflow files), and that file is a self-regenerating **output** (`find … > external-files.txt`), not a gate input. Inert.
- **Age / orphaning**: `limitsManager.ts` — 1 commit, 2025-11-17, never modified; orphaned by #893, which **is** the base commit here, so the claim is verifiable at this tip. `PreAnalysisGuidance.tsx` — 10 commits, last touched 2026-03-19.
- **Open-PR gate**: swept all 20 open UI PRs. Only #901 collides (gate 4). #898/#899 touch `src/components/results/**`, #622 touches `.github/dependabot.yml`; the 16 dependabot PRs touch no source. No open branch references any deleted module.

---

### What this evidence excludes

- **I did not run the full suite locally** — CI ran the repository, and its green at the exact SHA is the authority. My local work was the targeted spec, my own controls, and the gate's consistency checks.
- **The 19/19 spec run used a hand-written `vitest` shim**, not real vitest. It is evidence for *zero coupling to the component* (which is independently proven by the import list and by 0 occurrences in the bundled output), **not** a claim about behaviour under the real runner.
- **No bundle probe of my own.** I did not re-derive the author's deployed-bundle chunk sweep; with zero importers, absence from the bundle follows by construction, so that probe is corroborative rather than load-bearing. I take no position on its numbers.
- **Could not reach:** a full-history pickaxe for whether `PreAnalysisGuidance` was *ever* imported — `timeout` is absent on this platform and the unbounded search exceeded my budget (reported rather than silently `|| true`-d into a false absence). This does not affect the verdict: zero importers **at base** is what licenses the deletion.
- **Visual Regression:** one sample per commit; no visual inspection, pixel analysis is numeric only; only that job was analysed.

*Instrument disclosure, in the spirit of the estate's own traps: three of my own probes failed before producing anything I relied on — a `$MB:src` zsh history-modifier expansion, a BSD `sed` that silently no-opped on the remedy, and a perl one-liner that read pattern emoji as bytes and rewrote 0 of 8 rows while reporting success on a different line. Each was caught by asserting the post-condition (`after_refs` unchanged, `rewritten=0`) rather than trusting the exit code. Every number above comes from a probe whose control fired in the same run.*

PR #904, opened 2026-08-27
chore(visreg): bless the linux visual references — get the advisory check comparing something

Talchain:
## What this does

Commits the ten linux reference images so `Visual Regression (advisory)` starts **comparing** instead of sitting in its first-run generate-and-fail state.

## The premise correction: the check was never broken

The record has been calling this check broken. It is not. Its job header documents exactly this state:

> FIRST RUN ON A BRANCH WITH NO LINUX REFERENCES: chromium renders text differently on darwin and linux, so the committed darwin references cannot be compared here. The job generates the linux set, uploads it for review, and FAILS with instructions. **It never commits them — re-blessing is an explicit human commit.**

That is precisely what it has been doing. Run `33079224593` generated the set, reported `[visreg] completeness guard OK — 10/10 states captured by name`, uploaded it, and failed deliberately. It was waiting for the human step it requires by design. This PR is that step.

## Provenance

| | |
|---|---|
| Artifact | `visual-references-generated-linux` id `9649446928`, 1,627,315 B |
| Run | `33079224593`, job `98541640367` |
| **Captured at SHA** | **`e562524d495c709cf67595031a9f22ec14e7041b`** (this branch's merge base) |
| Generating run | 29 passed, 5 skipped, completeness guard 10/10 |

Generated **on the CI runner**, which `playwright.visual.config.ts` requires. A container that merely reports `process.platform === 'linux'` substitutes one glyph (`···` for `…`), which moves every node and lands fitView at 56 rather than 51. These captures read **51 @1280** and **58 @1440** — runner captures.

## Every image was looked at

5 states × 2 viewports. `fresh-draft` and `graph-default-zoom` are full-viewport; the other three are clipped to the 416px dock.

Canvas states were checked against the layout discriminator that actually discriminates — **distinct X positions relative to node count, plus nodes per column** — not origin-stacking and not gap uniformity: **~7 distinct columns for 19 nodes, ~2.7 nodes per column**. No origin stack, no single-column collapse. Dock clips measure 416px, matching the `shellLayout` assertion.

## ⚠ One thing these references record that reviewers should know

The tab strip is truncated to `Ol… / Anal… / Analysis (… / M…` at both viewports, in all ten images.

This is **not a defect being quietly entombed**. Traced to `f343e5bb` (today): the deliberately **temporary, unflagged** `Analysis (New)` comparison tab added beside `Analysis`. Truncation is the strip's designed overflow strategy — `shellLayout.visual.spec.ts` guards only that truncation has not eaten the protected freshness/verify affordances, and the run witnessed both painted.

Worth naming: the nav overflow guard passes (`scrollWidth 286 == clientWidth 286`) *because* the labels truncate. **That guard is structurally blind to label legibility.**

**Consequence, so it is not misread later:** when `Analysis (New)` retires, all ten references go red at once. Expected, and one re-bless commit fixes it. It must not be read as a regression.

Pre-existing canvas quality issues are recorded rather than created (node overlap, factor text wrapping to ~one word per line at 51%, top node clipped by fitView at 1280) — all visible in the older darwin set too. Freezing them is the point: the baseline exists to stop them getting worse.

## The self-test needs no re-enabling

`selftest.visual.spec.ts:153` skips **only** under `VISREG_BLESS=1`, by design — in bless mode `toHaveScreenshot` overwrites rather than compares, so it cannot assert a failure, and its perturbation test would write a 35%-widened panel as the canonical reference. CI was in bless mode because no references existed, hence 5 skipped. With references committed the job takes the compare branch (`pnpm run visual`, no bless) and **all five self-tests run**.

## Safety

Advisory only: `continue-on-error: true`, absent from the `Staging Gate` aggregator's `needs`. No branch protection depends on it — `staging` requires only `Staging Gate` (`needs: [tsc, typecheck-selftest, vitest, vitest-summary, build]`). Promotion to a required check is a separate, later decision and is **not** proposed here.

## No collision with the Playwright infrastructure lane

Confirmed from the configs rather than assumed. The visual suite is `playwright.visual.config.ts` — `testDir: 'e2e/visual'`, `testMatch: '**/*.visual.spec.ts'`, own `globalSetup`/`globalTeardown`, own port 5187 (`strictPort`, `reuseExistingServer: false`), own `outputDir` and report folder. The undeclared `axe-playwright` import is at `e2e/palette.spec.ts:12` (contrast control: 6 files use the declared `@axe-core/playwright`), collected by `playwright.config.ts` (`testDir: 'e2e'`) and **not** by the visual config. This PR touches only PNGs under `e2e/visual/references/linux/`.


Talchain:
## Discrimination proof — the deliverable, not the commit

All three runs polled to conclusion. `running=0` before any verdict.

### 1. GREEN on an unchanged tree
Run `33081433536`, job `98549550853` — **success**.

`Found 10 committed linux reference image(s)` — so the **compare** branch was taken, not generate. `34 passed`, **0 skipped** (the generating run was 29 passed / 5 skipped). `[visreg] completeness guard OK — 10/10 states captured by name`.

### 2. The self-test now runs — no change was needed
All five passed, having been skipped only because the earlier run was in bless mode:

```
✓ tolerance separates antialiasing noise from a real regression, by an order of magnitude
✓ the real assertion path REDS on a perturbed page
✓ a missing reference is a hard failure, never a silent bless
✓ a blank reference is rejected as non-substantive
✓ a large but near-uniform reference is rejected by the non-modal-pixel guard
```

### 3. Linux noise floor — the figure the job header says is unknown

```
tolerance maxDiffPixelRatio = 0.0005 (0.05%)
  noise floor (unmodified)          0.0000%  (0/1296000 px)
  panel width +35 percent           2.6679%  (34576/1296000 px)
  sticky footer overlap             0.6826%  (8846/1296000 px)
  marginal 1px control nudge        4.2564%  (55163/1296000 px)
```

**The linux noise floor is 0 differing pixels of 1,296,000 — identical to the measured darwin figure.**

Measured **twice independently** (runs `33081433536` and `33081821654`), both exactly zero, each comparing a fresh runner capture against references generated on a *third* run (`33079224593`). Three capture events, zero drift.

Recording it here because the job header asks for exactly this before anyone considers promotion. **This PR does not propose promoting the check** — that remains a separate, later decision.

### 4. RED on a deliberate visual change
Run `33081811447`, job `98550858533` — **failure**. `--info-rgb` `39 122 157` → `200 60 40` in `brand.css` (`--primary-rgb` aliases it).

**8 of 10 captures RED**, named: `fresh-draft`, `blocked-provisional`, `model-tab`, `graph-default-zoom` at both viewports. Diffs 1,117–2,801 px against a 0-px noise floor.

⭐ **`olumi-tab` stayed GREEN at both viewports.** That matters more than a clean sweep would have: the check reds the captures whose pixels actually moved and leaves the others alone. A blanket 10/10 red would have been equally consistent with "any diff reds everything".

The self-test's noise-floor case also correctly RED'd (`0.1468%`, 1,903 px) — it detected that the tree no longer matched its reference.

### 5. GREEN control — the other half of the pair
Run `33081821654`, job `98550924829` — **success**. A **real rendered layout change** (`max-w-[360px]` → `max-w-[280px]`) on the `Analysis (New)` tab **body**, a surface none of the five captured states mounts.

Not vacuous: `Found 10 committed linux reference image(s)`, `34 passed`, completeness guard `10/10`, noise floor `0.0000%`.

**Together, 4 and 5 prove the check binds to the rendered output of the captured states — not that it fires on any source diff.**

---

Proof PRs #905 and #906 are closed and their branches deleted; the evidence persists in the run logs above.

Talchain:
## Review — APPROVE, with two findings added to the record before blessing

I inspected the references myself rather than reviewing the report, because for a ten-PNG zero-code PR **looking at the images IS the review.** Author ≠ reviewer; there is no production code to gate.

### The proof reproduces, verified at the JOB level
⚠ Worth recording how: the GREEN-control **run** reads `status=in_progress, conclusion=null`, which a run-level filter would score as "not success". The job itself had concluded. Checked per-job:

```
33081433536  GREEN unchanged        Visual Regression: completed / success
33081811447  RED on --info-rgb      Visual Regression: completed / failure
33081821654  GREEN control          Visual Regression: completed / success
```

⭐ **The RED result is stronger than a clean sweep**: 8 of 10 red, `olumi-tab` staying GREEN at both viewports. A 10/10 red would fit "any diff reds everything" equally well. And the control is non-vacuous — it found 10 references, ran 34 tests and the completeness guard read 10/10 while staying green on a real layout change to a surface no captured state mounts. **The pair proves binding to rendered output rather than to any source diff.**

**Linux noise floor 0 differing pixels of 1,296,000**, measured twice independently against a third run's references. Identical to darwin. That number is the evidence for any later promotion decision — **and this PR correctly does not promote it.**

### Finding 1 — the tab strip, as disclosed, and I am confirming it
All ten images render `Ol… | Anal… | Analysis (… | M…`. **Four tabs, and the only legible one is the experimental `Analysis (New)` comparison tab** — the temporary unflagged surface has pushed every permanent tab into illegibility.

⭐⭐ The author's own observation is the important half and I want it kept: **the nav overflow guard passes (`scrollWidth 286 == clientWidth 286`) *precisely because* the labels truncate. That guard is structurally blind to label legibility** — it is satisfied by the very thing that causes the defect.

### Finding 2 — NOT disclosed, found on inspection
`model-tab--1440x900.png` truncates **factor labels** to the point of non-identification: **`In-Hou…`** and **`Time to Li…`**. Option rows fare little better (`Build In-House Metering and Invoic…`, `Buy Vendor Solution (Metronome o…`). The value column's `Not set / Olumi: …` text wraps across three lines and collides with the `AI estimate` pill, and the bottom row is clipped mid-content.

This is the **same shape as finding 1 one surface along**, and it is a *product* legibility issue rather than a capture artefact. A factor a user cannot name is a factor they cannot act on — and this estate has already measured a separate defect where the assistant could not enumerate factors by value state.

### Why I am still merging
**A reference set captures current truth. Its job is to make future change visible, not to assert the present is good.** Blessing a state with a known legibility defect is correct *provided the defect is recorded rather than silently frozen* — silent blessing was the one hazard I named, and it has been avoided on finding 1 and is now avoided on finding 2.

Both findings are **rowed as product defects**, not as blockers here. Re-blessing is one commit; a check that cannot compare is a permanent blind spot.

⭐ Carried forward, as the author flagged: **when `Analysis (New)` retires, all ten references red at once.** Expected, one re-bless commit, and it must not be misread as a regression.

### Also confirmed
The self-test was never disabled: `selftest.visual.spec.ts:153` skips **only** under `VISREG_BLESS=1`, by design — in bless mode `toHaveScreenshot` overwrites rather than compares, so its perturbation test would write a 35%-widened panel as canonical. Committing references flipped the job to the compare branch and **all five self-tests now run, with no code change.**

Collision clean both ways: the visual suite is fully separate (`testDir: 'e2e/visual'`, own port, teardown and report), and the undeclared `axe-playwright` sits in a file only the main config collects.

PR #905, opened 2026-08-27
DO NOT MERGE — RED proof for #904 (visual check must fail on a real visual change)

Talchain:
Throwaway. Discrimination proof for #904.

Changes `--info-rgb` in `src/styles/brand.css` from `39 122 157` to `200 60 40`. `--primary-rgb` aliases `--info-rgb`, so this recolours affordances visible on every captured surface.

**Expected: `Visual Regression (advisory)` REDs**, naming the affected captures. Closed and deleted once observed. Do not merge.

PR #906, opened 2026-08-27
DO NOT MERGE — GREEN control for #904 (visual check must NOT fire on an unrendered change)

Talchain:
Throwaway. Negative half of the discriminating pair for #904.

Changes a REAL rendered layout value — `max-w-[360px]` to `max-w-[280px]` — on the `Analysis (New)` tab **body**. That surface is a real product surface, but none of the five captured states selects it (only its truncated tab *label* appears in the strip).

**Expected: `Visual Regression (advisory)` stays GREEN.** Paired with #905, this proves the check binds to the rendered output of the captured states rather than firing on any source diff. Closed and deleted once observed. Do not merge.


Talchain:
Proof observed and recorded in #904. Closing; branch deleted. Evidence persists in the run logs.

PR #909, opened 2026-08-27
Analysis (New): 'At a glance' — a 5-to-10-second read designed natively for 320px

Talchain:
Adversarial pass over the two concept mock-ups, then the strongest defensible version. Measured first: the dock is **416px at 1024/1440/1920** — fixed, not resizable — giving **320px** usable content. The concepts render at ~1030px, so this is designed natively rather than shrunk.

## No producer, therefore absent
- **'Biggest leverage opportunity'** — swept with a contrast control: no leverage producer exists. The only hits are a Tornado caption that explicitly *disclaims* the word, and a canvas-local edge-weight sum.
- **Per-driver robust/uncertain dots** — `attribution_stability`, `flip_risk_category`, `rank_flip_rate` appear in zero fixtures; the field always present (`confidence`) is a known placeholder the display policy exists to suppress.
- **'72% model health'** — `useModelHealth()` returns an issue *list*, not a score.

## Dropped on design grounds
The drivers→outcome diagram (every driver points at the same focal object — the topology is a star and the arrows carry nothing the labels don't, for ~2/3 of the horizontal budget); the 41/22/15% figures (they sum to 78% against one outcome, which reads as a share; neither basis licenses that); icon tiles; and a primary intervention row that Strengthen renders identically ~120px below.

## What it shows, all traced
| Visible | Producer | Absent behaviour |
|---|---|---|
| Headline | `verdict.hasLeadingOption` + leader label | none — glance leads with drivers |
| Win share | `winProbability`, same gate | omitted |
| Trust word | `robustness.display_verdict` → 4-member map | `not_assessed` → **nothing** |
| Its reason | `display_verdict_reason`, verbatim | word alone |
| Bars | `displayInfluence`, scaled to the run's **max** | section omitted |
| Could change if | `flipThresholds`, gated on `flipThresholdsStatus` | **omitted** |

## Two branches deleted, not merely deduplicated
Mounted on a real run, all three key insights restated the glance ~120px above. The robustness and comparative branches fired under *exactly* the glance's conditions, so once it shipped they were unreachable. Both deleted; the comparative's win share now rides the glance trust line, so nothing was lost. The hinge insight **stays** and is suppressed only when the glance actually rendered a condition — different producers (`topFragileEdge` vs `flipThresholds`), pinned with a discriminating twin.

## Defects the mounted witness caught that ~90 tests did not
- `'… passes 1'` — a bare number on the model's own scale. Now pairs with `current_value`, or drops the number when there is no baseline.
- Key insights printed 'No insight is grounded well enough to lead with yet' while the glance above stated three. Now renders nothing.

## Verification
Typecheck gate at baseline; DS v5 no net-new (its raw-weight guard caught a redundant `font-normal`); analysisNew + ci-guards + results = 478 passing. Mounted at 320px on a genuine completed analysis, plus empty and pre-run states. Old Analysis untouched — no existing results component is modified.

### DIFF
--- a/src/components/results/analysisNew/AnalysisNewTabBody.tsx
+++ b/src/components/results/analysisNew/AnalysisNewTabBody.tsx
@@ -37,6 +37,7 @@ import { ANALYSIS_NEW_COPY as COPY } from './analysisNewCopy'
 import { ANALYSIS_NEW_LIMITS } from './buildAnalysisNewViewModel'
 import { useAnalysisNewViewModel } from './useAnalysisNewViewModel'
 import { AnalysisNewSection } from './sections/AnalysisNewSection'
+import { AtAGlance } from './sections/AtAGlance'
 import { StrengthenTheReasoning } from './sections/StrengthenTheReasoning'
 import { DeeperAnalysis } from './sections/DeeperAnalysis'
 
@@ -137,6 +138,9 @@ export function AnalysisNewTabBody({
             {vm.status.statusNote}
           </p>
         ) : null}
+
+        {/* ── AT A GLANCE — the 5-to-10-second read ───────────────────────── */}
+        <AtAGlance glance={vm.atAGlance} onFocusTarget={focusTarget} />
 
         {/* ── 1. KEY INSIGHTS ─────────────────────────────────────────────── */}
         <AnalysisNewSection
--- a/src/components/results/analysisNew/AnalysisNewTabBody.tsx
+++ b/src/components/results/analysisNew/AnalysisNewTabBody.tsx
@@ -146,7 +146,17 @@ export function AnalysisNewTabBody({
         <AnalysisNewSection
           title={COPY.sections.keyInsights}
           findings={vm.keyInsights.insights}
-          emptyMessage={vm.status.isPreRun ? null : COPY.empty.keyInsights}
+          // ⚠ "No insight is grounded well enough to lead with yet" is FALSE
+          // when the run DID produce insights and the glance is simply stating
+          // them — witnessed on a real run, where the glance carried all three
+          // and this line then contradicted the surface directly above it. An
+          // empty list with a non-zero candidate count means "shown above", so
+          // the section renders nothing at all rather than a claim that is not
+          // true. The honest empty state survives for a run that genuinely
+          // produced none.
+          emptyMessage={
+            vm.status.isPreRun || vm.keyInsights.candidateCount > 0 ? null : COPY.empty.keyInsights
+          }
           onFocusTarget={focusTarget}
           onRunIntervention={runIntervention}
           testId="analysis-new-key-insights"
--- a/src/components/results/analysisNew/__tests__/AnalysisNewTabBody.spec.tsx
+++ b/src/components/results/analysisNew/__tests__/AnalysisNewTabBody.spec.tsx
@@ -80,9 +80,15 @@ describe('F · the three scenario classes (§24F)', () => {
   })
 
   it('GENUINE DECISION — comparative material appears, phrased as "currently scores higher"', () => {
+    // Stated ONCE, by "At a glance". It used to appear here AND as a key
+    // insight one viewport below — measured on a real run, all three insights
+    // were restatements of the glance.
     renderBody(genuineDecision())
-    expect(screen.getByTestId('analysis-new-key-insights')).toHaveTextContent(
+    expect(screen.getByTestId('analysis-new-glance-headline')).toHaveTextContent(
       'Raise price currently scores higher',
+    )
+    expect(screen.getByTestId('analysis-new-key-insights').textContent).not.toContain(
+      'currently scores higher',
     )
   })
 
@@ -162,10 +168,9 @@ describe('staleness contextualises without dominating (§20)', () => {
     expect(screen.getByTestId('analysis-new-status-stale')).toHaveTextContent(
       'The model has changed since this analysis ran.',
     )
-    // One line, not a banner stack: the findings are still on screen.
-    expect(
-      within(screen.getByTestId('analysis-new-key-insights')).getAllByTestId('analysis-new-key-insights-row').length,
-    ).toBeGreaterThan(0)
+    // One line, not a banner stack: the read is still on screen.
+    expect(screen.getByTestId('analysis-new-glance')).toBeInTheDocument()
+    expect(screen.getByTestId('analysis-new-glance-headline')).toBeInTheDocument()
   })
 })
 
--- a/src/components/results/analysisNew/__tests__/atAGlance.spec.tsx
+++ b/src/components/results/analysisNew/__tests__/atAGlance.spec.tsx
@@ -119,11 +119,26 @@ describe('could change if — a tipping point, gated on the honesty field', () =
 
   const ROW = { label: 'Two-month timeframe', node_id: 'n_time', current_value: 2, flip_value: 3 }
 
-  it('renders when the producer computed one', () => {
-    expect(withFlip('computed', ROW)).toEqual({
-      text: 'Two-month timeframe passes 3',
+  it('renders when the producer computed one, with its unit', () => {
+    expect(withFlip('computed', { ...ROW, unit: '%' })).toEqual({
+      text: 'Two-month timeframe passes 3%',
       targetId: 'n_time',
     })
+  })
+
+  it('pairs the flip value with the current one when the producer sent NO unit', () => {
+    // Witnessed on a real run before this was fixed: "Price increase for new
+    // customers passes 1" — a bare number on the model's own scale, which the
+    // reader cannot place. `current_value` is the reference, from the same row.
+    expect(withFlip('computed', ROW)!.text).toBe('Two-month timeframe moves from 2 to 3')
+  })
+
+  it('states the condition WITHOUT a number when neither a unit nor a baseline exists', () => {
+    // No baseline means no direction of travel — printing a lone figure would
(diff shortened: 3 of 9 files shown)


Talchain:
## Independent adversarial review — post-merge audit

**Verdict: FIX-FORWARD (3 gates, 6 rows). No revert.** The surface is well-built and its own honesty commentary is unusually good; reverting would cost more than it saves. But three live falsehoods reach users on the currently-deployed build.

**Status correction first.** I was dispatched as a merge gate. #909 merged at 22:57:01Z with zero human review and is **deployed now** — `version.json` = `8e667e35`, and I confirmed the surface is in the served bundle. So these are findings about live code, not about a proposal.

**Mount (P2): UNGATED and live.** `getOutputTabsForParity()` gates `olumi` and `compare`, then `return true` — nothing gates `analysisNew`. Confirmed in the deployed artefact itself (71-chunk BFS crawl): `analysisNew:{…,label:"Analysis (New)",presentedAsTab:!0,hiddenReason:""}`, vs `compare:{…presentedAsTab:!1}`. Contrast controls fired ("Compare" 12 chunks, `outputs-dock-tab-` 1). Every defect below is reaching people.

---

### GATE 1 — the Drivers empty state states a falsehood (proven by execution)

`COPY.empty.drivers` = *"This run did not return factor influence."* renders whenever `findings.length === 0` and not pre-run — `AnalysisNewTabBody.tsx:198`, no other condition. But `buildDrivers` filters `.filter(d => d.zeroReason == null)` (`buildAnalysisNewViewModel.ts:317-320`).

Executed at the merged tip:

```
input : driversStatus 'computed', 2 drivers, zeroReason 'zero_outcome_diff' / 'disconnected'
output: findings 0  ->  "This run did not return factor influence."   <-- FALSE
CONTRAST: driversStatus 'unavailable', drivers []  ->  THE SAME SENTENCE
```

The run *did* return factor influence and measured it at zero. And the two states are **indistinguishable to the user** — the sentence cannot tell "we measured it and it was zero" from "we got nothing".

**The honest discriminator exists and is unused:** `data.drivers.driversStatus` (`'computed'|'unavailable'|'skipped'|'error'`, `types.ts:718`) — the adapter **already reads it** at `buildAnalysisNewViewModel.ts:480`. This is the identical assessed-vs-unassessed split the file performs correctly for uncertainty and omits for drivers.

`COPY.empty.drivers` has **zero** test references; the uncertainty pair has 3.

### GATE 2 — a PARTIAL analysis carries no provisional marker (RED-first → GREEN, remedy tested)

`buildStatus` computes `isProvisional = analysisStatus === 'partial' || completeness.status === 'partial'` (`:685`), typed at `analysisNewTypes.ts:197`. **Zero renderers.** Complete manifest — all six `analysisNew/*.tsx` — read it 0 times. Contrast control in the same sweep: `isStale` = 4 hits in `AnalysisNewTabBody`.

The only user-reachable disclosure is `row('Result completeness', …)` inside `buildDeeper` → `DeeperAnalysis`, which is `useState(false)`, i.e. **collapsed by default**, rendering the bare enum `partial` in a `<dd>`. On the 5-to-10-second read a partial analysis is visually identical to a complete one — same headline, same "Ahead in N% of simulated futures", same "Stable".

Reachable, not theoretical: `useResultCompleteness.ts:264` returns `'partial'` from UI-side derivation over a real report.

Worth noting: `buildAnalysisNewViewModel.spec.ts:202` is titled *"discloses partial completeness as provenance"* but only asserts the view model **computed** the boolean. Nothing asserts it renders.

**Remedy proven** in an isolated worktree (isolation established by sentinel write + differing inodes; source hash unchanged). Pristine: 3 tests collected by name → 1 failed, with the precondition case passing so the failure is provably about the missing render, not a dud fixture. After an ~8-line fix (`COPY.status.provisional` + a `role="status"` line beside the stale notice): 6 spec files / 82 tests green, and `scripts/ci/typecheck-gate.sh` exit 0.

### GATE 3 — decision-VOI copy breaches the ceiling its owning module documents

New (`analysisNewCopy.ts:109`): *"Resolving the open unknowns could still change this decision."*

Owner (`voi/resolveNextCopy.ts:114-121`), verbatim: *"Says the run DID NOT MEASURE ZERO. It does NOT say there is value — that claim needs a decision-level precision handle (a noise floor, CI or `n_samples`) which the wire does not carry, so a small positive value is not distinguishable from estimator noise. 'Did not come back at zero' is the ceiling of what this field can support."*

Same verdict, same source (`Number.isFinite(raw) && raw !== 0`). The owner's own sentence is *"this run did not come back at zero"*. "Could still change this decision" **is** the value claim the owner says the wire cannot support.

---

### ROWS (not gates)

1. **"Could change if" prints unrounded floats and garbles non-currency units.** #909's own delta (`AtAGlance.tsx` and `glanceCondition` do not exist at base `885985dc`; contrast: `useResultsSectionData.ts` does). Executed: `%`→`"passes 0.361111%"`, `£`→`"£0.361111"`, `index`→**`"passes index0.361111"`**, `scale`→**`"passes scale0.361111"`**. Two defects: no rounding *anywhere* (affects every unit), and every non-`%` unit prefixed (right for currency, wrong for a scale *name*). Tests pin `unit:'%'` only; `index`/`scale` = 0 hits in `analysisNew/__tests__`. Remedy tested → `"passes 0.36%"`, `"£0.36"`, and `index`/`scale` fall through to the reference-point form (this function's own stated rule for an unprintable unit). 82 tests green, typecheck 0.
2. **The `no_flip_in_range` guard is dead code, and its comment credits the wrong guard.** `FlipThreshold` (`types.ts:99-116`) does not declare it, and `useResultsSectionData.ts:2148-2158` rebuilds every row from a **closed 7-key literal with no spread**, dropping it. So the cast at `:604` always reads `undefined` and the clause never suppresses. Verified by execution that the guard *does* fire when the field is present — it is correct code on an object production never produces. The comment at `:590-597` claims it replaced a "right by luck" check; on the live path the protection is still only `flip_reason === 'found'`. No observed harm (0 of 57 captured rows pair a numeric `flip_value` with `no_flip_in_range: true`) — latent.
3. **`empty.uncertaintyAssessed`** (*"Nothing was flagged as consequentially uncertain"*) is gated on `evidenceGapsAssessed` = `Array.isArray(m1Coaching?.evidence_gaps)` — **one of the four sources** the findings list is built from. The gate attests to one; the sentence makes a negative claim over all four.
4. **`empty.keyInsights`** (*"No insight is grounded well enough to lead with yet"*) asserts a grounding judgement. `buildKeyInsights` applies no grounding threshold — it is pure presence-checking.
5. **Re-typed copy outside its single-source owner.** `'Could change if'` is defined exactly twice repo-wide: `heroCopy.ts:331` (owner, 2026-07-04, 35 commits — the *existing* Analysis tab hero, i.e. the very surface this is compared against) and `analysisNewCopy.ts:60`. Worse: `'Show on canvas'` is re-typed as bare JSX at `StrengthenTheReasoning.tsx:150` and `DisclosureRow.tsx:126` while `FOCUS_ON_CANVAS_LABEL` (`focusOnCanvasCopy.ts:72`) owns it — and that module's docblock says the fix was *"delete the competing spelling, **not** to add a third"*. Measured: the const appears in 0 files under `analysisNew/`, 6 across `src/`; contrast `truncateAtWordBoundary` = 2 under `analysisNew/`, so the directory is searchable.
6. **The headline claim has no browser-level artefact.** `analysisNew|analysis-new` in `tests/visual-regression` + `e2e` = **0 files**. Contrast control, same sweep: the existing Analysis tab = **8**, including `e2e/geometry/analysisAnswerFirst.measure.ts` — a geometry harness built for exactly this class of claim. To be clear, `firstViewportCensus.spec.tsx` is **honest** about this: it explicitly disclaims pixels, the fold and layout. The PR does not overclaim in code. But "5-to-10-second read, natively for 320px" currently rests on an unrecorded manual drive.

---

### Corrections I owe

- **A finding I refuted before reporting.** My producer sweep concluded `flipThresholdsStatus` is *"unreachable on the live V5 path"* — greps for the literal in `mapV5AnalysisToReport.ts` and `analysisEnrichmentShape.ts` returned 0 with firing contrast controls. It is **not established**: `enrichmentToV2ResponseShape` builds its result with `...parsed.enrichment`, typed `Record<string, unknown>` — the whole envelope — and `v5RunSnapshotFactory.ts:80` passes it as `rawV2Response`, which the hook reads at `:1316`. **A name-grep over a file that spreads is structurally blind to every field it carries.** The field is rare on the wire (1 of 15 captures) but not dark. Not a defect.
- **What this evidence excludes:** I did not run the full suite (CI did, green) or the full required-check step sequence; all my runs asserted collected counts by name. No access to CEE/PLoT/ISL — every producer claim rests on this repo's dated captures (2026-05-10 → 2026-08-17) and its own types. No live wire capture of today's payload, and no browser drive of the deployed surface. Gates 1 and 2 and row 1 are proven by execution; gate 3 and rows 2–6 at the bytes. Rows 3–4 are static reachability, not witnessed on a real payload.

### What passed, and deserves saying

Leader entitlement is correctly enforced (`hasLeadingOption === true` strict, and `winShare` gated on the *same* variable, so the win-probability number cannot render without the entitlement). No CTA terminates in a refusal. `OutputsDock.analysisNewTab.spec.tsx` asserts the mount path itself with contrast controls, and its preservation case has a real positive control (`RESULTS_TESTID_FLOOR`) so it cannot certify a stubbed-out surface. The single-driver bar suppression, the refusal to scale bars to a sum, and the "never clip producer prose" fix are all correct and well-reasoned.

